PR #24, opened 2021-03-14
fix for missing image table in website/html_img/mean_diag.html

bitterbark:
This is a tiny change but is necessary for the mean_diag.html file to be written with actual images in it.

### DIFF
--- a/lib/CamDiag.py
+++ b/lib/CamDiag.py
@@ -702,7 +702,6 @@ class CamDiag:
 
                         #Create output file (don't worry about analysis type for now):
                         outputfile = img_pages_dir / f'plot_page_{var}_{season}_{ptype}.html'
-
                         img_data = [os.pardir+os.sep+assets_dir.name+os.sep+img.name, alt_text]  # Hacky - how to get the relative path in a better way?
                         title = f"Variable: {var}"              #Create title
                         tmpl = jinenv.get_template('template.html')  #Set template
--- a/lib/CamDiag.py
+++ b/lib/CamDiag.py
@@ -675,7 +675,6 @@ class CamDiag:
             idest = assets_dir / img.name
             shutil.copyfile(img, idest) # store image in assets
 
-
         mean_html_info = OrderedDict()  # this is going to hold the data for building the mean
                                         # plots provisional structure:
                                         # key = variable_name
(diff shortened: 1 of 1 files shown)


brianpm:
Look ok to me, but just took a quick peek without testing. I want to blame @nusbaume for that, but I'm not confident he didn't inherit it from me. 

Probably more important is that we should try to formulate a file naming scheme that will be robust and facilitate making these scripts+templates as simple as possible.

nusbaume:
Hi @bitterbark 

Sadly this fix doesn't work for me, because all of the image files have names like `LWCF_ANN_LatLon_Mean.png`, which means the new glob won't catch them (but the original glob does).  Can you point me to the test (and code) you ran that caused you to find this problem?

bitterbark:
That's weird. The glob command to fill mean_diag.html wasn't returning
anything to me in its original form, because my files are named with the
Mean as the third field
LWCF_ANN_Mean_LatLon.png
LWCF_DJF_Mean_LatLon.png
LWCF_JJA_Mean_LatLon.png
...
(see /glade/scratch/bundy/cam_diag_plots/website/assets/)

I'm running unmodified (other than that) code I downloaded Sunday, in
 /glade/u/home/bundy/diag/CAM_diagnostics/CAM_diagnostics/
using input config_dani.yaml

(Note I'm just using y'all's data files, that functionality worked great!)
plotting into  /glade/scratch/bundy/cam_diag_plots/website.

As a side-note, I can see that I could just have * instead of Mean to be
more consistent with your original line
although since all the figures are Mean it currently works for me. Unless
we might end up with non-mean figures in there.
ORIG : for img in assets_dir.glob(f"{var}_{season}_{ptype}_*.png"): PULL
REQ: for img in assets_dir.glob(f"{var}_{season}_Mean_{ptype}*.png"
BETTER?    :     for img in assets_dir.glob(f"{var}_{season}_*_{ptype}.png"
But that doesn't solve the mystery of the figures being written to
different names.



On Mon, Mar 15, 2021 at 3:15 PM Jesse Nusbaumer ***@***.***>
wrote:
>
> Hi @bitterbark
>
> Sadly this fix doesn't work for me, because all of the image files have
names like LWCF_ANN_LatLon_Mean.png, which means the new glob won't catch
them (but the original glob does). Can you point me to the test (and code)
you ran that caused you to find this problem?

nusbaume:
Hi @bitterbark 

Thanks for the info!  I think I found the problem.  Can you copy this file on Glade to your `scripts/plotting` directory:

`/glade/work/nusbaume/SE_projects/model_diagnostics/CAM_diagnostics_fork/scripts/plotting/plot_example.py`

and then try it again with the old glob method?  

I believe the issue is that I tried to standardize the plots names, but I missed the naming in `plot_example.py`, which of course is probably the one I should have done first.

Anyways, please let me know how it goes, and thanks for the help!

PR #26, opened 2021-04-06
Flexible function calls

### LINKED ISSUE #6: Should config file have ability to pass configuration arguments directly to scripts?
There are a myriad of different ways a user can configure a plotting script in python.  How should the CAM diagnostics manage the desire of a user to modify simple configuration options (e.g. the color bar used) for the diagnostic figures?  I can so far think of three options:

1.  Do not allow the user to configure the plots through the namelist.  If the user wants a plot configured differently, they can simply modify the plotting script directly, or add their own script to the namelist.  This is the simplest option from a development and maintenance standpoint, but requires users to know python in order to change anything.

2.  Allow all possible configuration options to be controlled via the namelist.  This greatly reduces the need of the user to know python in order to configure the scripts to their liking.  However, it increases the effort needed to develop and maintain the diagnostics, and may cause the namelist file itself to increase dramatically in size.

3.  Allow _some_ configuration options in the namelist, but require all other configuration options to either be determined automatically, or through option 1 (direct modification of the script itself).  This would likely be a a trade-off between options 1 and 2, both in terms of advantages and difficulties.

Any thoughts or opinions?

brianpm:
These changes are mainly focused on providing a more flexible way to call the functions that are inside the `scripts` directory. A new function called `function_caller` is added to `CamDiag`. 

As a way to test this, I also modified how the yaml file is treated. I allowed averaging scripts to be a nested dict, not just a list of functions. This allows the user to provide a list of args and kwargs as strings, and then we inspect `locals` to find out if they are actually variables. I'm not sure if this is robust, but it at least worked for my tests so far.

Other changes are part of my attempt to test this capability. I was using two CAM simulations that only have time series files, not history files. Skipping the time series file creation required some modifications. I also modified code to try to keep going if a variable is not found (used `warnings` for when that happens).

Oh, and I just realized I left a bunch of debugging/print statements in there, which should be deleted. 

Sorry for the messy-ness of this PR. Let me know if you have trouble extracting the useful parts.

Fixes #6

### DIFF
--- a/lib/CamDiag.py
+++ b/lib/CamDiag.py
@@ -81,7 +81,9 @@ for root, dirs, files in os.walk(_DIAG_SCRIPTS_PATH):
     #Add all sub-directories to python path:
     for dirname in dirs:
         sys.path.append(os.path.join(root,dirname))
-
+print(100*"=")
+print(sys.path)
+print(100*"=")
 #################
 #Helper functions
 #################
@@ -151,6 +153,27 @@ def construct_index_info(d, fnam, opf):
     if plot_type not in d[vname]:
         d[vname][plot_type] = {}
     d[vname][plot_type][temporal] = opf
+
+#####
+
+def function_caller(func_name: str, func_args: list, func_kwargs: dict = {}, module_name=None):
+    """Call a function with given arguments.
+    
+    func_name : string, name of the function to call
+    func_args : list, the arguments to pass to the function
+    func_kwargs : [optional] dict, the keyword arguments to pass to the function
+    module_name : [optional] string, the name of the module where func_name is defined; if not provided, assume func_name.py
+    
+    return : the output of func_name(*func_args, **func_kwargs)
+    """
+    if module_name is None:
+        module_name = func_name #+'.py'
+    # note: when we use importlib, specify the module name without the ".py" extension.
+    module = importlib.import_module(func_name)
+    if hasattr(module, func_name) and callable(getattr(module, func_name)):
+        func = getattr(module, func_name)
+    return func(*func_args, **func_kwargs)
+
 
 
 ######################################
@@ -262,6 +285,11 @@ class CamDiag:
 
         #Check if climatologies are being calculated:
         if cam_climo_dict['calc_cam_climo']:
+            # Skip history file stuff if time series are pre-computed:
+            if ('cam_ts_done' in cam_climo_dict) and (cam_climo_dict['cam_ts_done'] == True):
+                # skip time series generation, and just make the climo
+                print("  Configuration file indicates time series files have been pre-computed, will rely on those files only.")
+                return None
 
             #Notify user that script has started:
             print("  Generating CAM time series files...")
@@ -366,8 +394,10 @@ class CamDiag:
         if cam_climo_dict['calc_cam_climo']:
 
             #If so, then extract names of time-averaging scripts:
-            avg_func_names = self.__time_averaging_scripts
-
+            avg_func_names = self.__time_averaging_scripts  # this is a list of script names 
+                                                            # _OR_ 
+                                                            # a **list** of dictionaries with script names as keys that hold args(list), kwargs(dict), and module(str)
+            
             #Extract necessary variables from CAM configure dictionary:
             input_ts_loc    = cam_climo_dict['cam_ts_loc']
             overwrite_climo = cam_climo_dict['cam_overwrite_climo']
@@ -375,9 +405,19 @@ class CamDiag:
 
             #Loop over all averaging script names:
             for avg_func_name in avg_func_names:
+                # option to specify the module's name:
+                if isinstance(avg_func_name, dict):
+                    assert len(avg_func_name) == 1  # must just be {function_name : {args:[...], <kwargs:{...}>, <module:'xxxx'>}}
+                    has_opt = True
+                    opt = avg_func_name[list(avg_func_name.keys())[0]]  # un-nests the dictionary
+                    avg_func_name = list(avg_func_name.keys())[0]  # not ideal, but change to a str representation; iteration will continue ok
+                else:
+                    has_opt = False
 
-                #Add file suffix to script name (to help with the file search):
-                avg_script = avg_func_name+'.py'
+                avg_script =  avg_func_name + '.py'  # default behavior: Add file suffix to script name
+                if has_opt:
+                    if 'module' in opt:
+                        avg_script = opt['module']
 
                 #Create full path to averaging script:
                 avg_script_path = os.path.join(os.path.join(_DIAG_SCRIPTS_PATH,"averaging"),avg_script)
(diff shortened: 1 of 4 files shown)


lib/CamDiag.py
```diff
@@ -81,7 +81,9 @@
     #Add all sub-directories to python path:
     for dirname in dirs:
         sys.path.append(os.path.join(root,dirname))
-
+print(100*"=")
```
nusbaume, line 84: Can you comment these `print` statements out and add an extra comment stating that they need to be added to a debug log?

lib/CamDiag.py
```diff
@@ -366,18 +394,30 @@ def create_climo(self, baseline=False):
...
             var_list        = self.__diag_var_list
 
             #Loop over all averaging script names:
             for avg_func_name in avg_func_names:
-
-                #Add file suffix to script name (to help with the file search):
-                avg_script = avg_func_name+'.py'
+                # option to specify the module's name:
+                if isinstance(avg_func_name, dict):
+                    assert len(avg_func_name) == 1  # must just be {function_name : {args:[...], <kwargs:{...}>, <module:'xxxx'>}}
```
nusbaume, line 410: Can you add an error message like so:

`assert len(avg_func_name) == 1, "YAML function line must just be {function_name : {args:[...], <kwargs:{...}>, <module:'xxxx'>}}"`

lib/CamDiag.py
```diff
@@ -386,19 +426,29 @@ def create_climo(self, baseline=False):
...
+                    sys.path.insert(0, avg_script_path)
+                # NOTE: when we move to making this into a proper package, this path-checking stuff should be removed and dealt with on the package-level.
+                    
+                # Arguments; check if user has specified custom arguments
+                avg_func_args = [case_name, input_ts_loc, output_loc, var_list]
+                avg_func_kwargs = {"clobber":overwrite_climo}
+                if has_opt:
+                    if ('args' in opt):
+                        # RULES: it has to be a list of strings, and then we will take whatever of those are in locals
+                        assert isinstance(opt['args'], list)
```
nusbaume, line 440: Can you add an error message like so:

`assert isinstance(opt['args'], list), "Function arguments must be of type list."`

lib/CamDiag.py
```diff
@@ -386,19 +426,29 @@ def create_climo(self, baseline=False):
...
+                        assert all(isinstance(item, str) for item in opt['args'])
```
nusbaume, line 441: Can you add an error message like so:

`assert all(isinstance(item, str) for item in opt['args']), "Items in function argument list must be strings."`

lib/CamDiag.py
```diff
@@ -386,19 +426,29 @@ def create_climo(self, baseline=False):
...
+                        avg_func_args = list()  # start over
+                        for variableToCheck in opt['args']:
+                            if variableToCheck in locals():
+                                avg_func_args.append(locals()[variableToCheck])
+                            else:
+                                print("{} is not available".format(variableToCheck))
+                    if 'kwargs' in opt:
+                        avg_func_kwargs = opt['kwargs']
+                print(f"DEBUG: \n \t avg_func_name = {avg_func_name}\n \t avg_func_args= {avg_func_args}\n \t avg_kwargs= {avg_func_kwargs}")
```
nusbaume, line 450: Can you comment this print statement out and add an extra comment stating that it needs to be added to a debug log?

lib/CamDiag.py
```diff
@@ -456,18 +506,8 @@ def regrid_climo(self):
...
-            exec(regrid_func_import_statement)
-
-            #Create actual function call:
-            regrid_func = regrid_func_name+\
-            '(case_name, input_climo_loc, output_loc, var_list, target_list, target_loc, overwrite_regrid)'
-
-            #Evaluate (run) averaging script function:
-            eval(regrid_func)
+            regrid_func_args = [case_name, input_climo_loc, output_loc, var_list, target_list, target_loc, overwrite_regrid]
+            function_caller(regrid_func_name, regrid_func_args, func_kwargs={}, module_name=regrid_func_name+'.py')
```
nusbaume, line 510: It might make sense to add all of the extra `opt` code you have in `create_climo` here as well.  If the code works as expected then it shouldn't modify the current behavior, but will allow one to add arguments to regridding scripts in the future.

lib/CamDiag.py
```diff
@@ -522,17 +562,8 @@ def perform_analyses(self, baseline=False):
...
-
-            #Run averaging script import statement:
-            exec(anly_func_import_statement)
-
-            #Create actual function call:
-            anly_func = anly_func_name+'(case_name, input_ts_loc, output_loc, var_list, write_html)'
-
-            #Evaluate (run) averaging script function:
-            eval(anly_func)
+            anly_func_args = [case_name, input_ts_loc, output_loc, var_list, write_html]
```
nusbaume, line 565: It might make sense to add all of the extra `opt` code you have in `create_climo` here as well.  If the code works as expected then it shouldn't modify the current behavior, but will allow one to add arguments to analysis scripts in the future.

lib/CamDiag.py
```diff
@@ -588,18 +619,8 @@ def create_plots(self):
...
-            #Run regridding script import statement:
-            exec(plot_func_import_statement)
-
-            #Create actual function call:
-            plot_func = plot_func_name+\
-            '(case_name, model_rgrid_loc, data_name, data_loc, var_list, data_list, plot_location)'
-
-            #Evaluate (run) plotting script function:
-            eval(plot_func)
+            plot_func_args = [case_name, model_rgrid_loc, data_name, data_loc, var_list, data_list, plot_location]
```
nusbaume, line 622: It might make sense to add all of the extra `opt` code you have in `create_climo` here as well.  If the code works as expected then it shouldn't modify the current behavior, but will allow one to add arguments to plotting scripts in the future.

scripts/analysis/amwg_table.py
```diff
@@ -97,7 +98,7 @@ def amwg_table(case_name, input_ts_loc, output_loc, var_list, write_html):
     if not input_location.is_dir():
         errmsg = "Time series directory '{}' not found.  Script is exiting.".format(input_ts_loc)
         end_diag_script(errmsg)
-
+    print(f"DEBUG: location of files is {str(input_location)}")
```
nusbaume, line 101: Can you comment this print statement out and add an extra comment stating that it needs to be added to a debug log?

scripts/averaging/averaging_example.py
```diff
@@ -70,33 +79,19 @@ def averaging_example(case_name, input_ts_loc, output_loc, var_list, overwrite_c
...
-        if "PS" in cam_ts_data:
-            ps_climo = cam_ts_data['PS'].groupby('time.month').mean(dim='time')
-            cam_climo_data = xr.merge([climo_data, ps_climo])
-        else:
-            cam_climo_data = climo_data.to_dataset()
-
-        #Also add some special variables that should be carried along:
-        for special in ['hyam', 'hybm', 'P0', 'hyai', 'hybi']:
-            if special in cam_ts_data:
-                cam_climo_data[special] = cam_ts_data[special]
```
nusbaume, line 95: Is this code no longer needed?  I thought for spectral element CAM runs these quantities are missing unless added explicitly?

scripts/regridding/regrid_example.py
```diff
@@ -58,7 +58,7 @@ def regrid_example(case_name, input_climo_loc, output_loc,
 
         #loop over regridding targets:
         for target in target_list:
-
+            print(f"DEBUG: target = {target}")
```
nusbaume, line 61: Can you comment this print statement out and add an extra comment stating that it needs to be added to a debug log?

scripts/regridding/regrid_example.py
```diff
@@ -71,11 +71,16 @@ def regrid_example(case_name, input_climo_loc, output_loc,
             if not regridded_file_loc.is_file():
 
                 #Create list of regridding target files (we should explore intake as an alternative to having this kind of repeated code)
+                # NOTE: This breaks if you have files from different cases in same directory!
                 tclim_fils = sorted(list(tclimo_loc.glob("{}*_{}_*.nc".format(target, var))))
-
+                print(f"DEBUG: tclim_fils (n={len(tclim_fils)}): {tclim_fils}")
```
nusbaume, line 76: Can you comment this print statement out and add an extra comment stating that it needs to be added to a debug log?

nusbaume (changes requested):
This looks like a great improvement!  However, I do have some (hopefully minor) requests.  If you have any questions or concerns with my suggestions just let me know!

scripts/averaging/averaging_example.py
brianpm: I'm going to defer on this for now. You're probably right, but I can't remember exactly why I did this. Let's skip it for now, and if needed, that can be dealt with in a different PR.

nusbaume (approved):
@brianpm Everything looks good now!  However, I did end up making some modifications to avoid some errors/issues that occurred during testing.   If those mods look good to you then I will go ahead and merge this PR, but if not please let me know and we can modify this PR further.  

Also, if you want to quickly see just the changes I made then simply click on the two commits below and they should show up on your screen.  Thanks!

brianpm:
Thanks, Jesse! This all looks good to me.

PR #27, opened 2021-05-17
Annotate zonal mean line plots

brianpm:
These changes are to add axis labels and title to the zonal mean line plots. 

- xlabel is hard-coded to "LATITUDE"
- ylabel tries to use the units attribute of the data, then looks for a units kwarg (this might not play well with matplotlib if it is set; in which case, it should be "popped" out of the kwargs). If neither is found, do nothing for ylabel.
- title looks for "long_name" first, and then looks for "name" 

`_zonal_plot_preslat` is modified just to have the labels on the xaxis only appear on the outer axes object.


nusbaume:
Looks good, thanks!

nusbaume (approved):

PR #28, opened 2021-05-19
New Annual cycle plot capability

brianpm:
Adds a new plotting function called `square_contour_difference` to `plotting_functions.py`. This is to provide plots of the annual cycle, as in AMWG set 8 ("Annual cycle contour plots of zonal means"). The function is a little more general, taking any two 2-dimensional DataArrays and plotting them and their difference. 

It's pretty simple, but it tries to do what it can for annotation. It tries to determine the case names for the axes titles, it uses the coordinate names for the axis labels, and it tries to figure out the variable name and units for the figure title. 

Returns the figure object, which can then be saved by whatever script calls it. I didn't bother to put in a new script to actually produce these plots, just wanted to get the functionality covered.

If you want to test it, you can generate appropriate zonal mans from history files with something like:
```
ds2 = xr.open_dataset(hfiles)
datavar2 = ds2[varname]
aczm2 = datavar2.mean(dim='lon', keep_attrs=True).groupby("time.month").mean(dim='time', keep_attrs=True)
```

### DIFF
--- a/lib/plotting_functions.py
+++ b/lib/plotting_functions.py
@@ -336,50 +336,50 @@ def plot_zonal_mean_and_save(wks, adata, apsurf, ahya, ahyb, bdata, bpsurf, bhya
 #
 
 def square_contour_difference(fld1, fld2, **kwargs):
-    """Produce a figure with square axes that show fld1, fld2, 
+    """Produce a figure with square axes that show fld1, fld2,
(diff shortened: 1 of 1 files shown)


nusbaume (approved):
Looks like a useful feature that will be used in the future.  Thanks!

PR #29, opened 2021-05-26
Weighted RMSE and other Enhancements

brianpm:
This PR is on top of the flexible function calls one (https://github.com/NCAR/CAM_diagnostics/pull/26). 

It adds an `environment.yml` file which could be used to set up a conda environment that is capable of running `CAM_diagnostics`. 

In `CamDiag` it allows for using all the available files when start and end years are not supplied. (This could be made more robust.)

In `CamDiag` the creation of the time series directory is more robust using `pathlib.Path.mkdir`. 

In `CamDiag` try to skip regridding when no scripts are provided. (I think this breaks later steps though.)

In `plotting_functions` Add a function `wgt_rmse` that correctly calculates the area-weighted RMSE and uses that to annotate the maps. I'm not sure why I wrote the RMSE incorrectly before, but this one works for 2-d fields and 1-d weights for DataArray or numpy arrays. 

There could be other differences, but I don't think they are important (mainly warnings / print messages).


nusbaume (approved):
Looks good!  There were a couple concerns, but I just went ahead and pushed fixes for those concerns to your branch.  Thanks again for all of the diagnostics development!

PR #37, opened 2021-08-18
introduce GeoCAT for pressure level interpolation

brianpm:
This PR changes the interpolation from hybrid levels to pressure levels (i.e., `lev_to_plev`)  to use GeoCAT's `interp_hybrid_to_pressure` function. This introduces a new dependency on GeoCAT.

The previous version didn't actually do the correct interpolation because I neglected to ever include functional code. So this actually provides the functionality that is needed. 

Note that the performance of `interp_hybrid_to_pressure` is not great in my testing, but I think it does produce a correct result.

### DIFF
--- a/lib/plotting_functions.py
+++ b/lib/plotting_functions.py
@@ -12,6 +12,7 @@ import xarray as xr
 import matplotlib as mpl
 import cartopy.crs as ccrs
 from cartopy.util import add_cyclic_point
+import geocat.comp as gcomp
 
 #Set non-X-window backend for matplotlib:
 mpl.use('Agg')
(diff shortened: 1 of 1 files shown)


nusbaume:
@brianpm this PR looks fine to me, but I assume you needed to add GeoCAT to your conda environment?  If so then did you want to add a new commit that includes GeoCAT as a dependency in the `conda_environment.yaml` file?  Thanks!

brianpm:
Oh, yeah, I forgot we had that now. Do you want me to do that, or do you want to put it in?

nusbaume:
Either way is fine with me, although if you want me to do it I would need to know the name of the specific GeoCAT package you installed.  I assume it was `geocat-comp`?

andrewgettelman:
Are we already using GeoCAT? Do we really need it? I recall one principle was a minimal set of packages. Or is this not something to worry about as long as we have a conda environment?

brianpm:
@andrewgettelman -- I think adding GeoCAT as a dependency will pay off in the long run. Their recent development seems well aligned with our use cases. In this particular case, the function for interpolating from hybrid to pressure levels is based on code that I gave them. The alternative is to either extract their version or implement my version, but since they potentially have better long-term support, I think it makes sense to use this as a first point of contact with GeoCAT.

andrewgettelman:
Sounds good @brianpm, thanks for clarifying.

brianpm:
I just pushed an update to the environment yaml file. @nusbaume , can you check if I did this correctly?

nusbaume:
@brianpm Yep, that update looks good to me!  As an FYI I am currently waiting to see if I can get the `ncar_pylib` environment updated so that it contains the latest GeoCAT version (which is necessary for this PR).  Once that is done I'll do a quick test and then get this PR merged.  Apologies for the delay!

PR #55, opened 2021-11-02
Change env name to reflect new repo name

mgrover1:
Should we change the conda env name to reflect the new repo name?

Changing from `camdiag` to `adf_diag`


nusbaume:
Looks good to me!  Just want to make sure it is also OK with @brianpm, since he set-up this conda environment file originally.

brianpm:
Oh, yeah, that is totally fine with me!

nusbaume (approved):

PR #80, opened 2021-12-05
Per variable plotting options via YAML file

brianpm:
This PR is a first draft to allow the capability to set plotting options/defaults on a per variable basis. 

The options are provided in a YAML file, and as a first try, I put a file into `lib`. This file is used as the default if the `config_file` contains `use_defaults: true` in the basic info dict. As an option, the user could also use an alternate file by supplying a path in `defaults_file`.

Whichever defaults file is being used, it is read in the `__init__` method of the `AdfDiag` class. The entire YAML file is then provided as an attribute of the AdfDiag object. I called it `variable_defaults`.

Then it is up to the plotting routines to make use of these however they want. 

As an example, I modified `global_latlon_map.py` to look for the defaults. To do it, I converted the module's only function to use the new ADF api, so the function signature is now: `global_latlon_map(adfobj)`. 

The whole defaults file is available, and is put into a dict called `res`:
```
res = adfobj.variable_defaults
```
Then when looping through the variables, it checks whether that variable is specified in `res`:
```
if var in res:
    vres = res[var]
else:
    vres = {}
```
And that dictionary is then used in two ways. First, the data is converted to specified units using a `scale` and `add_offset` options along with a `new_units` string. Second, the dictionary `vres` is sent into `plot_map_and_save` as a keyword dictionary (i.e., `pf.plot_map_and_save(plot_name, mseasons[s], oseasons[s], dseasons[s], **vres)`). 

Finally, I modified `plot_map_and_save` to use some of the entries in `vres`. For this example, I worked on specifying the colormap and contour levels for both the mean plots and the difference. 

I built in an extra feature to pass matplotlib keyword arguments based on the matplotlib function. I don't expect this to be used much, but might be useful for generalizing some of the functions in `plotting_functions`. 

This is a first try at this, so I expect that we'll need to iterate. I also hadn't planned on doing anything to `AdfDiag` yet, but it seemed like that was going to be easier than having a separate way to get the file contents into the functions. @nusbaume  might want to take a look and see if I'm doing any major damage in there. @andrewgettelman said he might be able to work on expanding to some of the other functions.


lib/plotting_functions.py
```diff
@@ -121,49 +149,89 @@ def plot_map_and_save(wks, mdlfld, obsfld, diffld, **kwargs):
...
+    # Difference options -- Check in kwargs for colormap and levels
+    if "diff_colormap" in kwargs:
+        cmapdiff = kwargs["diff_colormap"]
+    else:
+        cmapdiff = 'coolwarm'
+    
+    if "diff_contour_levels" in kwargs:
+        levelsdiff = kwargs["diff_contour_levels"]  # a list of explicit contour levels
+    elif "diff_contour_range" in kwargs:
+            assert len(kwargs['diff_contour_range']) == 3  # start, stop, step
```
nusbaume, line 191: It would be good to put an assert failure message here like `'diff_contour_levels' must have exactly three values.`

lib/plotting_functions.py
```diff
@@ -121,49 +149,89 @@ def plot_map_and_save(wks, mdlfld, obsfld, diffld, **kwargs):
...
+
         if i == len(wrap_fields)-1:
-            levels = levelsd #Using 'levels=12' casued len() error in mpl. -JN
-            #Only use "vcenter" if "matplotlib" version is greater than 2:
-            if(mplv > 2):
-                norm = normfunc(vmin=-1*absmaxdif, vcenter=0., vmax=absmaxdif)
-            else:
-                norm = normfunc(vmin=-1*absmaxdif, vmax=absmaxdif)
-            cmap = 'coolwarm'
+            levels = levelsdiff #Using 'levels=12' casued len() error in mpl. -JN
```
nusbaume, line 222: We can probably remove my comment here, as it clearly seems to be working now.

scripts/plotting/global_latlon_map.py
```diff
@@ -44,6 +47,43 @@ def global_latlon_map(adf):
...
+    if compare_obs: 
+        data_name = "obs"  # does not get used, is just here as a placemarker
+        data_list = adfobj.read_config_var("obs_type_list")  # Double caution!
+        data_loc = basic_info_dict['obs_climo_loc']
+
+    else:
+        data_info_dict = adfobj.read_config_var("diag_cam_baseline_climo") 
+        data_name = data_info_dict["cam_case_name"] # does not get used, is just here as a placemarker
+        data_list = [data_name] # gets used as just the name to search for climo files HAS TO BE LIST
+        data_loc = data_info_dict["cam_climo_loc"]
```
nusbaume, line 76: All of this is already done below (likely brought in when you merged in main), so this whole code section can be deleted (except for possibly grabbing the `basic_info_dict` which is needed for `plot_type`).

nusbaume (changes requested):
Thanks for tackling this @brianpm !  The implementation looks good to me, although I do have some general clean-up requests.

Also, if you would like me to manage these requests just let me know and I can push the relevant changes to your branch.  Thanks!

.gitignore
```diff
@@ -6,3 +6,110 @@
...
+
+# pyflow
+__pypackages__/
+
+# Environment
+.env
+.venv
+env/
+venv/
+ENV/
```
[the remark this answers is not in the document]
brianpm: Yes, sorry about that... I grabbed a generic gitignore to deal with my own issues; I'll clean this up.

lib/adf_variable_defaults.yaml
```diff
@@ -0,0 +1,26 @@
...
+  contour_levels_range: [980, 1052, 4]
+  diff_colormap: "PuOr_r"
+  diff_contour_range: [-15, 15, 1]
+  scale_factor: 0.01
+  add_offset: 0
+  new_unit: "hPa"
+
+PS:
+  colormap: "Oranges"
+  contour_levels_range: [980, 1052, 4]
```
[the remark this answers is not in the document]
brianpm: That's very true! I'll change this range to better see Antarctica etc.

lib/plotting_functions.py
brianpm: I'm removing it, but we should check on setting an integer number of levels that are automatically selected by mpl.

scripts/plotting/global_latlon_map.py
brianpm: @nusbaume -- Yeah, I think something weird is going on with this. I think the merge did things that I didn't expect. I'll need to take a closer look, and maybe touch base with you later to make sure I merge things correctly.

andrewgettelman:
@brianpm, I'm trying to run this and don't see the plots having the right modified units, etc. My version does not seem to be picking up the 'vres' correctly? What have I done wrong? I'm pretty sure I'm executing the right code, and have wiped all my plotting directories so they are being rebuilt. Maybe something to do tomorrow during the hackathon...

andrewgettelman:
This does not seem to be satisified for me:
`        
if var in res:
            vres = res[var]
`
But I'm not sure how to debug it from here.

scripts/plotting/global_latlon_map.py
```diff
@@ -44,6 +47,43 @@ def global_latlon_map(adf):
...
+    # Use ADF api to get all necessary information
+    #
+    var_list = adfobj.read_config_var("diag_var_list")
+    basic_info_dict = adfobj.read_config_var("diag_basic_info")
+    plot_location = basic_info_dict["cam_diag_plot_loc"]
+    compare_obs = basic_info_dict["compare_obs"]
+
+    case_info_dict = adfobj.read_config_var("diag_cam_climo")
+    case_name = case_info_dict["cam_case_name"]
+    model_rgrid_loc = case_info_dict["cam_regrid_loc"]
```
andrewgettelman, line 61: In order to run it needs to look like this:
   
 model_rgrid_loc = adfobj.get_basic_info("cam_regrid_loc", required=True)

andrewgettelman, line 61: In order to run, it needs to look like this:

    model_rgrid_loc = adfobj.get_basic_info("cam_regrid_loc", required=True)

andrewgettelman (changes requested):
There are also a few instances of `adf.` as the object that should be `adfobj.`

andrewgettelman:
Got it working. Seems like it works well. 

I've also added 2 more variables and adjusted PSL....
[adf_variable_defaults.yaml.txt](https://github.com/NCAR/ADF/files/7663751/adf_variable_defaults.yaml.txt)
.

It's a yaml file, but had to call it .txt to embed it.

andrewgettelman:
Sample..
TGCLDLWP_ANN_LatLon_Mean
..

andrewgettelman:
Also, I tried to propagate 'units' to the colorbar through the plot_map_and_save routine in plotting_functions.py, but had some trouble in extending the 'mpi:' part and figuring out dictionaries....maybe a hackathon issue for tomorrow.

scripts/plotting/global_latlon_map.py
nusbaume: @brianpm I think in this case the issue can be remedied fairly easily (it looks like the merge put the conflicting code section just below yours).  I can go ahead and open a PR to the branch on your fork with the relevant fixes tonight (it will also give me a chance to test this new feature myself).

config_cam_baseline_example.yaml
```diff
@@ -79,6 +79,15 @@ diag_basic_info:
     #Location where diagnostic plots are stored:
     cam_diag_plot_loc: /some/where/you/want/to/have/diag_plots #MUST EDIT!
 
+    #Use default variable plot settings?
+    #If "true", then variable-specific plotting attributes as defined in
+    #ADF/lib/adf_variable_defaults.yaml will be used:
+    use_defaults: false
```
nusbaume, line 85: @brianpm can you change this value to `true` as requested by @andrewgettelman (unless you have a reason not to)?  Once that change has been made I'll go ahead and merge it.

nusbaume (approved):
Looks good to me!  Just one last minor change (after which I'll get this merged).

config_cam_baseline_example.yaml
brianpm: Yeah, just pushed this.

nusbaume:
Great, thanks for all the work @brianpm!

PR #85, opened 2021-12-07
Fixed map_plot_and_save attributes passing and example

### LINKED ISSUE #84: Fix a problem with passing plotting attributes in plotting_functions.py
### ADF run type

Model vs. Model

### What happened?

For the plot_map_and_save function in plotting_functions.py, the 'mpl' (MatPlotLib) passing attributes functionality was not properly initialized. This is corrected in this PR, and an example of adding labels for units to the colorbar in the map plots is illustrated in adf_variable_defaults. 

### ADF Hash you are using

35c75be

CISL machine

NPL (CISL machines only)

### Extra info

Borderline between bug fix and a feature (only hit if feature is enabled)

andrewgettelman:
This Closes #84 

I modified the plotting_functions.py code for the map_plot_and_save to fix an error in being able to assign matplotlib (mpl) attributes. This is implemented with a sample label on the colorbar for a particular variable.

It fixes a bug. Not an ideal solution (would be better to perhaps pass units for any variable? But this is needed for the mpl functionality to work....

### DIFF
--- a/lib/adf_variable_defaults.yaml
+++ b/lib/adf_variable_defaults.yaml
@@ -24,6 +24,9 @@ PRECC:
   scale_factor: 86400000
   add_offset: 0
   new_unit: "mm/d"
+  mpl:
+    colorbar:
+      label : "mm/d"
 
 TGCLDLWP:
   colormap: "Blues"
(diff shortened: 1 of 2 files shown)


nusbaume (approved):

nusbaume:
@brianpm when you get the chance can you review this PR?  It should be really quick.  Thanks!

brianpm (approved):
This all looks great to me. The change in `plotting_functions.py` fixes a bug that I introduced.

PR #91, opened 2021-12-13
Enable multi-case processing

### LINKED ISSUE #40: Allow for a list of cases to support multi-run analysis
Currently the diagnostics package  can only process one simulation (case) at a time.  However, as brought up by @andrewgettelman, there are various scientific reasons why it would be helpful to process multiple cases all at once, compare them all to the same observations or baseline, and then plot them all on the same figure.

In order to facilitate this new capability, the diagnostics package should be able to take in a list of cases that can then all be processed during the same diagnostics run, thus providing a single data collection that can be used to generated the relevant multi-case figures.

nusbaume:
Allow for a user to pass in multiple cases via the 'diag_cam_climo' section in an ADF config file.  Also allow for this setting to automatically generate a "main_website" directory which includes a webpage that brings together all of the individual case websites into a single location.

Closes #40


nusbaume:
@andrewgettelman I believe I have finished the development of multi-case processing in the ADF.  Do you want to try it out before I merge everything in?  I believe you just need to clone my fork:

https://github.com/nusbaume/ADF.git

check out the "enable_case_lists" branch, and then add a list of cases to the "diag_cam_climo" section as shown in `config_cam_baseline_example.yaml`.  If you want to look at the results then you should see a "main_website" in your "cam_diag_plot_loc" directory that contains all of the relevant web pages in a single location.

Of course if you don't have time that's perfectly fine, I just figured it would be good to run it by you before bringing it all in to the ADF itself.

Thanks!

andrewgettelman:
Hi @nusbaume , 

Thanks so much for doing this. I won’t be able to test it this week (I am at AGU and pretty busy).  I think it looks great for now. @brianpm I hope you will take a look and see. I looked at the yaml file and it seems straightforward. When I get back (which may be after the holidays) I’m happy to try to hack this into some of the plotting scripts. We can do some strategizing about tweaks to it (e.g., adding a group name and a ‘short name’ for each case).

So I think fine to push this out to the ADF group now and let people play with it.

Thanks again!

brianpm:
I took a pretty careful look through the changes, and it all seems in order. Assuming that tests are passing, I'd say go ahead with this. I actually test the functionality yet, but I think we can move forward with this as a feature and make any changes that are needed as they arise.

PR #104, opened 2022-02-01
Add link to wiki to documentation

bitterbark:
In helping Judith & Will, I realized that the wiki tab isn't too obvious, so I added a link at the top of the 'how to run' section.


nusbaume (approved):
Looks good, thanks @bitterbark!

Issue #108, opened 2022-02-15
Bug in Taylor Plots


andrewgettelman:
### ADF run type

Model vs. Model

### What happened?

There is a bug in the taylor plots: there is no standard deviation in the xarray weighted method. 

This occurs here:
https://github.com/NCAR/ADF/blob/main/scripts/plotting/cam_taylor_diagram.py#L372

A fix would be to write a weighted_std routine (as with the weighted_correlation routine). 

### ADF Hash you are using

187c929

### What machine were you running the ADF on?

CISL machine

### What python environment were you using?

NPL (CISL machines only)

### Extra info

Brian has provided a fix to this, which Andrew has implemented locally. Probably not working yet. 

I will try to check in Brian's running modification to a branch in my repo and link to it in this issue.

andrewgettelman:
Okay, I have implemented a fix that runs, but it needs to be checked by @brianpm since I'm not sure it is doing the right thing yet. Please go to:

https://github.com/andrewgettelman/ADF/tree/taylor_bug

Specifically: 

https://github.com/andrewgettelman/ADF/blob/taylor_bug/scripts/plotting/cam_taylor_diagram.py

This runs but produces a standard deviation that **MAY** be too low (see figure):
![amwg_taylor_diagram_ANN](https://user-images.githubusercontent.com/9616874/154163578-74a3be51-8d51-4c0d-98c3-2a2f9cc22f1a.png)

brianpm:
Ooof. Yeah, that doesn't seem right. I'll try to double check the weighted standard deviation calculation.

brianpm:
I have identified the bug. Turns out writing a function that quickly during a hackathon isn't necessarily a good idea. I normalized the weights incorrectly.

The correct function is this:

```python

def weighted_std(x, weights):
    """Weighted standard deviation.
    
    x -> xr.DataArray
    weights -> array-like of weights, probably xr.DataArray

    If weights is not the same shape as x, will use `broadcast_like` to 
    create weights array.

    Returns the weighted standard deviation of the full x array.

    """
    xshape = x.shape
    wshape = weights.shape
    if xshape != wshape:
        wa = weights.broadcast_like(x)
    else:
        wa = weights
    mean_x = x.weighted(weights).mean()
    dev_x = x - mean_x
    swdev = (weights * dev_x**2).sum()
    total_weights = wa.where(x.notnull()).sum()
    return np.sqrt(swdev / total_weights)
```

I'm getting a PR ready.

PR #116, opened 2022-02-22
correct weighted standard deviation in Taylor diagram

brianpm:
This fixes https://github.com/NCAR/ADF/issues/108

Adds the missing function for weighted standard deviation, called `weighted_std`. Does not provide much in terms of bells and whistles, but it is correct. Cross referenced with the ESMLab version and it gives nearly the same answer (precision error probably explains differences). 

The `taylor_stats_single` function uses the new function, and is update to provide the weights. I tested this with the standard ADF test case and it seems to provide sensible answers. 

I'm not sure how the bug was introduced without me noticing that I broke the script. I must have assumed that Xarray provided a `std` method for weighted data and just changed it at the last minute before the PR. Sorry about that.

### DIFF
--- a/scripts/plotting/cam_taylor_diagram.py
+++ b/scripts/plotting/cam_taylor_diagram.py
@@ -352,6 +352,32 @@ def weighted_correlation(x, y, weights):
     return cov_xy / np.sqrt(cov_xx * cov_yy)
 
 
+def weighted_std(x, weights):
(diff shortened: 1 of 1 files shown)


nusbaume (approved):
Looks good (and ran fine for me)!  I did push one commit to remove some old commented-out code, but that shouldn't impact anything.

Issue #112, opened 2022-02-22
zonal mean 3D variable trying to use a DASK array


bitterbark:
### ADF run type

Model vs. Model

### What happened?

 Generating zonal mean plots...
NOTE: Plot type is set to png
	 ⌛ zonal mean plots for SWCF
	 ⌛ zonal mean plots for LWCF
	 ⌛ zonal mean plots for PSL
	 ⌛ zonal mean plots for Q
Q has lev dimension.
...

Traceback (most recent call last):
  File "/glade/u/apps/opt/conda/envs/npl/lib/python3.7/site-packages/xarray/core/ops.py", line 190, in _call_possibly_missing_method
    method = getattr(arg, name)
AttributeError: 'Array' object has no attribute 'item'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "./run_adf_diag", line 175, in <module>
    diag.create_plots()
  File "/glade/u/home/bundy/cam/diag/ADF_top/ADF/lib/adf_diag.py", line 844, in create_plots
    log_section = "create_plots")
  File "/glade/u/home/bundy/cam/diag/ADF_top/ADF/lib/adf_diag.py", line 435, in __diag_scripts_caller
    module_name=func_name)
  File "/glade/u/home/bundy/cam/diag/ADF_top/ADF/lib/adf_diag.py", line 468, in __function_caller
    return func(self)
  File "/glade/u/home/bundy/cam/diag/ADF_top/ADF/lib/../scripts/plotting/zonal_mean.py", line 248, in zonal_mean
    oseasons[s], ops_season, ohya, ohyb, **vres)
  File "/glade/u/home/bundy/cam/diag/ADF_top/ADF/lib/plotting_functions.py", line 550, in plot_zonal_mean_and_save
    diffnorm = normfunc(vmin=min(np.min(diff),-1*np.max(diff)), vcenter=0.0, vmax=np.max(diff))
  File "/glade/u/apps/opt/conda/envs/npl/lib/python3.7/site-packages/matplotlib/colors.py", line 1324, in __init__
    super().__init__(vmin=vmin, vmax=vmax)
  File "/glade/u/apps/opt/conda/envs/npl/lib/python3.7/site-packages/matplotlib/colors.py", line 1135, in __init__
    self._vmin = _sanitize_extrema(vmin)
  File "/glade/u/apps/opt/conda/envs/npl/lib/python3.7/site-packages/matplotlib/colors.py", line 100, in _sanitize_extrema
[...]

PR #118, opened 2022-02-23
check for dask arrays in plotting functions

brianpm:
This addresses https://github.com/NCAR/ADF/issues/112 (I think).

This changes `plotting_functions.py` to try to import dask and then check whether a few arrays are dask.Array instances. This is needed because the color normalization in the zonal mean plots uses the min/max values of the field, which comes from pressure level interpolation that potentially produces a dask array. 

I think the fix is just to see if the fields are dask arrays, and if so, call their `compute` method. 

I only did minor testing of this fix. It would be great if @bitterbark could check if this actually addresses the issue correctly.

### DIFF
--- a/lib/plotting_functions.py
+++ b/lib/plotting_functions.py
@@ -41,8 +41,8 @@ def use_this_norm():
 def get_difference_colors(values):
     """Provide a color norm and colormap assuming this is a difference field.
 
-       Values can be either the data field or a set of specified contour levels. 
-    
+       Values can be either the data field or a set of specified contour levels.
+
        Uses 'OrRd' colormap for positive definite, 'BuPu_r' for negative definite,
        and 'RdBu_r' centered on zero if there are values of both signs.
     """
@@ -101,9 +101,9 @@ def wgt_rmse(fld1, fld2, wgt):
     assert len(fld1.shape) == 2,     "Input fields must have exactly two dimensions."
     assert fld1.shape == fld2.shape, "Input fields must have the same array shape."
     # in case these fields are in dask arrays, compute them now.
-    if has_attr(fld1, "compute"):
+    if hasattr(fld1, "compute"):
         fld1 = fld1.compute()
-    if has_attr(fld2, "compute"):
+    if hasattr(fld2, "compute"):
         fld2 = fld2.compute()
     if isinstance(fld1, xr.DataArray) and isinstance(fld2, xr.DataArray):
         return (np.sqrt(((fld1 - fld2)**2).weighted(wgt).mean())).values.item()
--- a/lib/plotting_functions.py
+++ b/lib/plotting_functions.py
@@ -14,6 +14,14 @@ import cartopy.crs as ccrs
 from cartopy.util import add_cyclic_point
 import geocat.comp as gcomp
 
+# check if we have dask (probably yes)
+# later we can check if objects are instances of da.Array
+try:
+    import dask.array as da
+    has_dask = True
+except ImportError:
+    has_dask = False
+
 #Set non-X-window backend for matplotlib:
 mpl.use('Agg')
 
@@ -100,6 +108,10 @@ def wgt_rmse(fld1, fld2, wgt):
     """
     assert len(fld1.shape) == 2,     "Input fields must have exactly two dimensions."
     assert fld1.shape == fld2.shape, "Input fields must have the same array shape."
+    if has_dask and isinstance(fld1, da.Array):
+        fld1 = fld1.compute()
+    if has_dask and isinstance(fld2, da.Array):
+        fld2 = fld2.compute()
     if isinstance(fld1, xr.DataArray) and isinstance(fld2, xr.DataArray):
         return (np.sqrt(((fld1 - fld2)**2).weighted(wgt).mean())).values.item()
     else:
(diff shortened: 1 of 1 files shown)


bitterbark:
Unfortunately this doesn't solve the problem I'm running into. 
(BTW this isn't important to me at all, just happened to be the configuration I was testing in--happy to just leave off zonal plots and/or Q to make progress on something else, or keep working on this because we'll need it fixed eventually).

Meanwhile, I've been playing around with the code rather ineffectively given my lack of understanding. It seems to get past the error from vmin if I add an intermediate step to use the np.min function to set it.

>    vmin = [np.min(diff),-1*np.max(diff)]
      vmin = np.min(vmin)
      diffnorm = normfunc(vmin=vmin, vcenter=0.0, vmax=np.max(diff))

Print statements imply this changes vmin from being a dask.array, although adding the `isinstance` logic doesn't catch it:
CODE
>    vmin = [np.min(diff),-1*np.max(diff)]
        print('vmin 1',vmin)
        if has_dask and isinstance(vmin, da.Array):
            print(' vmin1 is dask')
        else:
            print('vmin1 is NOT dask')
        **vmin = np.min(vmin)**
        print('vmin 2',vmin)
        print(' vmin item',vmin.item)

OUTPUT
> vmin 1 [<xarray.DataArray 'Q' ()>
dask.array<_nanmin_skip-aggregate, shape=(), dtype=float64, chunksize=(), chunktype=numpy.ndarray>
Attributes:
    units:    none, <xarray.DataArray 'Q' ()>
dask.array<mul, shape=(), dtype=float64, chunksize=(), chunktype=numpy.ndarray>
Attributes:
    units:    none]
**vmin1 is NOT dask**
vmin 2 -0.0012450343697265782
 vmin item <built-in method item of numpy.float64 object at 0x7feb96338810>

I tried playing around with the type giving in `isinstance`, since it is imported as da but used as da.array but the print statement says dask.array... but I don't know what I'm doing well enough to find it yet.

Even worse, getting past the vmin error just gets it to the same complaint about vmax, which doesn't make sense according to my (lack of) understanding because np.max is used to directly set vmax. (This was why I tried it with vmin!). 

>File "/glade/u/home/bundy/cam/diag/ADF_top/ADF/lib/plotting_functions.py", line 591, in plot_zonal_mean_and_save
    diffnorm = normfunc(vmin=vmin, vcenter=0.0, vmax=np.max(diff))
  File "/glade/u/apps/opt/conda/envs/npl/lib/python3.7/site-packages/matplotlib/colors.py", line 1324, in __init__
    super().__init__(vmin=vmin, vmax=vmax)
  File "/glade/u/apps/opt/conda/envs/npl/lib/python3.7/site-packages/matplotlib/colors.py", line 1136, in __init__
    self._vmax = _sanitize_extrema(**vmax**)
  File "/glade/u/apps/opt/conda/envs/npl/lib/python3.7/site-packages/matplotlib/colors.py", line 100, in _sanitize_extrema
    ret = ex.item()

Taking vmax out for a separate step is redundant and also doesn't change it 

> vmax = np.max(diff)
      print('vmax 1',vmax)
      vmax = np.max(vmax)
      print('vmax 2',vmax)
      diffnorm = normfunc(vmin=vmin, vcenter=0.0, vmax=vmax)
      print('vmax item',vmax.item)

>  vmax 1 <xarray.DataArray 'Q' ()>
dask.array<_nanmax_skip-aggregate, shape=(), dtype=float64, chunksize=(), chunktype=numpy.ndarray>
Attributes:
    units:    none
vmax 2 <xarray.DataArray 'Q' ()>
dask.array<_nanmax_skip-aggregate, shape=(), dtype=float64, chunksize=(), chunktype=numpy.ndarray>
Attributes:
    units:    none
vmax item <bound method _values_method_wrapper.<locals>.func of <xarray.DataArray 'Q' ()>
dask.array<_nanmax_skip-aggregate, shape=(), dtype=float64, chunksize=(), chunktype=numpy.ndarray>

So, not very useful so far, happy to try other stuff if someone sees it clearly or has an idea.

brianpm:
Oh, that's interesting, @bitterbark . Thanks for investigating. I'll try to reproduce the issue exactly, and then hopefully I can pin down what is going on. I might ping you for your yaml file if I can't get the same error.

brianpm:
@bitterbark & @nusbaume -- I just went back and figured it out.

The problem was what I thought: geocat's interpolation was returning a dask object, which does not play nice with numpy's functions during that color normalization step.

So the fix is to check whether there is a `compute` method available on the interpolated quantity (which there will be) and then run it to force the computation when it hasn't been done yet. This happens in the `lev_to_plev` function. A consequence is that this appears to speed up the computation (slightly). 

As I was correcting that, I also noticed that the line that was actually causing the problem could (a) be simplified, and (b) could be deleted. So I deleted it. The `norm` that gets used is calculated just above and is called `normdiff` rather than `diffnorm`. 

Since the check for dask was not effective, I also removed that `has_dask` stuff from everywhere I put it in. I did include another check for `compute()` in `wgt_rmse`. I don't think this is needed, but it should not do any harm. I updated my branch and pushed it, and it looks like it is ready in this PR.

bitterbark:
Thanks!

nusbaume:
@bitterbark I was able to reproduce your error and found that @brianpm's mods fix the issue (after I made a small change to the python syntax).  However, it would be good for you to check that it fixes the issue on your end as well.  Can you download the fixed version by doing the following:

```
git clone https://github.com/brianpm/ADF.git brianpm_ADF
cd brianpm_ADF
git checkout check_for_dask_arrays
```

And then re-run the ADF with the conda npl environment using that ADF version?  If it works for you too then this PR will be good to go.  Thanks!

lib/plotting_functions.py
```diff
@@ -80,7 +81,6 @@ def global_average(fld, wgt, verbose=False):
             break
     fld2 = np.ma.masked_invalid(fld)
     if verbose:
-        print("(global_average)-- fraction input missing: {}".format(fraction_nan(fld)))
```
nusbaume, line 83: Also @brianpm was there a reason you removed this `print` statement?

brianpm: Yes, because `fraction_nan` did not seem to be defined. Maybe I missed it?

bitterbark:
It worked for me! Thanks for the detailed instructions and sorry for the
delay!

PR #130, opened 2022-03-09
adding seasonal weighting to lat/lonplots and vertical levels

### LINKED ISSUE #121: Need ability to plot 3-D variables on 2-D maps when given pressure levels to use
### New feature type

New plot and/or plot enhancement, New regridding method or regridding enhancement

### What is this new feature?

Currently the lat/lon plotting script only accepts 2-D input variables, which must be 2-D in the model output file themselves.  However, it would be beneficial to have the ability to specify pressure levels on which to generate lat/lon maps of 3-D variables (e.g. `U` or `Q`).   This will require a relatively-expensive vertical interpolation, which should be done once during the re-gridding stage, instead of multiple times during the plotting stage.  

The pressure levels themselves could be passed into the ADF using either the variable defaults file, or as arguments in the config file itself (ideally both).

### Assistance required?

Yes, I will need some assistance before making a PR

### Extra info

Bringing this issue to the attention of @juliecaron , @WillyChap, and @judithberner , based on the quick discussion we all had at the last hackathon.

WillyChap:
From @nusbaume :

Fixes #121 
Addresses #132

### DIFF
--- a/Untitled.ipynb
+++ b/Untitled.ipynb
@@ -1,0 +1,92 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 3,
+   "id": "b29d0e9c-f150-4b6c-ab2f-fba31939caac",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "levs_des = [200,850]"
+   ]
+  },
+  {
+   "cell_type": "code",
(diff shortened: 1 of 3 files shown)


WillyChap:
@nusbaume feel free to push. I am in a "stable" state now. Good to go!

Issue #125, opened 2022-03-04
Polar Map I think is broken.


andrewgettelman:
### ADF run type

Model vs. Model

### What happened?

I tried to run the polar plots, straight off main. Model v. Model and model v. obs.

something is not working with passing the mseasons and oseasons data to the plot: goes in as an xarray data array, but has no units. Here is the error:

 File "/glade/u/home/andrew/adf/test2/lib/../scripts/plotting/polar_map.py", line 236, in polar_map
    nhfig = make_polar_plot(mseasons[s], oseasons[s], dseasons[s], hemisphere="NH", **vres)
  File "/glade/u/home/andrew/adf/test2/lib/../scripts/plotting/polar_map.py", line 373, in make_polar_plot
    ax1.set_title(f"{d1.name} [{d1.units}]")

I was not quite able to figure this out, so I am reporting it. 

### ADF Hash you are using

274516e

### What machine were you running the ADF on?

CISL machine

### What python environment were you using?

NPL (CISL machines only)

### Extra info

Think it is a pretty simple fix, maybe it's supposed to be a dataset not a data array?

andrewgettelman:
Thanks for checking Brian. Closing this issue and we will see if it comes up again.

PR #137, opened 2022-03-16
make polar plots more performant by reducing array to specified domain

brianpm:
As noted in the hackathons by several people, the polar map script takes a long time to run. I did a bunch of testing to see what was going on. (No issue was ever opened for this. I started messing with the performance in response to #125, but as I mentioned, I couldn't replicate that bug.)

There seem to be two culprits. 

First is that `contourf` can be kind of slow. Switching to `pcolormesh` saves quite a bit of time, especially when plotting differences (in my testing). This PR **does not** make this change, as it probably isn't as desired as regular contour plots.

Second is that `savefig` was taking a long time, often more than half the total time. I tried a few things, but the only one that helps is to reduce the arrays that are passed into `contourf`. Since the domain being plotted is only the polar region, this PR just reduces the arrays to only contain that data. As far as I can tell, the plots look identical to the current version, but the script runs much faster. 

I am assuming the reason for this result is that `contourf` was calculating the contours for the global domain, and then matplotlib and/or cartopy have to do a lot of work to transform the resulting contours to the polar projection. Possibly "hiding" the rest of the data is a burden in that transforming step (or whatever part of it happens at render time). 

I also changed the default yaml file to point to the updated CERES files.

### DIFF
--- a/lib/adf_variable_defaults.yaml
+++ b/lib/adf_variable_defaults.yaml
@@ -153,9 +153,9 @@ SWCF:
   mpl:
     colorbar:
       label : "Wm$^{-2}$"
-  obs_file: "CERES_EBAF_Ed4.1_SWCF_climo.nc"
+  obs_file: "CERES_EBAF_Ed4.1_2001-2020.nc"
   obs_name: "CERES_EBAF_Ed4.1"
-  obs_var_name: "SWCF"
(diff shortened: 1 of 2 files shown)


nusbaume (approved):
Looks good to me, and passed all of my tests.  Thanks!

Issue #92, opened 2021-12-14
Speed up computing climos 


cecilehannay:
### New feature type

Other (please explain below)

### What is this new feature?

I tried to use ADF to compute climos from the ne30-L58 runs. I was trying to compute a 10-year climo on the years 1980-1989

The script didn't go anywhere in 12 hours. I talked to Jesse who recommended that I added an issue. 

Examples of history files are at:
/glade/scratch/hannay/archive/f.e21.FWscHIST.ne30_L48_BL10_cam6_3_035.tphysac_reorder_zm2_tau7200.001.hf

### Assistance required?

Yes, I am requesting a new feature but won't have time to develop it

brianpm:
This seems odd. Looking at those files, I don't see a reason that it should be slower than the example data we've been using. @cecilehannay , were you trying to process many, many variables? 

The only other thing I can think of is that there wasn't enough memory available for `ncrcat` to work efficiently. I'll try to reproduce the issue, but I don't have any extra time until after Christmas, probably.

cecilehannay:
I was trying to process ALL the variables in the h0 files.

No emergency. I just want to add this to the issues as Jesse recommended.
Also, maybe I was doing something wrong.

cecilehannay:
@nusbaume : It was just hanging. First I thought it was slow. And I came back the following day and it was still at the same place. I will try it again and I will let you know.

nusbaume:
@cecilehannay That sounds like it might be a "real" bug.  If you are on the CISL machines can you point me to your config (XXX.yaml) file?

nusbaume:
@swrneale You should be able to look in the directories specified by either `cam_ts_loc` or `cam_climo_loc` under the `diag_cam_climo` section and see if any files are present.  If there are and they are slowly being added then it is likely just running slow.  However, if there are no files at all then the ADF is likely hanging.

Also, if it is hanging can you point me to your config (XXX.yaml) file?  That might help me figure out what exactly is going wrong.  Thanks!

PR #143, opened 2022-03-29
Faster timeseries and climos

brianpm:
This PR addresses https://github.com/NCAR/ADF/issues/92

Uses `multiprocessing` in `adf_diag` to speed up generation of time series files. This should help in situations where many variables are being processed because separate processes were being initiated for each variable (potentially bogging down system). Using `multiprocessing.Pool` sets up the processes to wait until previous ones are done. As @justin-richling showed at the hackathon, this dramatically decreases the time to generate time series files. 

Introduces `create_climo_files` as a drop-in replacement for `averaging_example`. This case is a little bit more of a change, but the same principle. Introduces a function that deals with calculating and saving a climo file for an individual variable. The list of arguments for all variables is collected and then the function is called in parallel using `multiprocessing.Pool`. Testing on a small set of variables, I was able to cut the time for ADF (time series generation + climo generation + nominal remapping (but no plots)) from about 12-13 minutes down to 2.5-3 minutes (on a Casper login node). 

In both cases, I just let multiprocessing use all the available CPUs. It would be easy to check for a user-specified number to control this from the config file.

### DIFF
--- a/config_cam_baseline_example.yaml
+++ b/config_cam_baseline_example.yaml
@@ -101,6 +101,13 @@ diag_basic_info:
     #given the same weight:
     weight_season: True
 
+    #Number of processors on which to run the ADF.
+    #If this config variable isn't present then
+    #the ADF defaults to one processor.  Also, if
+    #you set it to "*" then it will default
+    #to all of the processors on a single node:
+    num_procs: '*'
+
 #This second set of variables provides info for the CAM simulation(s) being diagnosed:
 diag_cam_climo:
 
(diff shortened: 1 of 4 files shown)


nusbaume:
Hi @brianpm, just checking if it is okay for me to push to this branch?

Also, is there any reason you can think of that the new `create_climo_files` couldn't replace `averaging_example` as the default climo generator?  If not then I'll go ahead and add it to the provided example config file.  Thanks!

brianpm:
Yes, I think it okay to push it and to replace `averaging_example`.  We should ask people to test it, but I don't think it should break anything.

nusbaume (approved):

PR #160, opened 2022-04-29
Add missing _load_dataset function to global_latlon_map.py

justin-richling:
Copy function from ```Zonal_mean.py```. Also pull out imports from ```global_latlon_map``` function and put at beginning of script


nusbaume:
These code changes were brought in with PR #160 , so this PR will now be  closed.

Issue #136, opened 2022-03-16
Interpolate to pressure levels for climos and internal ADF use.


andrewgettelman:
### New feature type

New regridding method or regridding enhancement

### What is this new feature?

For a number of reasons it would be nice to internally convert climatology files to standard pressure levels:

1. Comparison to observations (Reanalysis) for vertical plots (This supersedes issue #123 , which will be closed)
2. Comparison across vertical coordinates (SE hybrid-sigma v. MPAS height)
3. Smaller storage use (probably need only about 11-20 levels, not 58, unless diagnostics for QBO and such). 

It does not seem like this would be too hard to do. See below 

### Assistance required?

Yes, I am requesting a new feature but won't have time to develop it

### Extra info

 @brianpm developed a vertical interpolation routine that can go from Hybrid pressure to standard pressure levels. I have modified this to be able to go from 3D pressure (provided by MPAS) to standard pressure levels. 

I've modified Brian's code to do interpolation in log pressure, and to add the second routine. I've verified off line this works for FV and MPAS output from CAM.  Code below.

I suggest that we can build this into the climatology generation as an option: specify a flag for using pressure, and then an optional list for the actual pressure levels desired (otherwise we use defaults). 

This should permit vertical interpolation as well internally if all on the same grid. It might require modifying some of the zonal mean-lat height plotting slightly. It will probably enable 3D observational comparisons too (#123) as noted above.

Comments welcome.

The code looks like this:

```python
#@njit(parallel=True)  # NOTE: if this is breaking, get rid of the decorator, and change 'prange' to range
def vert_remap(x_mdl, p_mdl, plev):
    """Apply simple 1-d interpolation to a field, x
       given the pressure p and the new pressures plev.
       x_mdl, p_mdl are numpy arrays of shape (nlevel, spacetime).

       NOTE: we try to use Numba here to parallelize the loop.
       If Numba is not available, I tried to 'spoof' the decoratory,
       but I have not tested much. If it breaks, remove the njit decorator,
       and change `prange` to `range' in the `for` statement.
       
       Andrew: changed to do interpolation in log pressure
    """
    out_shape = (plev.shape[0], x_mdl.shape[1])
    output = np.full(out_shape, np.nan)
    for i in prange(out_shape[1]):
        output[:,i] = np.interp(np.log(plev), np.log(p_mdl[:,i]), x_mdl[:,i])
    return output


def pres_from_hybrid(psfc, hya, hyb, p0=100000.):
[...]

Issue #133, opened 2022-03-10
Need to perform common calculations once in a single location


nusbaume:
### New feature type

New infrastructure or infrastructure enhancement

### What is this new feature?

Right now many plotting scripts repeat various common calculations, such as vertical interpolation and seasonal averaging.  This results in redundancies that hurt the performance of the ADF, and also make it difficult to modify these sorts of calculations in the future (as they are spread across multiple different python files).  Thus it would be better if these calculations were done in a single location or set of locations before any plotting or statistical analyses are done, such as during the re-gridding or climatology generation phase.

### Assistance required?

No, I will make a PR when the feature is ready

### Extra info

This may be  best done by adding new `averaging` and `regridding` scripts to the ADF that do these calculations in a modular and easily editable fashion before the plotting and analysis scripts are run.

andrewgettelman:
I'm going to throw a big note of caution here. There is a tradeoff between the efficiency that Jesse is discussing and modularity and ease of code development. What happens when data are not just monthly means? I don't think the ADF infrastructure should do too much work. Fine if we want to make some of our core plotting work with this, but we should not require it. Think what happens with other use cases. 

The integrated calls where there are scripts that call other functions make development more complicated (in my experience). 

Fine to make recommended functions/scripts available if plots want to use them.

nusbaume:
The way I see it the only new addition to the ADF infrastructure is one new variable that records the different available types of model data (monthly history files, monthly time series, monthly climatologies, seasonal climatologies, climatologies on pressure-levels, etc.) and where they are located, which is likely needed to implement daily and sub-daily model data usage anyways.  

Otherwise all of the actual calculations will be done in separate averaging and re-gridding scripts like what is done currently.  Then the plotting scripts are simply expected to check this new model data variable to see if the model data they want is available.

TL/DR: if a user wants seasonal climatologies or pressure-level model data they either need to calculate it themselves in their plotting scripts OR add the associated averaging/re-gridding script to the relevant script list.  Either way it's a responsibility of the user and not the ADF itself (with one user-defined way being more computationally efficient).

PR #168, opened 2022-05-19
Centralize vertical interpolation, and allow for 3D Model/Obs comparisons

### LINKED ISSUE #121: Need ability to plot 3-D variables on 2-D maps when given pressure levels to use

nusbaume:
This PR adds a new regridding and vertical interpolation script which interpolates all model variables with a vertical component onto a standard set of pressure levels.  This allow one to now compare 3D model variables against 3D observations, assuming the observations are also on the same set of pressure levels.

Also, this PR enables the use of specifying plotting pressure levels for the polar plots as well, so now all current 2-D map plots can display 3D variables.

Finally, this PR removes some example scripts which are no longer used in the example config file.

Addresses #133 
Addresses #136
closes #121 

## Tests run:

Ran both single-case and multi-case ADF runs on Casper using both the `npl` and `adf_v0.07` conda environments.


nusbaume:
@andrewgettelman Just letting you know that once this PR is in then we can bring in the MPAS vertical interpolation (I've already added code to check for the `lev` long name, although right now it is only looking for hybrid levels).

andrewgettelman:
Thanks Jesse!  This is awesome. It will be simple to add into it then.

PR #175, opened 2022-06-07
Enable interpolation of data on MPAS vertical coordinate

### LINKED ISSUE #136: Interpolate to pressure levels for climos and internal ADF use.

nusbaume:
This PR brings in code to check for the MPAS height-based vertical coordinate, and if present to enable the pressure-to-pressure vertical interpolation required to get the MPAS data onto the standard pressure levels used by the ADF.

This PR also removes the extra variables used for vertical interpolation from any time series that contains a 2-D model field, as that field should never see the vertical interpolation routines, and eliminating these extra variables can help save disk space.  

Closes #136

## Tests run:

Performed single and multi-case ADF runs on Casper using both the `npl` and `adf_v0.07` conda environments.  Also performed a couple of ADF runs on horizontally-regridded MPAS data to make sure the MPAS vertical interpolation works as expected.

### DIFF
--- a/lib/adf_diag.py
+++ b/lib/adf_diag.py
@@ -649,19 +649,20 @@ class AdfDiag(AdfObs):
                             #Set model to height (z) vertical levels:
                             vert_coord_type = "height"
                         else:
-                           #Print a warning, and assume that no vertical level information is needed.
-                           wmsg = "WARNING! Unable to determine the vertical coordinate"
-                           wmsg +=f" type from the 'lev' long name, which is:\n'{lev_long_name}'."
-                           wmsg += "\nNo additional vertical coordinate information will be"
-                           wmsg += " transferred beyond the 'lev' dimension itself."
-                           print(wmsg)
+                            #Print a warning, and assume that no vertical
+                            #level information is needed.
+                            wmsg = "WARNING! Unable to determine the vertical coordinate"
+                            wmsg +=f" type from the 'lev' long name, which is:\n'{lev_long_name}'."
+                            wmsg += "\nNo additional vertical coordinate information will be"
+                            wmsg += " transferred beyond the 'lev' dimension itself."
+                            print(wmsg)
 
-                           vert_coord_type = None
+                            vert_coord_type = None
                         #End if
                     else:
                         #Print a warning, and assume hybrid levels (for now):
                         wmsg = "WARNING!  No long name found for the 'lev' dimension,"
-                        wmsg += f" so no additional vertical coordinate information will be"
+                        wmsg += " so no additional vertical coordinate information will be"
                         wmsg += " transferred beyond the 'lev' dimension itself."
                         print(wmsg)
 
@@ -695,7 +696,9 @@ class AdfDiag(AdfObs):
             list_of_commands = []
             for var in self.diag_var_list:
                 if var not in hist_file_var_list:
-                    print(f"WARNING: {var} is not in the file {hist_files[0]}. No time series will be generated.")
+                    msg = f"WARNING: {var} is not in the file {hist_files[0]}."
+                    msg += " No time series will be generated."
+                    print(msg)
                     continue
 
                 #Check if variable has a "lev" dimension according to first file:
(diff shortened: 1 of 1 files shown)


nusbaume:
Great, thanks @justin-richling!

PR #180, opened 2022-06-10
Change check for missing climo years

### LINKED ISSUE #149: Error from unspecified start and end climo years in config yaml file
### ADF run type

Model vs. Model

### What happened?

Running the ADF without start and end dates crash the package when getting to the ```case_name``` var in the ```perform_analyses``` function in ```lib/adf_diag.py``` script.

### Expected behavior
In the config yaml file it states if the start and end dates are left blank, the package would grab available years from the datasets

### Error
```
Traceback (most recent call last):
  File "./run_adf_diag", line 180, in <module>
    diag.perform_analyses()
  File "/glade/work/richling/ADF/ADF/lib/adf_diag.py", line 788, in perform_analyses
    if syears[case_idx] and eyears[case_idx]:
TypeError: 'NoneType' object is not subscriptable
```

### ADF Hash you are using

5848b92

CISL machine

NPL (CISL machines only)

### Extra info

_No response_

justin-richling:
Closes #149 

This will now change how the script deals with unspecified climo start and end years in the config file.

### Changes

Prior to this change, the script would fail at the ```case_name``` var in the ```lib/adf_diag``` script if the climo years were not specified in the config file. This instead will check if either the start or end year are ```NoneType``` and handle them accordingly.

A warning message was also added, but can be removed if deemed unnecessary.


justin-richling:
This is still in progress, ran into some last minute problems when trying it on the multi-case examples. Stay tuned.

PR #182, opened 2022-06-14
Plot clean up

### LINKED ISSUE #176: Consistency across plotting scripts
I've noticed that the plotting functions for the polar maps could use some consistency compared to the other plot types. 

For example, the plotting function for polar maps currently reside in the ```polar_map.py``` file, where as the plotting functions for the other plot types reside in the ```plotting_function.py``` file. 

Also for the polar maps, the ```savefig``` and ```plt.close``` calls are not included in the ```make_polar_plot``` function but instead get called after the ```make_polar_plot``` call.

I can clean these up if there are no specific reason why they are laid out this way now.

justin-richling:
Closes #176 

Work towards a consistent setup throughout the plotting script files and plots.

This PR will move the plotting function for polar maps to ```lib/plotting_functions.py``` where complimentary functions for other plot types live. Likewise, it will move the ```domain_stats``` helper function as well.

This will also layout the lat/lon vector plots to match the lat/lon global, ie move colorbar positions and sizing, match fig size, and add lat/lon ticks.

### DIFF
--- a/scripts/plotting/polar_map.py
+++ b/scripts/plotting/polar_map.py
@@ -1,5 +1,4 @@
 from pathlib import Path  # python standard library
-from typing import Optional  # this is just for type hints
 
 # data loading / analysis
 import xarray as xr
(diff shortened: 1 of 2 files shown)


justin-richling:
Looks like in ```polar_map.py``` the code snippets (lines 260-264 and 333-337):
``` 
       # Save file:
       fig.savefig(plot_name, bbox_inches='tight', dpi=300)
                                        
       # Close figure to avoid memory issues:
       plt.close(fig)
```
were kept (probably from the pull from NCAR/ADF main). This is redundant as this part of the code is in the ```make_polar_plot``` function in ```plotting_functions.py``` now.

scripts/plotting/polar_map.py
```diff
@@ -256,7 +255,7 @@ def polar_map(adfobj):
                                         hemi = "SH"
                                     #End if
                                     
-                                    fig = make_polar_plot(mseasons[s], oseasons[s], dseasons[s], hemisphere=hemi, **vres)
+                                    fig = pf.make_polar_plot(plot_name, mseasons[s], oseasons[s], dseasons[s], hemisphere=hemi, **vres)
```
nusbaume, line 258: If you remove the `return fig` statement from the `make_polar_plot` function then this can be changed to:

```suggestion
                                    pf.make_polar_plot(plot_name, mseasons[s], oseasons[s], dseasons[s], hemisphere=hemi, **vres)
```

Also the `fig.savefig` and `plt.close` calls below should be deleted.

scripts/plotting/polar_map.py
```diff
@@ -329,7 +328,7 @@ def polar_map(adfobj):
                                             hemi = "SH"
                                         #End if
                                     
-                                        fig = make_polar_plot(mseasons[s], oseasons[s], dseasons[s], hemisphere=hemi, **vres)
+                                        fig = pf.make_polar_plot(plot_name, mseasons[s], oseasons[s], dseasons[s], hemisphere=hemi, **vres)
```
nusbaume, line 331: If you remove the `return fig` statement from the `make_polar_plot` function then this can be changed to:

```suggestion
                                        pf.make_polar_plot(plot_name, mseasons[s], oseasons[s], dseasons[s], hemisphere=hemi, **vres)
```

Also the `fig.savefig` and `plt.close` calls below should be deleted.

nusbaume (changes requested):
Looks good, but I do have some (hopefully) minor requests.

nusbaume (approved):
Looks great now, thanks!

PR #185, opened 2022-06-17
Display only generated plot types on webpage

### LINKED ISSUE #165: New website throws a browser error when a plot type is missing
### ADF run type

Model vs. Model

### What happened?

If someone runs the ADF without a specific plot set (e.g. lat/lon vector) and generates a website, then the website will still contain a link to those plots, that when clicked will result in a missing file browser error (which could look semi-intimidating to a new user).

One solution would be to check if any files of that plot type exist, and if not, generate a webpage that just says this plot type doesn't exist.  An alternative solution would be to simply remove the plot type entirely from the website list if no plots of that type are available.

### ADF Hash you are using

4f7fce9

CISL machine

ADF-provided Conda env

### Extra info

_No response_

justin-richling:
Closes #165 

This change will only display user generated plot types from config yaml file on the output webpage.

Tested with single and multiple case ADF runs, as well as with model vs obs case. Examples on the ADF project page (while still available) [here](https://project.cgd.ucar.edu/projects/ADF/webpage-missing-ptype/).

### DIFF
--- a/lib/adf_diag.py
+++ b/lib/adf_diag.py
@@ -1109,10 +1109,10 @@ class AdfDiag(AdfObs):
 
         #Since polar has more than one plot type name, make a list of lists
         #that grab all the paths and names
-        ptype_html = sorted([ptype_html_dict[x] for x in plot_func_names])
-        ptype_order = sorted([ptype_order_dict[x] for x in plot_func_names])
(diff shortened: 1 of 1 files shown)


nusbaume (approved):
Looks good, thanks for getting this done so quickly!

PR #186, opened 2022-06-21
check variables are present for Taylor Diags, otherwise skip Taylor Diags

cecilehannay:
Simple check to ensure all the needed variables for Taylor diags are present

### DIFF
--- a/config_cam_baseline_example.yaml
+++ b/config_cam_baseline_example.yaml
@@ -330,6 +330,7 @@ plotting_scripts:
     - global_latlon_vect_map
     - zonal_mean
     - polar_map
+    - cam_taylor_diagram
 
 #List of CAM variables that will be processesd:
 #If CVDP is to be run PSL, TREFHT, TS and PRECT (or PRECC and PRECL) should be listed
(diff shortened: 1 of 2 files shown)


nusbaume (changes requested):
Just one change, otherwise the code modifications look good!

Also, would it make sense to add the Taylor diagram script (and its associated variables) to the example `config_cam_baseline_example.yaml` file, now that it is hooked into the website and won't trip up if variables are missing?  If so then feel free to do that in this PR (otherwise I can do it in a later PR).  Thanks!

nusbaume:
@justin-richling @cecilehannay Just FYI that I added a function to the Taylor diagrams that will now look for a `PS` climo file if it can't find the surface pressure in the variable files themselves (which may happen if you are reading CMIP6 time series directly).   

All this means is that as long as `PS` is listed in `diag_var_list` then the Taylor diagram routine should now be able to find it and then run normally (assuming `hyam` and `hybm` are still contained within each 3-D variable time series).  Of course if this is not the case please let me know.  Thanks!

Issue #169, opened 2022-05-19
Need equatorial cross section


brianpm:
### New feature type

New plot and/or plot enhancement

### What is this new feature?

The request is to add capability similar to old AMWG diagnostics set 4a: "Vertical contour (XZ) plots of DJF, MAM, JJA, SON and ANN meridional means". 

### Assistance required?

No, I will make a PR when the feature is ready

### Extra info

This is from our initial assessment of desired features (in the spreadsheet). I talked with Meg (whose name is attached), and she's not planning to get to it.

I should be able to provide a new plotting script that adds this basic feature, and I will at least provide methods to do both "XZ" (i.e., longitude-pressure) plots as well as just "X" (longitude) plots. 

**stretch goal**
I may also explore a generalization to allow arbitrary cross sections (assuming regular gridded data). I've got code that does it fine, but the complication is how to let the user specify the cross section, variables to use, and possibly other choices that need to be made (e.g., interpolation vs nearest-neighbor)

nusbaume:
This issue has been resolved by PR #188 , so closing now.

PR #188, opened 2022-06-23
Meridional plot

brianpm:
This addresses https://github.com/NCAR/ADF/issues/169

It very closely follows the zonal mean plots. To reduce code duplication, I modified a bit of `plotting_functions.py` so that zonal and meridional mean plots can share some code.

Adds plotting script: `meridional_plot.py`.

In `plotting_functions` -- 
- add function `validate_dims` and modified `lat_lon_validate_dims` and `zm_validate_dims` to use that function.
- add function `_plot_line` and modified `_zonal_plot_line` to use that function
- add function `_meridional_plot_line` which also uses `_plot_line`
- add function `_meridional_plot_preslon` which is very similar to existing `_zonal_plot_preslat`
- add function `meridional_plot` which is very similar to existing `zonal_plot`
- add function `prep_contour_plot` which produces a dictionary of plotting parameters used for contour plots
- modified `plot_zonal_mean_and_save` to use `prep_contour_plot`
- add function `plot_meridonal_mean_and_save` that is similar to `plot_zonal_mean_and_save` and uses `prep_contour_plot`

I believe the code is complete, but I have not had a chance to test it yet. Making this a draft PR for now in case anyone has thoughts or wants to take a look.


brianpm:
This should be ready to go.

I also added an option to plot log(p) as the vertical axis. This is done as a variable preference with the keyword `plot_log_pressure`. If that evaluates to True, then both the zonal and meridional contour plots will switch to log(p) by using Matplotlib's `set_yscale("log")` method. Tested this with `OMEGAT` in the regular ADF test data; added that to the `adf_variable_defaults.yaml` along with `H2O` (but that wasn't present in both datasets).

brianpm:
@nusbaume -- Yes. I just made a couple of minor changes based on comments from Dan, but now I'll probably ignore for the next 3 weeks.

nusbaume (approved):
Thanks for the new plot type @brianpm!

PR #191, opened 2022-06-29
Taylor casename fix

justin-richling:
When running Taylor Diagrams, there was a missing casename argument for the ```get_surface_pressure``` function in the ```cam_taylor_diagram.py``` script.

This has been fixed as well as a check for Taylor Diagram with Obs scenario. The website would still generate a broken link to the Taylor Diagram if it was included in the config file. This PR checks for that and removes it from the list of HTML generated links.


nusbaume (approved):
Thanks for fixing those bugs!

PR #195, opened 2022-07-08
Add check for CMIP files for AMWG table

justin-richling:
Currently the AMWG tables aren't created for CMIP cases and crash the script if AMWG tables are set in config file.

This will allow the script to continue if test case is compared to CMIP files and still create the AMWG table for the test case.


justin-richling:
This may be a temporary fix as eventually we aim to incorporate CMIP file structure into the ADF. 

Also, it may be subject to failure since it assumes that the case name given in the config yaml for the CMIP case indeed has "CMIP" in the name.

nusbaume (approved):
Looks good, thanks!

PR #197, opened 2022-07-13
New website generator and API

### LINKED ISSUE #48: Need to create ADF case "info" class
For most ADF diagnostic runs there are usually quantities describing the simulations being used (model start and end date, horizontal resolution, available model variables, etc.) that if not provided by the user in the config file may require the opening of at least a few model files, and that can be useful to scripts throughout the ADF.

Thus it would be helpful to create a new ADF class that extracts this info directly from the model case files when not provided by the user, and make it available to any scripts that require it.  Given that ecg-tools can basically do all of this already, this new class could serve as an interface for ecg-tools/Intake-ESM (issue #25) and the rest of the ADF.

### LINKED ISSUE #106: ADF needs way to have new plots included in website automatically
### New feature type

New infrastructure or infrastructure enhancement

### What is this new feature?

Currently whenever a new plot type is added to the ADF the internal ADF variable `plot_type_order` has to be updated to include that plot in the generated website.  Ideally any plotting script that wants its plots to be included in the website will instead notify the ADF via a standard interface, and the ADF will then include those plots automatically.

This way a future plot developer won't need to modify any code in the underlying ADF infrastructure while at the same time giving control over the plot/website interaction to the plot developer.

### Assistance required?

### Extra info

This should be done after issue #99 has been resolved.

### LINKED ISSUE #194: Website will not be created if only creating tables
### ADF run type

Model vs. Model

### What happened?

The website will not be created if only creating tables (and no plots)

Here is the error I am getting.

----- 

The analysis_scripts and plotting_scripts in my config_file is set to: 
```
#List of analysis scripts being used.
#These scripts must be located in "scripts/analysis":
analysis_scripts:
    - amwg_table

#List of plotting scripts being used.
#These scripts must be located in "scripts/plotting":
plotting_scripts:
    #- global_latlon_map
    #- global_latlon_vect_map
    #- zonal_mean
    #- polar_map
    #- cam_taylor_diagram
    #- qbo 
```

The error is: 
```
Generating Diagnostics webpages...
Traceback (most recent call last):
  File "./run_adf_diag", line 190, in <module>
    diag.create_website()
  File "/glade/work/hannay/ADF/ADF/lib/adf_diag.py", line 1112, in create_website
    ptype_html = sorted([ptype_html_dict[x] for x in plot_func_names if x in ptype_html_dict])
TypeError: 'NoneType' object is not iterable
```

### ADF Hash you are using

682c95b

CISL machine

NPL (CISL machines only)

### Extra info

None

nusbaume:
Re-factor the web generation code such that a user that adds a new plotting or analysis script won't have to do anything to the website code itself, but instead will simply have to call a new `add_website_data` ADF method and pass in the relevant descriptors.  Right now this method can only use paths to image files (for plots) or Pandas data frames (for tables), but can always be expanded in the future if need be.

This PR also adds the ability to specify a variable's web category using the variable defaults file, removes some superfluous Jinja templates, and creates two new ADF classes ("info" and "web") to help with modularization and testing.

Closes #48
Closes #106
Closes #194
Closes #196


nusbaume:
@justin-richling This is a pretty big PR, so feel free to review as much (or as little) as you want.  If there is one section I would recommend for you to look over before this gets merged it would be the new `create_website` function in the `adf_web.py` file, as that is where we'll need to implement your new website design.  Thanks!

justin-richling:
@nusbaume This all looks good, thanks for the hard work refactoring the website creation! I'll start looking into adapting my multi-case, color tables and new layout into this.

nusbaume:
Great, thanks for reviewing @justin-richling!

PR #200, opened 2022-07-15
provide Jupyter notebook 

### LINKED ISSUE #56: Put together an example of running this from a Jupyter Notebook
As was mentioned in the meeting, it would be good to have an example of running this from a Jupyter Notebook

bitterbark:
This provides a Jupyter notebook that runs the cam baseline example as defined in the config file.

It also changes the config file to be almost out-of-the-box for users, using default directories provided that they enter their user name at the top. The jupyter notebook has a check for this but we may want to write a check into the code for it, too. 

Resolves #56


nusbaume:
@bitterbark It looks like there are some failing tests in this PR.  Do you want me to just push the fixes to your branch directly, or would you prefer to fix them yourself?  Just let me know either way.  Thanks!

bitterbark:
I’ll be in to look at these tomorrow, or feel free to push fixes if you know what it needs.

bitterbark:
So the failures look to be because I changed some of the defaults in the config...yaml file. 
This fails
    self.assertEqual(cam_case_name_val, "/some/where/you/want/to/have/regridded_files")
because I changed the default to 
 /glade/scratch/${user}/ADF/regrid

I believe these were agreed upon changes, so does that mean I should figure out how to change the test?

 It looks like several of the tests that failed ran into this particular variable; I'm not sure if there are other errors also.

nusbaume (approved):
Looks great, thanks @bitterbark!

PR #201, opened 2022-07-29
Generalize history file number (default h0). Only impacts timeseries generation.

bitterbark:
This works to generate the h* files, as long as the variables are listed appropriately. I'm happy to modify anything you want.


nusbaume (approved):
Thanks for the added input file flexibility @bitterbark!

Issue #157, opened 2022-04-28
Add ability to specify land/sea mask for plotted variables


nusbaume:
### New feature type

New plot and/or plot enhancement

### What is this new feature?

As brought up by @juliecaron, it might be useful to have the ability to specify whether a variable should be masked over land or ocean when plotting, possibly as another variable defaults attribute.

### Assistance required?

No, I will make a PR when the feature is ready

PR #202, opened 2022-08-04
Allow use of new ocean mask variable default

### LINKED ISSUE #190: Incorrect SSTs
### ADF run type

Model vs. Obs

### What happened?

I am looking at the SSTs on the table:(https://webext.cgd.ucar.edu/BWsc1850MOM/b.cesm3_cam058_mom_c.B1850WscMOM.ne30_L58_t061.009/atm/b.cesm3_cam058_mom_c.B1850WscMOM.ne30_L58_t061.009_1_10_vs_obs/html_table/amwg_table_b.cesm3_cam058_mom_c.B1850WscMOM.ne30_L58_t061.009.html)

The SSTs are 215K. It seems there is a bug.

SST | K | 215
-- | -- | --

### ADF Hash you are using

Ask Justin

Other (please explain below)

Other (please explain below)

### Extra info

@justin-richling ran the ADF. He can provide more info about hash, machine and env

### LINKED ISSUE #193: ADF tables may not be handling missing data properly
### ADF run type

Model vs. Model

### What happened?

Multiple variables listed in the generated tables appear to be pulling missing values into the calculations, which results in nonsensical statistics being displayed for those variables.  

The same thing appears to happen for the mean values displayed on certain plots, where variables that have a large number of missing values (e.g. `SST` has missing values over land) have incorrect mean values.  Thus this issue is likely related to issue #190 as well.

### ADF Hash you are using

Ask Justin

Other (please explain below)

Other (please explain below)

### Extra info

I am assuming it is for the latest version of the ADF, but @justin-richling will need to provide the specifics, as the issue was raised by @cecilehannay for ADF runs that Justin had put online.

### LINKED ISSUE #214: calc_cam_climo logic needs help
### ADF run type

Model vs. Obs

### What happened?

Note that this is a very unusual case: I'm trying to use ADF to calculate time series but not climatologies. Furthermore, the error occurs after the timeseries are generated, so it just means the script quits on an error, not that it doesn't do what I need. So this is not a high priority bug, but I figure I should file it because it isn't that likely to get exercised again!

When compare_obs = true,  diag_cam_baseline_climo:calc_cam_climo = None.
In lib/adf_diag.py, the logic to determine if climos should be calculated is not properly triggered:
> 	if self.get_cam_info('calc_cam_climo') or \
>            self.get_baseline_info('calc_cam_climo'):

tests positive and goes into trying to calculate the climos.

I tried to solve this but have just become more confused.  I read about [truthy and falsey](https://www.freecodecamp.org/news/truthy-and-falsy-values-in-python/) to come up with this test:
> print(f"got calc_cam_climo cam {self.get_cam_info('calc_cam_climo')} baseline {self.get_baseline_info('calc_cam_climo')}")
> bool_cam      = bool(self.get_cam_info('calc_cam_climo'))
> bool_baseline = bool(self.get_baseline_info('calc_cam_climo'))
> print(f"bool(cam) {bool_cam} (baseline) {bool_baseline}")

The results are 
> got calc_cam_climo cam [False] baseline None
> bool(cam) True (baseline) False

So the None evalues to False, which is good, we could just use bool. 
But the [False] evaluates to True???

### ADF Hash you are using

52cbe23

CISL machine

NPL (CISL machines only)

### Extra info

_No response_

juliecaron:
This PR adds a new `mask: "ocean"` option to the ADF variable defaults list that allows a user to have only values over the open ocean included in table calculations and plot generation, except for (currently) the special regional map generation.  This new mask is applied by default to `SST`, but can always be added to additional variables as well by default later.

Also, there are currently no "observations" of ocean fraction in the ADF observations collection, and thus for now the mask is only applied for model vs. model comparisons

Finally, it should be noted that in order for this new feature to be applied properly, the provided conda environment had to be updated, which brings the ADF environment closer in line with CISL's NPL envrionment.  This modification will also require a new ADF tag to be made after this PR is merged, so that a user can easily find the exact conda environment file they need.

Along with the mask addition, a bug was fixed in the calculation of climo files.

Fixes #190 
Fixes #193 
Fixes #214 
Addresses #157


justin-richling (approved):
This all looks pretty good, thanks for adding in some of the extra features too; the `add_diag_var` will come in handy.

You should be good to merge, thanks.

Issue #198, opened 2022-07-13
File glob not working for climo years formatted 00XX


justin-richling:
### ADF run type

Model vs. Model

### What happened?

Expected behavior
---

In the config yaml file when working with climo years in the format 00XX (ie 0001 or 0010, etc.) the glob should only grab the range of years specified by the ```start_year``` and ```end_year```.

Bug behavior
---

It seems for this climo year format for ```start_year``` and ```end_year```, the ADF is grabbing all available files.

Example
---

For test case ```b.cesm3_cam058_mom_c.B1850WscMOM.f09_L58_t061.011``` with h0 files located at:
```/glade/scratch/hannay/archive/b.cesm3_cam058_mom_c.B1850WscMOM.f09_L58_t061.011/atm/hist/```

for climo years:
```
start_year: 0001

end_year: 0009
```

The actual years grabbed when creating time series are 0001-0021
case ts files at: ```/glade/scratch/richling/adf-output/climo_yrs_bug/case-ts/```

Similarly for baseline case: ```b.cesm3_cam058_mom_c.B1850WscMOM.ne30_L58_t061.009``` at:
```/glade/scratch/hannay/archive/b.cesm3_cam058_mom_c.B1850WscMOM.ne30_L58_t061.009/atm/hist/```

with the same start and end climo years, the actual years grabbed are 0001-0063
baseline ts files at: ```/glade/scratch/richling/adf-output/climo_yrs_bug/baseline-ts/```

Tested config yaml
---
```/glade/u/home/richling/adf_config_files/tests/config_climo_yrs_bug.yaml```

### ADF Hash you are using

2ea95d0

### What machine were you running the ADF on?

CGD machine

### What python environment were you using?

NPL (CISL machines only)

### Extra info

This was run from one of my branches, as well as the main branch of the ADF. I have not checked if this only happens w/ 00XX formatted climo years. 

I will look into this and will request assistance if I can't find the problem.

justin-richling:
Closing this, issue has been taken care of in PR #205

Issue #166, opened 2022-05-12
Short Case Names and Titles


andrewgettelman:
### New feature type

New plot and/or plot enhancement

### What is this new feature?

The new web-sites and plots are fantastic!  Thanks @justin-richling (et al).

One note on the plots though: Some have no titles (zonal) and most of the plots (LatLon, Zonal, Polar) do not have the case name. The case name should also be more descriptive for the user. 

I would propose we add a 'case_text'  for each case (for obs it can just be 'Observations') that is short and can be used in plot titles (e.g. 'FWsc No CLUBB' or 'FWsc L58 001'. I guess this would be read in in the configuration file, and be part of the adf object, which can then just be used in each plot to construct titles. 

Happy to discuss how to do it ( I could try if I get guidance), and feel free to suggest another name for the case text....

### Assistance required?

Yes, I am requesting a new feature but won't have time to develop it

### Extra info

I'm happy to help implement the titles in some of the routines if I have time, but not sure I would know how to add something to the configuration file. Could get some help and try it though.

justin-richling:
Thanks for pointing this out @andrewgettelman, I also noticed there is a big discrepancy in titles throughout all the plot types. I agree the title should be more descriptive for the case name too (without having the whole name?) and we should probably have the variable and season info for the actual png images. 

I'd be happy to take this on if you can't, it seems pretty straight forward if it's just another variable the user assigns in the config file. This would give me good practice getting data from the yaml file to the python scripts.

Maybe we can chat more about the details at the next hackathon.

andrewgettelman:
Thanks @justin-richling. That would be great to have the help, and I agree this should be a simple test of going from config file --> adf object so that the case names are available to plotting routines. 

We do need to make sure it propagates to the 'multi-case' functionality (so the names can be a list). 

Thanks!

justin-richling:
Closing this, issue was fixed in PR #205

PR #205, opened 2022-08-30
Website update, plot cleanup, and climo years

justin-richling:
This PR will bring in the new website layout, which is still clunky, but will get us a working site for the time being.

This will also add option for case nicknames in config file. For now this only creates titles for plots, but can be expanded anywhere else a shortened name is desired. If left blank in the config file, the nickname will default to climo case name given.

Additionally, this will also add much needed plot details which will be consistent throughout into the actual image:

- case titles (used from the case nicknames introduced in this PR and in the config file)
- statistical values, ie mean, max, min
- explicit description of what the difference plot is (test minus baseline)
- change all instances of control to baseline for comparing case name
- add climo years for both cases

Likewise, this will also provide climo years in the header of each web page, as well as give climo year ranges for cases in newly formed directory (which goes into the diagnostics plot location from the config file).

Closes:
#198 (glob not getting climo years correctly)
#166 (case nicknames)


justin-richling:
Put in bug fixes for climo year generation if compared to obs.

This allows for climo years if not specified in config file (closes #149)

scripts/plotting/global_latlon_map.py
```diff
@@ -451,5 +497,15 @@ def _load_dataset(fils):
...
+def _get_climo_yrs(cam_ts_loc):
+    starting_location = Path(cam_ts_loc)
+    files_list = sorted(starting_location.glob('*nc'))
+    try:
+        syear = int(files_list[0].stem[-13:-9])
+        eyear = int(files_list[0].stem[-6:-2])
+    except:
+        print("Smoethign is borken...")
+    return syear, eyear
+
```
nusbaume, line 509: If we save the start and end years as part of the ADF object in `adf_info.py` then this function and related code isn't needed.

scripts/plotting/global_latlon_vect_map.py
```diff
@@ -449,5 +500,19 @@ def global_latlon_vect_map(adfobj):
...
+def _get_climo_yrs(cam_ts_loc):
+    starting_location = Path(cam_ts_loc)
+    files_list = sorted(starting_location.glob('*nc'))
+    try:
+        syear = int(files_list[0].stem[-13:-9])
+        eyear = int(files_list[0].stem[-6:-2])
+    except:
+        print("Smoethign is borken...")
+    return syear, eyear
+
```
nusbaume, line 516: If we save the start and end years as part of the ADF object in `adf_info.py` then this function and related code isn't needed.

scripts/plotting/meridional_mean.py
```diff
@@ -250,3 +294,18 @@ def _load_dataset(fils):
     else:
         sfil = str(fils[0])
         return xr.open_dataset(sfil)
+    #End if
+#End def
+
+def _get_climo_yrs(cam_ts_loc):
```
nusbaume, line 300: If we save the start and end years as part of the ADF object in `adf_info.py` then this function and related code isn't needed.

scripts/plotting/polar_map.py
```diff
@@ -366,5 +411,22 @@ def polar_map(adfobj):
 ##############
 #END OF `polar_map` function
 
+#########
+# Helpers
+#########
+
+def _get_climo_yrs(cam_ts_loc):
```
nusbaume, line 418: If we save the start and end years as part of the ADF object in `adf_info.py` then this function and related code isn't needed.

scripts/plotting/zonal_mean.py
```diff
@@ -271,3 +317,18 @@ def _load_dataset(fils):
     else:
         sfil = str(fils[0])
         return xr.open_dataset(sfil)
+    #End if
+#End def
+
+def _get_climo_yrs(cam_ts_loc):
```
nusbaume, line 323: If we save the start and end years as part of the ADF object in `adf_info.py` then this function and related code isn't needed.

nusbaume (changes requested):
Great work @justin-richling!  However, I do have some possibly major code change requests that might be good to implement, plus a bunch of smaller, less important modifications.  The two major requests (which are repeated below) are:

1.  Instead of calculating the start and end year in multiple locations, it would probably be better to do it once in `adf_info.py` and then save those lists as variables in the ADF object itself.  Just look at the various `@property` functions for an example, and of course I can help if you run into any problems or issues.

2.  It looks like you removed the use of `prep_contour_plot` in several of the plotting scripts.  I am not sure if it was on purpose, but in general I would recommend re-implementing the use of that function, even if you have to add more code to it.  This is because it helps greatly reduce the total amount of code in `plotting_functions.py`, and will help as we continue to modularize the ADF's plotting abilities.  Again, if you have any questions or concerns just let me know.

Thanks!

lib/adf_diag.py
```diff
@@ -136,7 +136,6 @@ class AdfDiag(AdfWeb):
 
     """
     Main ADF diagnostics object.
-
```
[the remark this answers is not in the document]
justin-richling: Yeah, that's weird. Not sure how that happened.

lib/adf_web.py
```diff
@@ -343,13 +332,59 @@ def create_website(self):
...
+        for case_idx, case_name in enumerate(case_names):
+                
+            if (syear_cases[case_idx] and eyear_cases[case_idx]) == None:
+                print(f"No given climo years for {case_name}...")
+                starting_location = Path(cam_ts_locs[case_idx])
+                files_list = sorted(starting_location.glob('*nc'))
+                #This assumes CAM file names stay with this convention
+                #Better way to do this?
+                syear_cases[case_idx] = int(files_list[0].stem[-13:-9])
+                eyear_cases[case_idx] = int(files_list[0].stem[-6:-2])
```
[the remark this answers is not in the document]
justin-richling: This is what I ultimately envisioned, but wasn't sure of the best way. Thanks!

lib/plotting_functions.py
```diff
@@ -590,16 +634,73 @@ def plot_map_and_save(wks, mdlfld, obsfld, diffld, **kwargs):
...
+        if ((minval < 0) and (0 < maxval)) and mplv > 2:
+            norm1 = normfunc(vmin=minval, vmax=maxval, vcenter=0.0)
+        else:
+            norm1 = mpl.colors.Normalize(vmin=minval, vmax=maxval)
+
+    # Difference options -- Check in kwargs for colormap and levels
+    if "diff_colormap" in kwargs:
+        cmapdiff = kwargs["diff_colormap"]
+    else:
+        cmapdiff = 'coolwarm'
```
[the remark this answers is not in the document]
justin-richling: I'm guessing my code was old enough it didn't have these changes in it yet.

lib/plotting_functions.py
```diff
@@ -373,20 +392,22 @@ def make_polar_plot(wks,d1:xr.DataArray, d2:xr.DataArray, difference:Optional[xr
...
     kwargs -> optional dictionary of plotting options
              ** Expecting this to be a variable-specific section,
                 possibly provided by an ADF Variable Defaults YAML file.**
-
     """
 
     # specify the central longitude for the plot:
     cent_long = kwargs.get('central_longitude', 180)
 
     # generate progjection:
```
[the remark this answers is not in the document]
justin-richling: > typoe

Was this a typo too, haha

lib/adf_info.py
```diff
@@ -128,8 +140,24 @@ def __init__(self, config_file, debug=False):
             #Set data name to baseline case name:
             data_name = self.get_baseline_info('cam_case_name', required=True)
 
-        #End if
+            #Attempt to grab baseline start_years (not currently required):
+            syear_baseline = self.get_baseline_info('start_year')
+            eyear_baseline = self.get_baseline_info('end_year')
+
+            if syear_baseline and eyear_baseline is None:
```
justin-richling, line 147: @nusbaume I'm still having some issues if I don't put parenthesis around the ```syear_baseline and eyear_baseline``` part of the if-statements here and on lines 179 and 189. The code skips over these without error without parenthesis, but it seems having the parenthesis was given pylint problems, correct?

justin-richling:
I've tackled most of the comments you provided (thanks for going through the code so meticulously). Other than my comment about parenthesis on the if-statements in ```adf_info.py```, I think this is close to being done. When you get around to looking at it again, let me know.

Thanks!

lib/website_templates/template.html
```diff
@@ -1,72 +1,82 @@
...
+
 <!DOCTYPE html>
 <html>
   <head>
     <title>ADF {{var_title}}</title>
     <link rel="stylesheet" href="../templates/adf_diag.css">
   </head>
   <body>
+    <div class="header">
+      <div id="screenFiller" style="background:url(https://webext.cgd.ucar.edu/AMWG/f.e21.FHIST_SSP370.f09_f09_mg17.CESM2-CLM4p5.LongRun001-f.e21.FHIST_SSP370.f09_f09_mg17.CESM2-CLM5.LongRun001/f.e21.FHIST_SSP370.f09_f09_mg17.CESM2-CLM4p5.LongRun001-f.e21.FHIST_SSP370.f09_f09_mg17.CESM2-CLM5.LongRun001/images/NCAR.gif) top right no-repeat; border:5px solid #555;">
```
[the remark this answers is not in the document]
nusbaume: Given the small size of the gif file I went ahead and added it as part of the ADF itself.  That way if that particular URL is ever deleted the ADF will still work as expected.

lib/adf_info.py
nusbaume: I ended up modifying this logic so that it checks for the start year and end year independently, so luckily this issue should no longer be a problem.

nusbaume:
Found a few bugs while integration testing, as well as some clean-up opportunities, but now everything looks great.  Thanks for getting the new website developed @justin-richling!

nusbaume (approved):

PR #219, opened 2022-11-09
Update float format in tables

### LINKED ISSUE #217: AMWG Tables aren't getting formatted correctly in html file creation
### ADF run type

Model vs. Model

### What happened?

The AMWG tables for the website are not getting formatted correctly when going from pandas DataFrame to HTML.

Currently the ADF is formatting all values in the ```to_html()``` call to pandas data frame being brought over in the ```web_data``` variable in ```adf_web.py```.  However it loses precision for numbers larger than two digits. 

For example, if a mean value is 288.156, the table formats it to 288, but if the mean is 44.7, it retains that precision. Right now the ```float_format``` value that goes into the ```to_html()``` call is ```.3g``` meaning it retains only 3 important digits, hence why 288.15 returns 288 and something like 0.000815 would retain 0.000815

#### Example:

adf_0 3g_format

### ADF Hash you are using

2b3eca8

CISL machine, CGD machine

NPL (CISL machines only)

### Extra info

I am testing a solution that finds what range for the mean and statistical table values and sets float format based on those in ```amwg_table.py``` before it gets to ```adf_web.py```.

justin-richling:
This will change the formatting in the pandas dataframe as well as in the html file generated from the dataframe.

#### Updates: 
The formatting in the original dataframe (```amwg_table.py```) is conditionally based on float value (greater or less than 1) with less than one getting 3 decimal place retention (```.3f```) and greater than one getting 3 sig decimal figs (```.3g```). This will also update the difference values in the comparison table.

The formatting in the html table from dataframe (```adf_web.py```) will update to ```6g``` -> 6 sig figs to catch the range of values (1e6 to 1e-6).

A function for generating stats from the data has been created as well, which will replace hard coded section. 

Closes #217

### DIFF
--- a/scripts/analysis/amwg_table.py
+++ b/scripts/analysis/amwg_table.py
@@ -1,3 +1,23 @@
+import numpy as np
+import xarray as xr
+from pathlib import Path
+import warnings  # use to warn user about missing files.
+
+try:
+    import scipy.stats as stats # for easy linear regression and testing
+except ImportError:
+    print("Scipy module does not exist in python path, but is needed for amwg_table.")
+    print("Please install module, e.g. 'pip install scipy'.")
+
+#Import "special" modules:
+try:
+    import pandas as pd
+except ImportError:
+    print("Pandas module does not exist in python path, but is needed for amwg_table.")
+    print("Please install module, e.g. 'pip install pandas'.")
+    sys.exit(1)
+
+
 def amwg_table(adf):
 
     """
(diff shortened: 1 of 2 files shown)


scripts/analysis/amwg_table.py
```diff
@@ -1,3 +1,23 @@
+import numpy as np
+import xarray as xr
+from pathlib import Path
+import warnings  # use to warn user about missing files.
+
+try:
+    import scipy.stats as stats # for easy linear regression and testing
+except ImportError:
+    print("Scipy module does not exist in python path, but is needed for amwg_table.")
+    print("Please install module, e.g. 'pip install scipy'.")
```
nusbaume, line 10: I think you'll want to include `sys.exit(1)` here, as otherwise `amwg_table` will still attempt to run even though `scipy` isn't present.

scripts/analysis/amwg_table.py
```diff
@@ -1,3 +1,23 @@
...
+
+#Import "special" modules:
```
nusbaume, line 12: I would put this comment above the `scipy.stats` import, as it is also a "special" module.

nusbaume (approved):
Looks good!  Just a couple small recommendations.  Also I should note that I haven't done any testing on my end, but I would be surprised if this would cause problems outside of the tables themselves, so if these mods work for you then feel free to merge this PR.  Thanks!

PR #221, opened 2022-12-01
Remake index.html every time

justin-richling:
** Impacts only for website generation **

If changing ADF after an initial run, the user currently has to delete the index.html before re-running ADF to make an updated index.html file in order for the changes to be reflected on the website.

This is a huge gotcha that would be good to eliminate and won't cost much in terms of computation. It will guarantee all changes to subsequent ADF runs will be reflected in the webpage.


nusbaume (approved):
Looks good, thanks Justin!

PR #224, opened 2022-12-13
Vector plot fixes

### LINKED ISSUE #220: user request to change sign of wind stress vectors
### New feature type

New plot and/or plot enhancement

### What is this new feature?

Per the discussion on this CESM forum thread:

https://bb.cgd.ucar.edu/cesm/threads/interesting-taux-tauy-in-cesm2-1-b1850cmip6.7805/#post-46945

I wonder whether the scale factor on TAUX and TAUY need to be flipped to get the expected vector plot? 

### Assistance required?

### Extra info

I think we should get some additional input on this to figure out what the preferred behavior is.

### LINKED ISSUE #223: Vectors offset from correct spatial location in vector maps
### ADF run type

Model vs. Obs

### What happened?

It appears that in the vector maps all of the actual vector arrows are meridionally offset from their correct geographic location by the `central_longitude` value, which implies that the vectors or "quivers" aren't seeing the central longitude change.  

An example can be seen below, where the vectors that should represent the Indian Monsoon are instead way out in the Pacific:

Wind_850hpa_JJA_LatLon_Vector_Mean

### ADF Hash you are using

6c8bcd8

CISL machine

NPL (CISL machines only)

### Extra info

@brianpm @justin-richling I am pretty sure the issue is in the use of matplotlib in `plot_map_vect_and_save` within `lib/plotting_functions.py`.  Since you both have more matplotlib expertise than me, I was wondering if at some point you could take a look and see if there is anything obviously wrong?  Thanks a bunch!

nusbaume:
Fixes #223 
Fixes #220

Also update the Github checkout versions for the linting (pylint) workflow.

@brianpm I went ahead and changed the sign of the wind stress vectors in this PR as discussed in #220 , as I personally agreed in that I usually think of the surface "stress" as going in the direction of the flow (as opposed to "drag").  Of course if you think this isn't a good idea at this point I can always revert the change.  Just let me know!

### DIFF
--- a/.github/workflows/ADF_linting.yaml
+++ b/.github/workflows/ADF_linting.yaml
@@ -15,10 +15,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     # acquire github action routines
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
     # acquire specific version of python
     - name: Set up Python 3.8
-      uses: actions/setup-python@v1
+      uses: actions/setup-python@v4
       with:
         python-version: '3.8' # Use python 3.8
     # install required python packages
--- a/lib/adf_variable_defaults.yaml
+++ b/lib/adf_variable_defaults.yaml
@@ -357,7 +357,7 @@ CCN3:
   category: "Clouds"
 
 #+++++++++++++++++
-# Category: CLUBB 
+# Category: CLUBB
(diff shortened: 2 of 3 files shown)


justin-richling (approved):
Changes look good to me @nusbaume, thanks!

PR #229, opened 2023-02-07
Enhance multiple case diagnostics

justin-richling:
Add multi-case diagnostics and webpages:

- Add multi-case plot section and case nicknames to config yaml
- Update all website templates (add new for multi-case)
- Update web generation
- Add multi-case subplots and all case comp tables
- Designate hub for all test cases individual vs baseline ADF analysis

### DIFF
--- a/lib/adf_web.py
+++ b/lib/adf_web.py
@@ -24,6 +24,7 @@ import os
 import os.path
 
 from pathlib import Path
+from collections import OrderedDict
 
 #+++++++++++++++++++++++++++++++++++++++++++++++++
 #import non-standard python modules, including ADF
@@ -56,6 +57,7 @@ class _WebData:
     """
 
     def __init__(self, web_data, web_name, case_name,
+                 plot_ext = None,
                  category = None,
                  season = None,
                  plot_type = "Special",
@@ -71,6 +73,7 @@ class _WebData:
         self.category   = category
         self.season     = season
         self.plot_type  = plot_type
+        self.plot_ext   = plot_ext
         self.data_frame = data_frame
         self.html_file  = html_file
         self.asset_path = asset_path
@@ -179,7 +182,8 @@ class AdfWeb(AdfObs):
 
     #########
 
-    def add_website_data(self, web_data, web_name, case_name,
+    def add_website_data(self, web_data, web_name, case_name, 
+                         plot_ext = None,
                          category = None,
                          season = None,
                          plot_type = "Special",
@@ -217,6 +221,7 @@ class AdfWeb(AdfObs):
         #Initialize Pandas data frame logical:
         data_frame = False
 
+        html_file = []
         #Check that the web_data is either a path
         #or a pandas dataframe:
         try:
@@ -271,7 +276,11 @@ class AdfWeb(AdfObs):
 
             #If multi-case, then save under the "multi-case" directory:
             if self.num_cases > 1:
-                html_file = self.__case_web_paths['multi-case']["table_pages_dir"] / html_name
+                #Avoid collecting individual case comparison tables
+                #Hacky - could probably use an update eventually - JR
+                if web_name != "case_comparison":
+                    html_file.append(self.__case_web_paths['multi-case']["table_pages_dir"] / html_name)
+                html_file.append(self.__case_web_paths[case_name]["table_pages_dir"] / html_name)
             else:
                 html_file = self.__case_web_paths[case_name]["table_pages_dir"] / html_name
             #End if
@@ -283,7 +292,7 @@ class AdfWeb(AdfObs):
         #End if
 
         #Initialize web data object:
-        web_data = _WebData(web_data, web_name, case_name,
+        web_data = _WebData(web_data, web_name, case_name, plot_ext,
                             category = category,
                             season = season,
                             plot_type = plot_type,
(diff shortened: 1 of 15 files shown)


scripts/plotting/global_latlon_map.py
```diff
@@ -65,6 +66,15 @@ def global_latlon_map(adfobj):
     #CAM simulation variables (this is always assumed to be a list):
     case_names = adfobj.get_cam_info("cam_case_name", required=True)
 
+    if len(case_names) > 1:
+        #Check if multi-plots are desired from yaml file
+        if adfobj.read_config_var('multi_case_plots'):
+            multi_plots = True
+            if multi_plots:
```
nusbaume, line 73: I don't think this if-statement is needed given that you just set `multi_plots` to `True` (so the statement will always pass).

scripts/plotting/global_latlon_map.py
```diff
@@ -65,6 +66,15 @@ def global_latlon_map(adfobj):
     #CAM simulation variables (this is always assumed to be a list):
     case_names = adfobj.get_cam_info("cam_case_name", required=True)
 
+    if len(case_names) > 1:
+        #Check if multi-plots are desired from yaml file
+        if adfobj.read_config_var('multi_case_plots'):
```
nusbaume, line 71: Should you check that `global_latlon_map` is present as an option under `multi_case_plots` here as well?

scripts/plotting/global_latlon_map.py
```diff
@@ -141,9 +151,14 @@ def global_latlon_map(adfobj):
                "MAM": [3, 4, 5],
                "SON": [9, 10, 11]
                }
-
+    
     # probably want to do this one variable at a time:
     for var in var_list:
+        if multi_plots:
+            for multi_var in adfobj.read_config_var('multi_case_plots')["global_latlon_map"]:
```
nusbaume, line 158: Could you use `get_multi_case_info` here instead of `read_config_var`?

scripts/plotting/global_latlon_map.py
```diff
@@ -266,6 +286,10 @@ def global_latlon_map(adfobj):
                         oseasons = {}
                         dseasons = {} # hold the differences
 
+                        #Initialize Ordered Dictionary for variable:
+                        #if case_name not in restom_dict:
+                        #restom_dict[case_name] = OrderedDict()
```
nusbaume, line 291: Delete commented-out code?

nusbaume (changes requested):
Great Work Justin!  I have lots of small suggestions and questions, but they should all be relatively minor, as the overall structure looked good to me.

The only somewhat large change might be to manage the setting of the case nickname list in `adf_info.py` instead of in the various table and plotting scripts.  That way any future scripts can use the nicknames directly, instead of having to do their own checking.

Anyways, please let me know if you have any questions or concerns, or need any help with any of my requests.  Thanks!

scripts/analysis/amwg_table.py
```diff
@@ -394,11 +420,60 @@ def _df_comp_table(adf, output_location, case_names):
...
+    for i,val in enumerate(csv_locs[:-1]): 
+        case = str(val)+f"/amwg_table_{case_names[i]}.csv"
+        df_case = pd.read_csv(case)
+        
+        if test_nicknames[i] == case_names[i]:
+            df_comp[['variable','unit',f"case {i+1}"]] = df_case[['variable','unit','mean']]
+            cols_comp.append(f"case {i+1}")
+        else:
+            df_comp[['variable','unit',f"{test_nicknames[i]}"]] = df_case[['variable','unit','mean']]
+            cols_comp.append(test_nicknames[i])
```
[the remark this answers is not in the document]
justin-richling: Essentially this is to check if the case name is used (or whether the test_nicknames are not assigned and defaults to case name) to populate table header. If so because the case name is generally pretty long, it just shortens the header to "case<#> " to keep the table from being very long, especially if the number of cases is getting large (~ >5).

Otherwise, if actual nicknames are used (assuming they aren't nearly as long as raw case names), the column header uses them.

lib/adf_web.py
```diff
@@ -443,9 +465,34 @@ def create_website(self):
...
+
+            #Also add path to case_sites dictionary:
+            #loop over cases:
+            for idx, case_name in enumerate(case_names):
+                #Check if case name is present in plot
+                if case_name in self.__case_web_paths:
+                    #Add path to case_sites dictionary:
+                    case_dir_ext = f"{case_name}_{syear_cases[idx]}_{eyear_cases[idx]}"
+                    base_dir_ext = f"{data_name}_{syear_baseline}_{eyear_baseline}"
+                    case_sites[case_name] = [os.path.join(os.curdir,
```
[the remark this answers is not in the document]
nusbaume: Actually nevermind, I forgot that `os.curdir` gives the relative path (e.g. `.`) and not the absolute path, which is what is needed for the website links.  So I think this code is good as-is, and will go ahead and mark this issue as resolved.

lib/adf_web.py
```diff
@@ -717,52 +840,301 @@ def create_website(self):
...
+                        #End with
+
+                    #if web_data.case in case_names:
+                    if web_data.case != data_name:
+                        table_html = web_data.data.to_html(index=False, border=1, justify='center',
+                                                            float_format='{:6g}'.format)
+
+                        #Construct amwg_table.html
+                        table_keys = [web_data.case,data_name,"case_comparison"]
+                        case_table_dict = {key: multi_table_html_info[key] for key in table_keys}
```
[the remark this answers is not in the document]
justin-richling: Apparently not, thanks for catching this

PR #230, opened 2023-02-16
Ncrcat fixes and make timeseries generation compatible with all model components & filenames

### LINKED ISSUE #222: timeseries creation adds incorrect cell_method = mean attribute to fields
### ADF run type

Model vs. Model

### What happened?

I have history files without cell_method = mean, but when I create timeseries files with ADF, they gain that attribute. 

### ADF Hash you are using

52cbe23

CISL machine

NPL (CISL machines only)

### Extra info

History files in /glade/p/cgd/amp/bundy/archive/bhist.e213cosp217.f09_g17.mdtf_v3.2009/atm/hist

First file has mean in the var attributes, second doesn't:
> ncdump -h bhist.e213cosp217.f09_g17.mdtf_v3.2009.cam.h3.2009-01-01-21600.nc | grep PRECC
> 	float PRECC(time, lat, lon) ;
> 		PRECC:units = "m/s" ;
> 		PRECC:long_name = "Convective precipitation rate (liq + ice)" ;
> 		PRECC:cell_methods = "time: mean" ;

> ncdump -h bhist.e213cosp217.f09_g17.mdtf_v3.2009.cam.h2.2009-01-01-21600.nc | grep PRECC
> 	float PRECC(time, lat, lon) ;
> 		PRECC:units = "m/s" ;
> 		PRECC:long_name = "Convective precipitation rate (liq + ice)" ;
> 

But resulting ts files have it in both:

> set dir = /glade/campaign/cgd/amp/bundy/mdtf/cesm_mdtfv3_timeslice_orig/bhist.e213cosp217.f09_g17.mdtf_v3.2009/atm/ts
> /glade/u/home/bundy/diag/ADF_top/ADF> ncdump -h $dir /bhist.e213cosp217.f09_g17.mdtf_v3.2009.h2.PRECC.2009010121600-2013010121600.nc | grep cell
> 		time:cell_methods = "time: mean" ;
> 		time_bnds:cell_methods = "time: mean" ;
> 		PRECC:cell_methods = "time: mean" ;
> /glade/u/home/bundy/diag/ADF_top/ADF> ncdump -h $dir /bhist.e213cosp217.f09_g17.mdtf_v3.2009.h3.PRECC.2009010121600-2013010121600.nc | grep cell
> 		time:cell_methods = "time: mean" ;
> 		time_bnds:cell_methods = "time: mean" ;
> 		PRECC:cell_methods = "time: mean" ;

bitterbark:
1. Modify timeseries generation so that it works on all model components and all history file numbers by replacing hist_num with hist_str (eg cam.h1 or pop.h.ecosys.nday1). Default if missing is cam.h0 for backward compatibility
2. Modify ncrcat call to include --no_cll_mth option, which fixes [Issue 222](https://github.com/NCAR/ADF/issues/222) until the nco's are fixed and updated on CISL machines
3. Modify ncrcat call so that it doesn't require PS to be in the same file as any 3D variable. Generates a warning but continues
4. Modify ncrcat call to add strings as necessary so that changes to the base call or flags don't have to be implemented in multiple places

Fixes #222


config_cam_baseline_example.yaml
```diff
@@ -64,6 +64,10 @@ user: 'USER-NAME-NOT-SET'
 #This first set of variables specify basic info used by all diagnostic runs:
 diag_basic_info:
 
+    #History file string to match (eg. cam.h0 or ocn.pop.h.ecosys.nday1)
+    #Default: cam.h0. Only affects timeseries as everything else uses timeseries
```
nusbaume, line 68: Should we add a warning here that there shouldn't be a trailing `.` at the end of the string?  It looks like the file globbing step includes the `.`, and so if someone has it in the string itself the file globbing will likely fail.  We also probably want to remove it from the example in the comment as well:

 ```suggestion
     #Default: cam.h0 Only affects timeseries as everything else uses timeseries 
```

File: lib/adf_diag.py [hunk 2]
```diff
@@ -533,30 +532,40 @@ def call_ncrcat(cmd):
...
-                               hist_files + ["-o", ts_outfil_str]
-                    elif vert_coord_type == "height":
+                    #PS might be in a different history file. If so, continue without error.
+                    ncrcat_var_list = ncrcat_var_list + f",hyam,hybm,hyai,hybi"
+
+                    if 'PS' in hist_file_ds[var].dims:  
+                        ncrcat_var_list = ncrcat_var_list + ",PS"
+                    else:
+                        print(f"WARNING: PS not found in history file. It might be needed at some point.")
+                    #End if
```
nusbaume, line 553: Should we have a check here to see if the `hist_str` variable contains the word `cam` at the start?  Otherwise someone processing, say, clm or pop files is going to get this warning constantly.

lib/adf_diag.py
```diff
@@ -533,30 +532,40 @@ def call_ncrcat(cmd):
...
+                    if vert_coord_type == "height":
                         #Adding PMID here works, but significantly increases
                         #the storage (disk usage) requirements of the ADF.
                         #This can be alleviated in the future by figuring out
                         #a way to determine all of the regridding targets at
                         #the start of the ADF run, and then regridding a single
                         #PMID file to each one of those targets separately. -JN
-                        cmd = ["ncrcat", "-O", "-4", "-h", "-v", f"{var},PMID,PS"] + \
-                                hist_files + ["-o", ts_outfil_str]
+                        ncrcat_var_list = ncrcat_var_list + ",PMID"
```
nusbaume, line 562: Should we have a check here to see if `PMID` is actually in the history file, and if not print a warning, similar to what was done for `PS`?

nusbaume (changes requested):
Thanks for the bug fix and new functionality!  Just had a few hopefully minor requests.  Also if you would like me to manage those modifications then just let me know.  Thanks!

config_cam_baseline_example.yaml
bitterbark: Adding the comment and removing the '.' from the example. It seems we should really test if it ends with a '.' and if so, not add one, but is it okay to do that later?

bitterbark:
I made changes, committed and pushed them; thought they would come in here automatically.

nusbaume:
Just a FYI to my future self that I updated the conda environment in this PR to add the missing matplotlib version number from the `adf_v0.1` version, and so a new `adf_v0.11` tag will need to be made from this PR.

nusbaume (approved):
Everything looks great now, thanks @bitterbark!

Issue #228, opened 2023-02-02
Code fails on assumption of 1 clim file per variable name not accounting for components (i.e precc, precl vs prect) or comparable variables (i.e ts vs sst)


juliecaron:
### ADF run type

Model vs. Obs

### What happened?

There are a couple of layers to this. But the source of the error is in regrid_and_vert_interp.py, in this section:
                    if len(mclim_fils) > 1:
                        #Combine all cam files together into a single data set:                                                                                                          
                        mclim_ds = xr.open_mfdataset(mclim_fils, combine='by_coords')
                    else:
                        #Open single file as new xsarray dataset:                                                                                                                        
                        mclim_ds = xr.open_dataset(mclim_fils[0])
                    #End if                                                               

First, if one wants to use the yaml file for the default AMWG behavior, the code currently assumes that "PRECT" is there and does not check for PRECL and PRECC. but i believe the rest of the code can handle PRECL and PRECC (presumably the timeseries file creation combines the terms into PRECT, but i don't remember for sure), but in this case I am using timeseries files already generated offline, not via ADF.  

Second, if one variable is not there (i.e. SST), it is also not trying to find TS and mask it. It looks for an SST clim file and fails.

So currently, in these instances, the code errors out because the len(mclim_fils) is actually zero. but the code snippet above assumes that it will only ever be mor than 1 or  1 (because the else has the action of subscripting assuming length 1). It'd be great to check if it's zero, and  then give a clear error message "No clim file for variable X found. " . Additionally, it'd be great if it would check for comparable variables and combine (in case of prect) and/or use an alternate like TS for SST. and let the user know with "Using Y instead" or "Can't find alternate variable. Skipping X."

### ADF Hash you are using

bcf3630

### What machine were you running the ADF on?

CISL machine

### What python environment were you using?

NPL (CISL machines only)

justin-richling:
Close via PR #260

PR #232, opened 2023-02-28
Add time series plots

justin-richling:
This will add time series plots to scripts directory and will have the option in the config yaml file to plot only certain variables, as opposed to all declared in `diag_var_list`

The time series variables get declared in `timeseries_var_list` and must also be in `diag_var_list`. It's not ideal and has some redundancy, but this can be addressed in the future if desired.

* Note: RESTOM can be declared in the `timeseries_var_list` w/o its constituents so long as they (FSNT and FLNT) are in the `diag_var_list`. 

It is also set up to plot any derived quantity, the user just needs to create functions to do so. The RESTOM calculation can be used a rough guide.

The script also looks at any variable with vertical levels and ignores it (similar to AMWG tables). In the future it might be nice to have an option to plot variables with vertical levels if single levels are designated, but currently it just skips them. 

The time series also calculates seasonally weighted averages based on month length.
* Note: RESTOM currently is only calculated annually, not seasonally.

There is also a flag in `adf_variable_defaults.yaml` to set a rolling 5-year mean. Currently RESTOM is defaulted to this, but can be changed.

In addition, some minor cleanup included:
- Add RESTOM config parameters in `adf_variable_defaults.yaml`
- Clean up amwg tables for derived quantities and move RESTOM to top of table
- Block out missing seasons on website
- Fix error if missing mclim_fils in regrid_and_vert_interp
       * Part of Issue [228](https://github.com/NCAR/ADF/issues/228)

### DIFF
--- a/scripts/analysis/amwg_table.py
+++ b/scripts/analysis/amwg_table.py
@@ -1,5 +1,6 @@
 import numpy as np
 import xarray as xr
+import warnings
 import sys
 from pathlib import Path
 import warnings  # use to warn user about missing files.
@@ -130,6 +131,7 @@ def amwg_table(adf):
 
     #CAM simulation variables (these quantities are always lists):
     case_names    = adf.get_cam_info("cam_case_name", required=True)
+    
     input_ts_locs = adf.get_cam_info("cam_ts_loc", required=True)
 
     #Check if a baseline simulation is also being used:
@@ -149,10 +151,13 @@ def amwg_table(adf):
         #Save the baseline to the first case's plots directory:
         output_locs.append(output_locs[0])
 
-    #-----------------------------------------
+    #Declare any derived quantities here:
+    derived_vars = {"RESTOM":["FSNT", "FLNT"]}
+    derived_list = [item for sublist in list(derived_vars.values()) for item in sublist]
+
     #Create (empty) dictionary to use for the
-    #residual top of model (RESTOM) radiation calculation:
-    restom_dict = {}
+    #derived calculations (inititally RESTOM radiation):
+    derived_dict = {}
 
     #Loop over CAM cases:
     for case_idx, case_name in enumerate(case_names):
@@ -183,7 +188,7 @@ def amwg_table(adf):
         #End if
 
         #Save case name as a new key in the RESTOM dictonary:
-        restom_dict[case_name] = {}
+        derived_dict[case_name] = {}
 
         #Create/reset new variable that potentially stores the re-gridded
         #ocean fraction xarray data-array:
@@ -270,13 +275,9 @@ def amwg_table(adf):
                 # Note: we should be able to handle (lat, lon) or (ncol,) cases, at least
                 data = _spatial_average(data)  # changes data "in place"
 
-            #Add necessary data for RESTOM calcs below
-            if var == "FLNT":
-                restom_dict[case_name][var] = data
-                #Copy units for RESTOM as well:
-                restom_units = unit_str
-            if var == "FSNT":
-                restom_dict[case_name][var] = data
+            #Add necessary data for derived (RESTOM) calcs below
+            if var in derived_list:
+                derived_dict[case_name][var] = [data, unit_str]
 
             # In order to get correct statistics, average to annual or seasonal
             data = data.groupby('time.year').mean(dim='time') # this should be fast b/c time series should be in memory
@@ -305,46 +306,29 @@ def amwg_table(adf):
 
         #End of var_list loop
         #--------------------
-
+        
+        #Space for derived quantities (ie RESTOM, etc.)
+        #----------------------------------------------
+        
+        # RESTOM
+        #-------
         if "FSNT" and "FLNT" in var_list:
-            #RESTOM Calcs
-            var = "RESTOM" #RESTOM = FSNT-FLNT
-            print(f"\t - Variable '{var}' being added to table")
-            data = restom_dict[case_name]["FSNT"] - restom_dict[case_name]["FLNT"]
-            # In order to get correct statistics, average to annual or seasonal
-            data = data.groupby('time.year').mean(dim='time') # this should be fast b/c time series should be in memory
-                                                                # NOTE: data will now have a 'year' dimension instead of 'time'
-            # These get written to our output file:
-            stats_list = _get_row_vals(data)
-            row_values = [var, restom_units] + stats_list
-            # col (column) values declared above
-
-            # Format entries:
-            dfentries = {c:[row_values[i]] for i,c in enumerate(cols)}
-
-            # Add entries to Pandas structure:
-            df = pd.DataFrame(dfentries)
-
-            # Check if the output CSV file exists,
-            # if so, then append to it:
-            if output_csv_file.is_file():
-                df.to_csv(output_csv_file, mode='a', header=False, index=False)
-            else:
-                df.to_csv(output_csv_file, header=cols, index=False)
-            #End if
-
+            _derive_restom(case_name, derived_dict, output_csv_file, cols)
         else:
             #Print message to debug log:
             adf.debug_log("RESTOM not calculated because FSNT and/or FLNT variables not in dataset")
         #End if
+        #End RESTOM table addition
+
+        #Other derived quantity
+        #-------
+        #End Other table addition
 
         # last step is to add table dataframe to website (if enabled):
         table_df = pd.read_csv(output_csv_file)
         adf.add_website_data(table_df, case_name, case_name, plot_type="Tables")
-
+        #End derived quantities
     #End of model case loop
-    #----------------------
-
 
     #Notify user that script has ended:
     print("  ...AMWG variable table has been generated successfully.")
@@ -369,16 +353,17 @@ def amwg_table(adf):
 ##################
 
 def _load_data(dataloc, varname):
-    import xarray as xr
     ds = xr.open_dataset(dataloc)
     return ds[varname]
 
 #####
 
 def _spatial_average(indata):
-    import xarray as xr
-    import numpy as np
-    import warnings
+    """
+    asdadfd
+    -----
+        - sdgff
+    """
 
     #Make sure there is no veritcal level dimension:
     assert 'lev' not in indata.coords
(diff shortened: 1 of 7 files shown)


scripts/plotting/time_series.py
```diff
@@ -0,0 +1,630 @@
...
+        #End if (min/max)
+
+    #End for (vars)
+    #End ANN plots
+
+    #Seasonally weighted plots
+    # - DJF, MAM, JJA, SON
+    ##########################
+    print("\n  Generating time series for seasonally weighted...")
+    for var in ts_var_list_s:
```
nusbaume, line 312: It feels like a lot of the code inside this loop is the same as the `ANN` code above.  Maybe it would be worth making a Github issue so that in the future one of us can remember to try and find a clever way to combine the two code blocks?

scripts/plotting/time_series.py
```diff
@@ -0,0 +1,630 @@
...
+                    #Grab mins/maxes
+                    y_mins.append(np.nanmin(vals[var][case_name][season]))
+                    y_maxs.append(np.nanmax(vals[var][case_name][season]))
+                #End for (cases)
+
+                #Set Main title for subplots:
+                ax.set_title(f"Time Series {title_var}: {var} - {season}",loc="left")
+                
+                if rolling:
+                    ax.set_title(f"5-yr rolling average",loc="right")
```
nusbaume, line 361: It doesn't look like a rolling average is applied to the seasonal data.  Should it be?  If not, then I would probably remove the references to the `rolling` variable both here and above.

scripts/plotting/time_series.py
```diff
@@ -0,0 +1,630 @@
...
+
+    data = ts_ds[var].squeeze()
+    month_length = data.time.dt.days_in_month
+    unit = data.units
+
+    #global weighting
+    w = np.cos(np.radians(data.lat))
+    avg = data.weighted(w).mean(dim=("lat","lon"))
+
+    yrs = np.unique([str(val.item().timetuple().tm_year).zfill(4) for _,val in enumerate(ts_ds["time"])])
```
nusbaume, line 434: Can't you just loop over the values directly instead of using `enumerate`?

```suggestion
    yrs = np.unique([str(val.item().timetuple().tm_year).zfill(4) for val in ts_ds["time"]])
```

scripts/plotting/time_series.py
```diff
@@ -0,0 +1,630 @@
...
+    del_s = []
+
+    vals = OrderedDict()
+    yrs = {}
+    units = {}
+    for var in ts_var_list:
+        if var not in vals:
+            vals[var] = OrderedDict()
+        for case_idx, case_name in enumerate(all_case_names):
+            fils = sorted(list(Path(case_ts_locs[case_idx]).glob(f"*{var}.*.nc")))
```
nusbaume, line 474: You don't need the `list` function here as `sorted` will automatically convert it to a list:

```suggestion
            fils = sorted(Path(case_ts_locs[case_idx]).glob(f"*{var}.*.nc"))
```

scripts/plotting/time_series.py
```diff
@@ -0,0 +1,630 @@
...
+def _get_seasonal_data(ts_var_list, all_case_names, case_ts_locs):
+    """
+    Gather seasonally weighted data
+    -----
+        - 
+    """
+
+    #Keep track of variables that have vertical levels for now
+    #There is probably a better way to ignore these vars - JR
+    del_s = []
```
nusbaume, line 465: It would probably be slightly more performant if `del_s` was a set instead of a list, since the only thing it is being used for is to check if a variable is a member:

```suggestion
    del_s = set()
```

scripts/plotting/time_series.py
```diff
@@ -0,0 +1,630 @@
...
+            if 'lev' in ts_ds.coords or 'ilev' in ts_ds.coords:
+                print(f"\t   Variable '{var}' has a vertical dimension, "+\
+                    "which is currently not supported for the time series plot. Skipping...")
+
+                if var in ts_var_list:
+                    #Keep track of vars with levels
+                    #Probably a better way to ignore variables
+                    #with levels than this.
+                    #Look into it - JR
+                    del_s.append(var)
```
nusbaume, line 487: If you convert this to a set, then you'll need to change this to:

```suggestion
                    del_s.add(var)
```

scripts/plotting/time_series.py
```diff
@@ -0,0 +1,630 @@
...
+    """
+    Gather climo year data and format x-axis
+    -----
+        - Set the x-axis plot limits to guarantee data range from all cases (including baseline)
+        - Set minor and major locators based on number of years
+        - Round the range to the nearest 5-year interval for cleaner appearance
+    """
+
+    #Grab all unique years and find min/max years
+    uniq_yrs = sorted({x for v in yrs.values() for x in v})
```
nusbaume, line 581: I think (?) you can remove the curly brackets here:

```suggestion
    uniq_yrs = sorted(x for v in yrs.values() for x in v)
```

nusbaume (changes requested):
Mostly looks good, except it doesn't appear that `timeseries_var_list` is actually defined anywhere in the ADF object, at least in this PR.  Maybe there is a local commit somewhere that still needs to be pushed? 

Otherwise mostly just a bunch of small change requests and slight feature expansions.  As always let me know if you have any questions or concerns with any of my requests.  Thanks!

scripts/analysis/amwg_table.py
```diff
@@ -149,10 +151,13 @@ def amwg_table(adf):
         #Save the baseline to the first case's plots directory:
         output_locs.append(output_locs[0])
 
-    #-----------------------------------------
+    #Declare any derived quantities here:
+    derived_vars = {"RESTOM":["FSNT", "FLNT"]}
+    derived_list = [item for sublist in list(derived_vars.values()) for item in sublist]
```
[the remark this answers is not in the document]
justin-richling: Good catch, thanks!

scripts/plotting/time_series.py
```diff
@@ -0,0 +1,630 @@
...
+    if adfobj.get_basic_info("compare_obs"):
+
+        #Extract variable-obs dictionary:
+        var_obs_dict = adfobj.var_obs_dict
+        base_nickname = "Obs"
+
+        #If dictionary is empty, then there are no observations to regrid to,
+        #so quit here:
+        if not var_obs_dict:
+            print("No observations found to plot against, so no lat/lon maps will be generated.")
```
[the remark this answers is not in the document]
justin-richling: Oops, thanks

scripts/plotting/time_series.py
justin-richling: Yeah, I gave it a couple of tries before and was unsuccessful. Maybe if/when you get some time you can see if something can make sense. I'll also give it another go soon since I've had some time away from it now

scripts/plotting/time_series.py
```diff
@@ -0,0 +1,630 @@
...
+    else:
+        case = case_names[0]
+        multi_case = False
+
+    case_ts_loc = adfobj.get_cam_info("cam_ts_loc", required=True)
+    data_ts_loc = adfobj.get_baseline_info("cam_ts_loc", required=True)
+
+    #Grab test case nickname(s)
+    test_nicknames = adfobj.get_cam_info('case_nickname')
+    for idx,nick_name in enumerate(test_nicknames):
```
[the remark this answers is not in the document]
justin-richling: I updated the call for nicknames to check for `None` in `adf_info`

scripts/plotting/time_series.py
```diff
@@ -0,0 +1,630 @@
...
+
+    #Check if existing plots need to be redone
+    redo_plot = adfobj.get_basic_info('redo_plot')
+    print(f"\t NOTE: redo_plot is set to {redo_plot}")
+
+    #Set seasonal ranges (sans ANN):
+    seasons = ["DJF","MAM","JJA","SON"]
+
+    #Grab only time series variables from YAML file
+    ts_var_list = adfobj.timeseries_var_list
```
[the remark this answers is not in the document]
justin-richling: Hmm, that's weird. I wonder how that got removed, thanks

scripts/plotting/time_series.py
justin-richling: You are correct, doesn't seem to be necesary. Thanks

scripts/analysis/amwg_table.py
```diff
@@ -149,10 +151,13 @@ def amwg_table(adf):
         #Save the baseline to the first case's plots directory:
         output_locs.append(output_locs[0])
 
-    #-----------------------------------------
+    #Declare any derived quantities here:
+    derived_vars = {"RESTOM":["FSNT", "FLNT"]}
```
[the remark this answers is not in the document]
justin-richling: That would probably be best, in case the derived quantities are desired in other plots/scripts. After looking into it quickly it seems like this might be a larger update, so since the time series plot doesn't really depend on it, I'll open an issue and fix that in a separate PR.

scripts/plotting/time_series.py
```diff
@@ -0,0 +1,630 @@
...
+    #matplotlib colors here: https://matplotlib.org/stable/gallery/color/named_colors.html
+    colors = ["k", "aqua", "r", "b", "magenta",
+              "orange", "slategrey", "rosybrown"]
+
+    case_ts_locs = case_ts_loc + [data_ts_loc]
+
+    #Make a separate list to ignore seasonally weighted varaibles
+    #ie RESTOM is only desired (currently) for annual, not seasonally.
+    #Simply add to this list for customization
+    ign = ["RESTOM"]
```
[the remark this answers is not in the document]
justin-richling: Yeah, I like this idea. It seems to fit better to keep this out of the code block, thanks for the suggestion!

scripts/plotting/time_series.py
justin-richling: Yeah, after some digging, the way the seasonal data is calculated, the rolling mean isn't as straight forward as the annual case. Therefore I am setting the rolling mean feature for the annual case only for the time being.

justin-richling:
Closing due to updated PR coming

PR #241, opened 2023-04-13
Provide consolidated averaging methods

brianpm:
I added these functions to `plotting_functions.py`:
- spatial_average : transplanted this from somewhere `amwg_table.py` to start consolidating all the averaging into one place
- annual_mean : provides a reliable way to produce time-weighted annual means
- seasonal_mean : provides a method to make seasonal means that are time-weighted

I then modified:
- `global_latlon_map.py`
- `zonal_mean.py`
- `meridional_mean.py`

To all use the same averaging (i.e., `seasonal_mean()`). 

I did **limited** testing to see if this works. My hope is that this helps to reduce any inconsistencies in the final output. For example, this should provide time-weighted means in both the table and the maps.

I moved the checks for the output files to *before* the averaging. That should allow a speed up when re-running the ADF because the calculations can be skipped once existing plots are detected. 

### an aside
I'm not sure whether we want to reconsider the option `weight_season`? I think whenever we have monthly data, we should use a weighted average (even for weird calendars that don't have different lengths of months, I guess). The weighting could be dropped for high-frequency output, but maybe that shouldn't a user-facing option?

### DIFF
--- a/lib/plotting_functions.py
+++ b/lib/plotting_functions.py
@@ -41,7 +41,7 @@ seasons = {"ANN": np.arange(1,13,1),
             "SON": [9, 10, 11]
             }
 
-            
+
 #################
 #HELPER FUNCTIONS
 #################
@@ -217,7 +217,13 @@ def spatial_average(indata, weights=None, spatial_dims=None):
             spatial_dims = [dimname for dimname in indata.dims if (('lat' in dimname.lower()) or ('lon' in dimname.lower()))]
 
     if not spatial_dims:
-        warnings.warn("No spatial dimensions were identified, so can not perform average.")
+        #Scripts using this function likely expect the horizontal dimensions
+        #to be removed via the application of the mean. So in order to avoid
+        #possibly unexpected behavior due to arrays being incorrectly dimensioned
+        #(which could be difficult to debug) the ADF should die here:
+        emsg = "spatial_average: No spatial dimensions were identified,"
+        emsg += " so can not perform average."
+        raise AdfError(emsg)
 
     return weighted.mean(dim=spatial_dims)
 
@@ -253,7 +259,7 @@ def wgt_rmse(fld1, fld2, wgt):
         return np.sqrt( wmse ).item()
 
 
-####### 
+#######
 # Time-weighted averaging
 
 def annual_mean(data, whole_years=False, time_name='time'):
@@ -265,7 +271,7 @@ def annual_mean(data, whole_years=False, time_name='time'):
         data_to_avg = data.isel(time=slice(first_january,last_december+1)) # PLUS 1 BECAUSE SLICE DOES NOT INCLUDE END POINT
     else:
         data_to_avg = data
-    date_range_string = f"{data_to_avg['time'][0]} -- {data_to_avg['time'][-1]}" 
+    date_range_string = f"{data_to_avg['time'][0]} -- {data_to_avg['time'][-1]}"
 
     # this provides the normalized monthly weights in each year
     # -- do it for each year to allow for non-standard calendars (360-day)
@@ -279,17 +285,17 @@ def annual_mean(data, whole_years=False, time_name='time'):
 
 def seasonal_mean(data, season=None, is_climo=None):
     """Calculates the time-weighted seasonal average (or average over all time).
-    
+
     If season is `ANN` or None, it will default to averaging over all available time.
 
-    If is_climo is True, expects data to have time or month dimenion of size 12. 
-    If is_climo is False, then time must be a coordinate, 
+    If is_climo is True, expects data to have time or month dimenion of size 12.
+    If is_climo is False, then time must be a coordinate,
     and the time.dt.days_in_month attribute must be available.
 
     If the data is a climatology, the code will make an attempt to understand the time or month
     dimension, but basically will assume that it is ordered from January to December.
     If the data is a climatology and is just a numpy array with one dimension that is size 12,
-    it will assume that dimension is time running from January to December. 
+    it will assume that dimension is time running from January to December.
 
     """
     if season is not None:
@@ -314,7 +320,7 @@ def seasonal_mean(data, season=None, is_climo=None):
                         has_time = True
             if not has_time:
                 # this might happen if a pure numpy array gets passed in
-                # --> assumes ordered January to December. 
+                # --> assumes ordered January to December.
                 assert ((12 in data.shape) and (data.shape.count(12) == 1)), f"Sorry, {data.shape.count(12)} dimensions have size 12, making determination of which dimension is month ambiguous. Please provide a `time` or `month` dimension."
                 time_dim_num = data.shape.index(12)
                 fakedims = [f"dim{n}" for n in range(len(data.shape))]
@@ -325,8 +331,8 @@ def seasonal_mean(data, season=None, is_climo=None):
         month_length = data.time.dt.days_in_month
 
     data = data.sel(time=data.time.dt.month.isin(seasons[season])) # directly take the months we want based on season kwarg
-    return data.weighted(data.time.dt.daysinmonth).mean(dim='time')    
-        
+    return data.weighted(data.time.dt.daysinmonth).mean(dim='time')
+
 
 
 #######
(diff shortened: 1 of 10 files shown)


scripts/plotting/global_latlon_map.py
```diff
@@ -336,6 +283,20 @@ def global_latlon_map(adfobj):
                             elif (redo_plot) and plot_name.is_file():
                                 plot_name.unlink()
 
+
+                            if weight_season:
+                                mseasons[s] = pf.seasonal_mean(mdata, season=s, is_climo=True)
+                                oseasons[s] = pf.seasonal_mean(mdata, season=s, is_climo=True)
```
nusbaume, line 289: I think we'll want to use `odata` here:

```suggestion
                                oseasons[s] = pf.seasonal_mean(odata, season=s, is_climo=True)
```

scripts/plotting/global_latlon_map.py
```diff
@@ -442,6 +375,21 @@ def global_latlon_map(adfobj):
                                 elif (redo_plot) and plot_name.is_file():
                                     plot_name.unlink()
 
+                                #If requested, then calculate the monthly-weighted seasonal averages:
+                                if weight_season:
+                                    mseasons[s] = pf.seasonal_mean(mdata, season=s, is_climo=True)
+                                    oseasons[s] = pf.seasonal_mean(mdata, season=s, is_climo=True)
```
nusbaume, line 381: I think we'll want to use `odata` here:

```suggestion
                                    oseasons[s] = pf.seasonal_mean(odata, season=s, is_climo=True)
```

scripts/plotting/global_latlon_map.py
```diff
@@ -442,6 +375,21 @@ def global_latlon_map(adfobj):
...
+                                else:
+                                    #Just average months as-is:
+                                    mseasons[s] = mdata.sel(time=seasons[s], lev=pres).mean(dim='time')
+                                    oseasons[s] = odata.sel(time=seasons[s], lev=pres).mean(dim='time')
+                                    # difference: each entry should be (lat, lon)
+                                    dseasons[s] = mseasons[s] - oseasons[s]
```
nusbaume, line 387: Should `dseasons` be outside the `else`  block?

scripts/plotting/meridional_mean.py
```diff
@@ -220,32 +220,22 @@ def meridional_mean(adfobj):
...
                         adfobj.add_website_data(plot_name, var, case_name, season=s,
                                                 plot_type="Meridional")
-
                         #Continue to next iteration:
                         continue
                     elif (redo_plot) and plot_name.is_file():
                         plot_name.unlink()
 
+                    mseasons[s] = pf.seasonal_mean(mdata, season=s, is_climo=True)
+                    oseasons[s] = pf.seasonal_mean(mdata, season=s, is_climo=True)
```
nusbaume, line 236: I think we'll want to use `odata` here:

```suggestion
                    oseasons[s] = pf.seasonal_mean(odata, season=s, is_climo=True)
```

scripts/plotting/zonal_mean.py
```diff
@@ -269,6 +252,16 @@ def zonal_mean(adfobj):
                         plot_name.unlink()
                     #End if
 
+                    #Seasonal Averages
+                    mseasons[s] = pf.seasonal_mean(mdata, season=s, is_climo=True) #mdata.sel(time=seasons[s]).mean(dim='time')
+                    oseasons[s] = pf.seasonal_mean(mdata, season=s, is_climo=True) #odata.sel(time=seasons[s]).mean(dim='time')
```
nusbaume, line 257: I think we'll want to use `odata` here, and get rid of the commented out code:

```suggestion
                    mseasons[s] = pf.seasonal_mean(mdata, season=s, is_climo=True)
                    oseasons[s] = pf.seasonal_mean(odata, season=s, is_climo=True)
```

scripts/plotting/meridional_mean.py
```diff
@@ -220,32 +220,22 @@ def meridional_mean(adfobj):
...
-                    # difference: each entry should be (lat, lon) or (plev, lat, lon)
-                    # dseasons[s] = mseasons[s] - oseasons[s]
-                    # difference will be calculated in plot_meridional_mean_and_save.
-
-                    # time to make plot; here we'd probably loop over whatever plots we want for this variable
-                    # I'll just call this one "Meridional_Mean"  ... would this work as a pattern [operation]_[AxesDescription] ?
-                    # NOTE: Up to this point, nothing really differs from global_latlon_map,
-                    #       so we could have made one script instead of two.
-                    #       Merging would make overall timing better because looping twice will double I/O steps.
-                    #
```
nusbaume, line 235: Just checking if you wanted to save this comment block down below, or completely get rid of it like what is being done currently?

nusbaume (changes requested):
Looks like a great improvement, thanks @brianpm!  I know it appears that I left a bunch of comments, but most of them should be trivial to resolve.

Also, I am happy to get rid of `weight_season` config argument.  I think the only reason I added it originally was because the old AMWG package had it and warned of a significant slowdown, but I don't get the sense that same slow down happens for the ADF.  Plus we are always adding new config variables, so getting rid of at least one is probably a good idea.

Finally, should we add monthly weighting to the polar and vector plots as well?

lib/plotting_functions.py
```diff
@@ -172,6 +183,44 @@ def global_average(fld, wgt, verbose=False):
...
+    if spatial_dims is None:
+        if 'ncol' in indata.dims:
+            spatial_dims = ['ncol']
+        else:
+            spatial_dims = [dimname for dimname in indata.dims if (('lat' in dimname.lower()) or ('lon' in dimname.lower()))]
+
+    if not spatial_dims:
+        warnings.warn("No spatial dimensions were identified, so can not perform average.")
+
+    return weighted.mean(dim=spatial_dims)
```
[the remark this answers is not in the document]
brianpm: Looks like `arr.mean(dim=[])` just returns the original `arr` without doing averaging.

scripts/analysis/amwg_table.py
```diff
@@ -376,33 +374,7 @@ def _load_data(dataloc, varname):
...
-        weights.name = "weights"
-    #End if
-
-    #Apply weights to input data:
-    weighted = indata.weighted(weights)
-
-    # we want to average over all non-time dimensions
-    avgdims = [dim for dim in indata.dims if dim != 'time']
-    return weighted.mean(dim=avgdims)
+    return pf.spatial_average(indata)
```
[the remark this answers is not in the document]
brianpm: Yeah, I don't see why not. Done.

scripts/plotting/global_latlon_map.py
brianpm: Well, that'd explain some of the strange behavior I saw in some recent testing.

scripts/plotting/global_latlon_map.py
brianpm: Yes.

scripts/plotting/meridional_mean.py
brianpm: I think I just wanted to delete it. I didn't think it was useful at this point.

brianpm:
@nusbaume -- I think I addressed all your comments. I also update the polar and vector plots to use `seasonal_mean`. I also added a `.sel(lev=pres)` in the maps for the case when we should be selecting a pressure level. I think that I missed that before. 

I just pushed the updates to my branch -- they appear automatically here, right?

lib/plotting_functions.py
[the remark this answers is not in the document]
nusbaume: Thanks for looking that up @brianpm!  I am worried that if we return the `weighted` array without any dimension reductions from averaging that the scripts using this function could produce un-expected or un-predictable behavior that might be really difficult to track down (e.g. everything "works" but the final values are wrong).  Given this I might go ahead and make the `if not spatial_dims` check and error instead of a warning.  If you or @justin-richling disagree with that plan please let me know.  Thanks!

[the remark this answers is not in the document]
brianpm: That is perfectly ok with me.

nusbaume (approved):
Looks good to me!  Once @justin-richling is happy with it we can go ahead and get it merged.

justin-richling (approved):
Looks good on my end, thanks @brianpm!

PR #243, opened 2023-05-02
Add check for incorrect climo years

### LINKED ISSUE #231: Incorrect climo years in website header compared to actual available climo years
### ADF run type

Model vs. Model

### What happened?

The climo years represented in the header of the website and in the plots reflect what the user sets in the config file, assuming the climo years were set.

However, in some cases the user may be incorrect in assigning the climo years to what available years are in the datasets.

Example I tried years 1-39:

Plot only goes 30-39:

Screen Shot 2023-02-21 at 11 06 38 AM (3)

Verify there is only 30-39 in h0 files:

Screen Shot 2023-02-21 at 11 08 08 AM (2)

I will investigate, but should be an easy fix; just need to verify the climo years from the range of h0 files in the early part of the ADF run.

### ADF Hash you are using

381cc5c

CISL machine

NPL (CISL machines only)

### Extra info

_No response_

justin-richling:
If user declares incorrect start and/or end climo years in config yaml file, check if in the datasets and force first and/or last years as default if not.

This will propagate to not just web headers but also any place climo years are displayed in plots as well as case_XXXX_XXXX_vs_case_XXXX_XXXX local directory name

Closes #231

### DIFF
--- a/lib/adf_config.py
+++ b/lib/adf_config.py
@@ -272,6 +272,7 @@ class AdfConfig(AdfBase):
             emsg = "Supplied 'conf_dict' variable should be a dictionary,"
             emsg += f" not type '{type(conf_dict)}'"
             raise TypeError(emsg)
+        #End if
 
         #Check that variable name exists in dictionary:
         if varname not in var_dict.keys():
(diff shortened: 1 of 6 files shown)


nusbaume:
@justin-richling Just FYI I expanded the year-checking code some so that the ADF will run when given (hopefully) any positive integer combination of start or end years.

nusbaume (approved):

PR #248, opened 2023-06-08
Small ADF updates

justin-richling:
* Add case nicknames
* Add check in climo years for years that have leading zeros
    - Update from previous PR
* Add obs file and variable name to plot titles
* Add log-p plots for zonal maps
* Update AMWG config file to match updates in baseline example config file
* Add check for number of levels in difference plots

### DIFF
--- a/lib/adf_variable_defaults.yaml
+++ b/lib/adf_variable_defaults.yaml
@@ -132,6 +132,19 @@ AODVIS:
   add_offset: 0
   new_unit: ""
 
+AODVISdn:
+  category: "Aerosols"
+  colormap: "jet"
+  contour_levels_range: [0.01, 1.01, 0.05]
+  diff_colormap: "PuOr_r"
+  diff_contour_range: [-0.4, 0.401, 0.05]
+  scale_factor: 1
+  add_offset: 0
+  new_unit: ""
+  obs_file: "MOD08_M3_192x288_AOD_2001-2020_climo.nc"
+  obs_name: "MODIS"
+  obs_var_name: "AOD_550_Dark_Target_Deep_Blue_Combined_Mean_Mean"
+
 BURDENBC:
   category: "Aerosols"
 
--- a/scripts/plotting/zonal_mean.py
+++ b/scripts/plotting/zonal_mean.py
@@ -300,9 +300,9 @@ def zonal_mean(adfobj):
 
                     if plot_name not in zonal_skip:
 
-                        #Grab seasonally weighted means
-                        mseasons[s] = mdata.sel(time=seasons[s]).mean(dim='time')
-                        oseasons[s] = odata.sel(time=seasons[s]).mean(dim='time')
+                        #Seasonal Averages
+                    mseasons[s] = pf.seasonal_mean(mdata, season=s, is_climo=True)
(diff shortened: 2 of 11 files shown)


scripts/plotting/zonal_mean.py
```diff
@@ -263,8 +259,8 @@ def zonal_mean(adfobj):
                         adfobj.add_website_data(plot_name, var, case_name, season=s,
                                                 plot_type="Zonal")
 
-                        #Continue to next iteration:
-                        continue
+                        #Continue to log-p plots:
+                        pass
```
nusbaume, line 263: I think with this change the non-log-p plots will still be run below even if `redo_plot` is `False`.  I think the best solution would be to move this `redo_plot` check to higher up in the routine (before the data is read in), and to have it check the existence of both the `Zonal_Mean` and `Zonal_Mean_logp` plots, and then to create new logical variables that skip the relevant data-reading and plotting steps as needed depending on if each individual plot type already exists.

Hopefully that makes sense?  If not then we can certainly chat about it in person later this week.

nusbaume (changes requested):
Looks good!  Mostly just some minor code clean-up stuff, with the one possible exception of modifying `redo_plot` in the zonal average plotting script to ensure that the plots and data reading are skipped appropriately.

lib/adf_variable_defaults.yaml
```diff
@@ -132,6 +132,19 @@ AODVIS:
...
+AODVISdn:
+  category: "Aerosols"
+  colormap: "jet"
+  contour_levels_range: [0.01, 1.01, 0.05]
+  diff_colormap: "PuOr_r"
+  diff_contour_range: [-0.4, 0.401, 0.05]
+  scale_factor: 1
+  add_offset: 0
+  new_unit: ""
+  obs_file: "MOD08_M3_192x288_AOD_2001-2020_climo.nc"
```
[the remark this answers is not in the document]
justin-richling: Yeah, that'd be great if you can house these files until we find a permanent home. This and another AOD obs file live here: /glade/work/richling/data/AOD_data/

scripts/plotting/zonal_mean.py
justin-richling: @nusbaume Take a look at my most recent commit for the order of checking for plots. 

There is a loop for all possible plot names and for each plot, if redo_plot is false and the file exists, each existing plot name will be added to a list before the data loop.
(For the zonal it has two checks one for regular and one for log-p. It's not the most elegant solution, so I'm open for improvements.)

In the data collection section when it iterates over plot names again before it calculates the seasonal mean, it will check to see if the plot name is in the new list and then calculates the mean if necessary.

My test run with ~10 variables for a single year, when re-running the ADF with existing plots, was a 10X improvement in elapsed time.

Let me know what you think.

lib/adf_variable_defaults.yaml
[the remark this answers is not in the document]
nusbaume: Done!  The files are now in the "default" obs directory.

scripts/plotting/zonal_mean.py
nusbaume: This update looks great to me, thanks!

justin-richling:
@nusbaume Looks like I didn't get them pushed. They have been committed now.

Issue #225, opened 2023-01-12
TEM Calculations/Plots


justin-richling:
### New feature type

New plot and/or plot enhancement

### What is this new feature?

The Transformed Eulerian Mean diagnostics would be a good addition to the ADF as an option similar to the QBO.

I've spoken with @dan800 and have his working code. There will need to be a little discussion about specifics like how the plots come out, single vs multiple time steps desired, etc. 

If you are looking to use this set of diagnostics, please comment below on what specifics the ADF should use for the TEM calcs/plots.

### Assistance required?

No, I will make a PR when the feature is ready

### Extra info

I currently have a working version in the ADF, with all plots in one image:

![FWsc2000climo_f09_TEM_test05_tem_test](https://user-images.githubusercontent.com/56696811/212189476-a64e35f0-20a9-41ee-a428-08d266dd6349.png)

justin-richling:
Some comments from Dan about current TEM ADF calcs/plots:

>I think we really need the seasonal averages to judge the model circulation and forcing. We now have the infrastructure in place and just need to tweak the figures to make them the most useful.
>
>We may also want to have the capability to compare the test case to the baseline side by side.

Some comments from Rolando:

> Comments/suggestions:
>
>1. Basically, the output looks good. It is insufficient for a full TEM momentum budget because accelerations due to (parameterized) GW are missing. Ideally, one would add terms like UTGWORO, etc., but I am aware that, in the past, it has been judged to be inefficient to output those terms on a daily basis. 
>
>2. One might also want to look at the TEM thermodynamic balance, in which case zonal-mean T and heating terms can be added to the output (QRL_TOT, QRS_TOT, etc.). I have often done this, and it is a good consistency check between TEM momentum and thermodynamic budgets.
>
>3. Color table scales need tweaking. In some cases, log scales would help, e.g., when showing Fy, Fz, which are strong functions of altitude (they grow as exp(z/H) unless the waves are dissipated.)
>
>4. Color tables vs. line contours: For published work, I find I often need to show contours. Color tables, in particular the blue-red one used here, do not allow fine discrimination among contour levels.
>
>5. It is not clear why “zonal-mean meridional wind” is shown in a set of TEM diagnostics. It doesn’t hurt, but it is not needed for TEM analyses.
>
[...]

PR #250, opened 2023-07-13
TEM diagnostics

justin-richling:
This will calculate TEM files and plots for:
* uzm - Zonal-Mean zonal wind
* epfy - northward component of the Eliassen–Palm flux
* epfz - upward component of the Eliassen–Palm flux
* vtem - Transformed Eulerian mean northward wind
* wtem - Transformed Eulerian mean upward wind
* psitem - Transformed Eulerian mean mass stream function
* utendepfd - tendency of eastward wind due to Eliassen-Palm flux divergence
* utendvtem - tendency of eastward wind due to TEM northward wind advection and the coriolis term
* utendwtem - tendency of eastward wind due to TEM upward wind advection

There are two scripts being added, `scripts/averaging/create_TEM_files.py` for reading in h4 history files, calculating TEM variables and creating new netcdf files. And `scripts/plotting/tem.py` for reading in either the newly created netcdf files, or existing TEM netcdf files and plotting.

Update `adf_variable_defaults.yaml` with TEM variables, including Obs files and variable names.

Update `config_cam_baseline_example.yaml` for TEM scripts

NOTE: If QBO is present, all variables will be calculated, if not, only `uzm` `epfy` `epfz` `vtem` `wtem` `psitem` and `utendepfd` will be calculated.

NOTE: If there are different vertical levels between cases, currently the difference plots don't work. I am working towards a vertical and horizontal interpolation for TEM as noted in issue #225

### DIFF
--- a/scripts/averaging/create_TEM_files.py
+++ b/scripts/averaging/create_TEM_files.py
@@ -5,11 +5,12 @@ from numpy import ma
 from datetime import date
 from pathlib import Path
 from glob import glob
+from itertools import chain
 
 
 def create_TEM_files(adf):
     """
-    Calculate the the TEM variables and create new netCDF files
+    Calculate the TEM variables and create new netCDF files
     
     """
 
@@ -24,24 +25,34 @@ def create_TEM_files(adf):
     start_years   = adf.climo_yrs["syears"]
     end_years     = adf.climo_yrs["eyears"]
 
-    #Extract baseline years (which may be empty strings if using Obs):
-    syear_baseline = adf.climo_yrs["syear_baseline"]
-    eyear_baseline = adf.climo_yrs["eyear_baseline"]
-
     #Check if comparing to observations
     if adf.get_basic_info("compare_obs"):
         base_name = "Obs"
     else:
-        base_name = adf.get_baseline_info("cam_case_name", required=True) # does not get used, is just here as a placemarker
+        base_name = adf.get_baseline_info("cam_case_name", required=True)
         cam_hist_locs.append(adf.get_baseline_info("cam_hist_loc", required=True))
+
+        #Extract baseline years (which may be empty strings if using Obs):
+        syear_baseline = adf.climo_yrs["syear_baseline"]
+        eyear_baseline = adf.climo_yrs["eyear_baseline"]
 
         case_names.append(base_name)
         start_years.append(syear_baseline)
         end_years.append(eyear_baseline)
     #End if
 
-    #New TEM netCDF file save location
-    output_loc = adf.get_basic_info("tem_loc")
+    #Grab TEM diagnostics options
+    tem_opts = adf.read_config_var("tem_info")
+    
+    #Set default to h4
+    #TODO: Read this history file number from the yaml file?
+    hist_num = tem_opts["hist_num"]
+    print("tem hist_num",hist_num)
+    if hist_num is None:
+        hist_num = "h4"
+
+    #Extract TEM file save location
+    output_loc = tem_opts["tem_loc"]
     
     #If path not specified, skip TEM calculation?
     if output_loc is None:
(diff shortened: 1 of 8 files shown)


nusbaume (changes requested):
Great work!  However, I do have some concerns, particularly with making sure we are actually grabbing all of the model history files.  If you have any questions or concerns with my change requests just let me know.  Thanks!

scripts/averaging/create_TEM_files.py
```diff
@@ -0,0 +1,392 @@
...
+
+        obs_loc = adf.get_basic_info("obs_data_loc")
+        
+        #Group all TEM observation files together
+        tem_base = []
+        for var in var_list:
+            if var in res:
+                #Gather from variable defaults file
+                obs_file = res[var]["obs_file"]
+                tem_base.append(Path(obs_loc) / obs_file)
```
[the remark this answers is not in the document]
justin-richling: I added a full path to the TEM obs file in my work drive and set only one of the TEM vars to it in the variable defaults file and left the rest to your defaults path. This seems to do the trick:
```python
            if var in res:
                #Gather from variable defaults file
                obs_file = res[var]["obs_file"]

                #It's likely multiple TEM vars will come from one file, so add it and check
                #to see if it already exists for each var.
                if obs_file not in tem_obs_fils:
                    tem_obs_fils.append(obs_file)
```

justin-richling:
@nusbaume I think I've tackled all your suggestions and have run a couple tests. Let me know if anything else needs addressing, thanks!

justin-richling:
@nusbaume Last changes are in place if you want to review this again, thanks!

scripts/averaging/create_TEM_files.py
```diff
@@ -136,53 +174,75 @@ def create_TEM_files(adf):
         if not starting_location.is_dir():
             emsg = f"Provided 'cam_hist_loc' directory '{starting_location}' not found."
             emsg += " Script is ending here."
-            adf.end_diag_fail(emsg)
+            return
```
nusbaume, line 177: I think we do want the `end_diag_fail` here, because if `cam_hist_locs`  can't be found then something has gone catastrophically wrong, and we want to kill the ADF as soon as possible.

scripts/averaging/create_TEM_files.py
```diff
@@ -136,53 +174,75 @@ def create_TEM_files(adf):
...
         if not output_loc_idx.is_dir():
             print(f"    {output_loc_idx} not found, making new directory")
             output_loc_idx.mkdir(parents=True)
         #End if
 
-        # write output to a netcdf file
-        dstem0.to_netcdf(output_loc_idx / f'{case_name}.TEMdiag_{start_year}-{end_year}.nc', 
-                            unlimited_dims='time', 
-                            mode = 'w' )
+        #Set baseline file name
```
nusbaume, line 197: I think this comment should be:

```suggestion
        #Set case file name
```

scripts/plotting/tem.py
```diff
@@ -74,11 +72,15 @@ def tem(adf):
     print(f"\t NOTE: redo_plot is set to {redo_plot}")
     #-----------------------------------------
 
-    #Location to saved TEM netCDF files
-    tem_loc = adf.get_basic_info("tem_loc")
+    #Grab TEM diagnostics options
+    tem_opts = adf.read_config_var("tem_info")
```
nusbaume, line 76: I think we need a check here that if `tem_opts` is `None`, then the script returns with a warning message (otherwise users will run into an ADF-killing error on the line below).

scripts/plotting/tem.py
```diff
@@ -74,11 +72,15 @@ def tem(adf):
...
+
+    #Location of saved TEM netCDF files
+    tem_loc = tem_opts["tem_loc"]
```
nusbaume, line 79: I think we need to use the `get` method here:

```suggestion
    tem_loc = tem_opts.get("tem_loc")
```

nusbaume (changes requested):
Looking a lot better!  I just have some (hopefully) final config variable checking and cleanup requests.

justin-richling:
@nusbaume Can you take another look and see if this is all good now. I ran it with your suggested changes and it appears to work as intended for when `tem_opts` is included and missing in the config file.

nusbaume (approved):
Everything looks great to me now, so feel free to merge whenever you are ready.  Thanks!

Issue #253, opened 2023-08-15
ADF not picking the correct year for the climo


cecilehannay:
### ADF run type

Model vs. Model

### What happened?

I am running the latest version of the ADF on a B1850 run. I am trying to select a particular set of years but I am not sure it is working correctly as I get a warning message when starting the ADF. 

For instance, I am trying to select the years 10-25 that is common to all my runs:
``` 
   start_year: 10
   end_year: 25
```

But I am getting a warming saying those years don't exist:
```
['b.e23_alpha16b.BLT1850.ne30_t232.033']
Given start year '10' is not in current dataset b.e23_alpha16b.BLT1850.ne30_t232.035, using first found year: 0001 

Given end year '25' is not in current dataset b.e23_alpha16b.BLT1850.ne30_t232.035, using last found year: 0028 

Given start year '10' is not in current dataset b.e23_alpha16b.BLT1850.ne30_t232.033, using first found year: 0001 

Given end year '25' is not in current dataset b.e23_alpha16b.BLT1850.ne30_t232.033, using last found year: 0026 
```

According to the message, the ADF is not using:
``` 
   start_year: 10
   end_year: 25
```
but instead, it seems to be using for the first run:
```
   start_year: 1
   end_year: 28
```
and for the second run:
```
   start_year: 1
   end_year: 26
```

I tried to set: 
``` 
   start_year: 0010
   end_year: 0025
```
to see if it was helping but it doesn't

### ADF Hash you are using

3e2ade6

### What machine were you running the ADF on?

CISL machine

### What python environment were you using?

NPL (CISL machines only)

brianpm:
This bug must have something to do with the way the file names are parsed in `adf_info.py`.

Here's the relevant code:
```python
            #Check if history file path exists:
            if baseline_hist_locs:

                starting_location = Path(baseline_hist_locs)
                files_list = sorted(starting_location.glob('*'+hist_str+'.*.nc'))
                base_climo_yrs = sorted(np.unique([i.stem[-7:-3] for i in files_list]))

                #Check if start or end year is missing.  If so then just assume it is the
                #start or end of the entire available model data.
                if syear_baseline is None:
                    print(f"No given start year for {data_name}, using first found year...")
                    syear_baseline = int(base_climo_yrs[0])
                elif str(syear_baseline) not in base_climo_yrs:
                    print(f"Given start year '{syear_baseline}' is not in current dataset {data_name}, using first found year:",base_climo_yrs[0],"\n")
[...]

PR #254, opened 2023-08-16
Climo bugfix

cecilehannay:
This bug fix addresses the issue described at:
https://github.com/NCAR/ADF/issues/253

The problem was that we were using a string to identify the start year and the end year of the climo years. This approach worked when we had a 4-digit year (e.g., 1979) as in the "F" cases. However, it failed for single-digit years (e.g., 1), as seen in the "B" cases.  Replacing str by int in the code seems to fix the issue.

For instance:
`elif str(syear_baseline) not in base_climo_yrs:`
is replaced by:
`elif (syear_baseline) not in base_climo_yrs:`


lib/adf_info.py
```diff
@@ -170,25 +170,28 @@ def __init__(self, config_file, debug=False):
...
+                for year in base_climo_yrs_str:
+                   base_climo_yrs.append(int(year))
+                
                 #Check if start or end year is missing.  If so then just assume it is the
                 #start or end of the entire available model data.
                 if syear_baseline is None:
                     print(f"No given start year for {data_name}, using first found year...")
                     syear_baseline = int(base_climo_yrs[0])
-                elif str(syear_baseline) not in base_climo_yrs:
+                elif (syear_baseline) not in base_climo_yrs:
```
nusbaume, line 183: Probably don't need the parentheses anymore:

```suggestion
                elif syear_baseline not in base_climo_yrs:
```

lib/adf_info.py
```diff
@@ -170,25 +170,28 @@ def __init__(self, config_file, debug=False):
...
                     print(f"Given start year '{syear_baseline}' is not in current dataset {data_name}, using first found year:",base_climo_yrs[0],"\n")
                     syear_baseline = int(base_climo_yrs[0])
                 #End if
                 if eyear_baseline is None:
                     print(f"No given end year for {data_name}, using last found year...")
                     eyear_baseline = int(base_climo_yrs[-1])
-                elif str(eyear_baseline) not in base_climo_yrs:
+                elif (eyear_baseline) not in base_climo_yrs:
```
nusbaume, line 190: Probably don't need the parentheses anymore:

```suggestion
                elif eyear_baseline not in base_climo_yrs:
```

lib/adf_info.py
```diff
@@ -170,25 +170,28 @@ def __init__(self, config_file, debug=False):
...
+                base_climo_yrs_str = sorted(np.unique([i.stem[-7:-3] for i in files_list]))
+                base_climo_yrs = []
+                for year in base_climo_yrs_str:
+                   base_climo_yrs.append(int(year))
+                
                 #Check if start or end year is missing.  If so then just assume it is the
                 #start or end of the entire available model data.
                 if syear_baseline is None:
                     print(f"No given start year for {data_name}, using first found year...")
                     syear_baseline = int(base_climo_yrs[0])
```
nusbaume, line 182: We no longer need any of these `int` calls in this code block for `base_climo_yrs`, as it is now always an integer.

lib/adf_info.py
```diff
@@ -261,24 +264,28 @@ def __init__(self, config_file, debug=False):
...
+                case_climo_yrs_str = sorted(np.unique([i.stem[-7:-3] for i in files_list]))
+                case_climo_yrs = []
+                for year in case_climo_yrs_str:
+                   case_climo_yrs.append(int(year))
 
                 #Check if start or end year is missing.  If so then just assume it is the
                 #start or end of the entire available model data.
                 if syear is None:
                     print(f"No given start year for {case_name}, using first found year...")
                     syear = int(case_climo_yrs[0])
```
nusbaume, line 276: Same note as above, I would remove the `int` calls here and below because `case_climo_yrs` is always an integer now.

lib/adf_info.py
```diff
@@ -261,24 +264,28 @@ def __init__(self, config_file, debug=False):
...
-                elif str(syear) not in case_climo_yrs:
+                elif (syear) not in case_climo_yrs:
```
nusbaume, line 277: No need to have the parentheses here:

```suggestion
                elif syear not in case_climo_yrs:
```

lib/adf_info.py
```diff
@@ -261,24 +264,28 @@ def __init__(self, config_file, debug=False):
...
                     print(f"Given start year '{syear}' is not in current dataset {case_name}, using first found year:",case_climo_yrs[0],"\n")
                     syear = int(case_climo_yrs[0])
                 #End if
                 if eyear is None:
                     print(f"No given end year for {case_name}, using last found year...")
                     eyear = int(case_climo_yrs[-1])
-                elif str(eyear) not in case_climo_yrs:
+                elif (eyear) not in case_climo_yrs:
```
nusbaume, line 284: No need to have the parentheses here:

```suggestion
                elif eyear not in case_climo_yrs:
```

nusbaume (approved):
Looks good to me, although I do have some non-required cleanup requests that should be easy to implement if you have the time.

Also, I want @justin-richling to review these changes as well before we get this bug fix merged, just in case I have missed anything obvious.   Thanks!

PR #255, opened 2023-08-31
Casename fix

justin-richling:
This PR will fix the recent issue of the case name coming out as `None` in the plot titles.

This fixes the bug in both single and mulit-case scenarios.

### DIFF
--- a/lib/adf_info.py
+++ b/lib/adf_info.py
@@ -117,15 +117,17 @@ class AdfInfo(AdfConfig):
         #Case names:
         case_names = self.get_cam_info('cam_case_name', required=True)
 
-        print(case_names)
-
         #Grab test case nickname(s)
-        test_nicknames = self.get_cam_info('case_nickname')
-        if test_nicknames is None:
-            test_nicknames = [] #Re-set to be an empty list
-            for case_name in case_names:
-                test_nicknames.append(case_name)
-            #End for
+        test_nickname_list = self.get_cam_info('case_nickname')
+
+        test_nicknames = [] #Re-set to be an empty list
+        for i,nickname in enumerate(test_nickname_list):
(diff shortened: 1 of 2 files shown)


lib/adf_info.py
```diff
@@ -117,16 +117,17 @@ def __init__(self, config_file, debug=False):
...
+        test_nickname_list = self.get_cam_info('case_nickname')
+
+        test_nicknames = [] #set to be an empty list
+        for i,nickname in enumerate(test_nickname_list):
+            if nickname is None:
+                test_nicknames.append(case_names[i])
+            else:
+                test_nicknames.append(test_nickname_list[i])
+            #End if
+        #End for
```
nusbaume, line 130: I realized that we'll still want to check if `test_nicknames_list` is None in case the `case_nickname` config variable isn't present at all:

```suggestion
        if test_nickname_list:
            test_nicknames = [] #set to be an empty list
            for i,nickname in enumerate(test_nickname_list):
                if nickname is None:
                    test_nicknames.append(case_names[i])
                else:
                    test_nicknames.append(test_nickname_list[i])
                #End if
            #End for
        else:
            test_nicknames = [] #Re-set to be an empty list
            for case_name in case_names:
                test_nicknames.append(case_name)
            #End for
        #End if
```

nusbaume (changes requested):
Way to track this bug down!  However, I realized that we'll still want to check that the `case_nicknames` config variable exists before trying to loop over it.

lib/adf_info.py
justin-richling: Ah, good call. I forgot that check, thanks!

justin-richling:
@nusbaume I updated the file per you suggestion, it should be good to review again

PR #258, opened 2023-09-07
one-character bug fix in the regridding script

### LINKED ISSUE #257: Regridded baseline variables set to wrong values during regrid_and_vert_interp
### ADF run type

Model vs. Model

### What happened?

I believe that `regrid_and_vert_interp.py` has an error such that when a variable is being masked (to remove land and sea ice points), the baseline file that is produced is a copy of the last diagnostic case. This happens at line 330:
```
if tgt_ocn_frc_ds:
    ofrac = tgt_ocn_frc_ds['OCNFRAC']
    # set the bounds of regridded ocnfrac to 0 to 1
    ofrac = xr.where(ofrac>1,1,ofrac)
    ofrac = xr.where(ofrac<0,0,ofrac)
    # mask the land in TS for global means
    tgdata_interp['OCNFRAC'] = ofrac
    ts_tmp = rgdata_interp[var] #### <<<< BUG
    ts_tmp = pf.mask_land_or_ocean(ts_tmp,ofrac)
    tgdata_interp[var] = ts_tmp
```
I think that line marked BUG should be:
```
    ts_tmp = tgdata_interp[var]
```

This results in incorrect plots because the test case and the reference case will be the same for any fields that are marked as masked in the variable defaults yaml file. 

### ADF Hash you are using

9a10372

CISL machine

Personl Conda env

### Extra info

_No response_

brianpm:
Resolves #257

### DIFF
--- a/scripts/regridding/regrid_and_vert_interp.py
+++ b/scripts/regridding/regrid_and_vert_interp.py
@@ -275,6 +275,7 @@ def regrid_and_vert_interp(adf):
 
                     #Finally, write re-gridded data to output file:
                     save_to_nc(rgdata_interp, regridded_file_loc)
+                    rgdata_interp.close()  # bpm: we are completely done with this data
 
                     #Now vertically interpolate baseline (target) climatology,
                     #if applicable:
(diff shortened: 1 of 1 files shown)


nusbaume (approved):
Looks good to me, thanks for tracking down this bug!

Also I requested @justin-richling's review, but once he approves it as well then feel free to merge this PR in at your convivence.

justin-richling:
@nusbaume and @brianpm Looks good to me, thanks for catching this small but consequential bug.

justin-richling (approved):

PR #260, opened 2023-09-22
Adding mods to derive PRECT

bstephens82:
from PRECC+PRECL if PRECT is not in the cam.h0 files.  This code could be extended to other variables.

### DIFF
--- a/lib/adf_diag.py
+++ b/lib/adf_diag.py
@@ -379,6 +379,9 @@ class AdfDiag(AdfWeb):
             hist_str = 'cam.h0'
         #End if
 
+        # get info about variables defaults
+        res = self.variable_defaults
+
         #Loop over cases:
         for case_idx, case_name in enumerate(case_names):
 
@@ -510,12 +513,24 @@ class AdfDiag(AdfWeb):
 
             #Loop over CAM history variables:
             list_of_commands = []
-            for var in self.diag_var_list:
+            vars_to_derive = []
+            #create copy of var list that can be modified for derivable variables
+            diag_var_list = self.diag_var_list
+            for var_index,var in enumerate(diag_var_list):
                 if var not in hist_file_var_list:
-                    msg = f"WARNING: {var} is not in the file {hist_files[0]}."
-                    msg += " No time series will be generated."
-                    print(msg)
-                    continue
+                    vres = res[var]
+                    if "derivable_from" in vres:
+                        constit_list = vres['derivable_from']
+                        for constit in constit_list:
+                            if constit not in diag_var_list:
+                                diag_var_list.append(constit)
+                        vars_to_derive.append(var)
+                        continue
+                    else:
+                        msg = f"WARNING: {var} is not in the file {hist_files[0]}."
+                        msg += " No time series will be generated."
+                        print(msg)
+                        continue
 
                 #Check if variable has a "lev" dimension according to first file:
                 has_lev = bool('lev' in hist_file_ds[var].dims)
@@ -592,6 +607,9 @@ class AdfDiag(AdfWeb):
             #Now run the "ncrcat" subprocesses in parallel:
             with mp.Pool(processes=self.num_procs) as mpool:
                 _ = mpool.map(call_ncrcat, list_of_commands)
+
+            if vars_to_derive:
+                self.derive_variables(vars_to_derive=vars_to_derive,directory=ts_dir[case_idx])
             #End with
 
         #End cases loop
@@ -914,5 +932,26 @@ class AdfDiag(AdfWeb):
         #End if
         print('For CVDP information visit: https://www.cesm.ucar.edu/working_groups/CVC/cvdp/')
         print('   ')
+
+    #########
+
+    def derive_variables(self,vars_to_derive=None,directory=None):
+
+        """
+        Derive variables acccording to steps given here.  Since derivations will depend on the 
+        variable, each variable to derive will need its own set of steps below.
+        """
+
+        for var in vars_to_derive:
+            if var == "PRECT":
+                 # PRECT can be found by simply adding PRECL and PRECC
+                 # grab file names for the PRECL and PRECC files from the case ts directory
+                 constit_files = sorted(glob.glob(directory+"/*PREC*"))
+                 # create new file name for PRECT
+                 prect_file = constit_files[0].replace('PRECC','PRECT')
+                 # append PRECC to the file containing PRECL
+                 os.system(f"ncks -A -v PRECC {constit_files[0]} {constit_files[1]}")
+                 # create new file with the sum of PRECC and PRECL
+                 os.system(f"ncap2 -s 'PRECT=(PRECC+PRECL)' {constit_files[1]} {prect_file}")
 
 ###############
--- a/lib/adf_diag.py
+++ b/lib/adf_diag.py
@@ -609,7 +609,7 @@ class AdfDiag(AdfWeb):
                 _ = mpool.map(call_ncrcat, list_of_commands)
 
             if vars_to_derive:
-                self.derive_variables(vars_to_derive=vars_to_derive,directory=ts_dir[case_idx])
+                self.derive_variables(vars_to_derive=vars_to_derive,ts_dir=ts_dir[case_idx])
             #End with
 
         #End cases loop
@@ -935,7 +935,7 @@ class AdfDiag(AdfWeb):
 
     #########
 
-    def derive_variables(self,vars_to_derive=None,directory=None):
+    def derive_variables(self,vars_to_derive=None,ts_dir=None):
 
         """
         Derive variables acccording to steps given here.  Since derivations will depend on the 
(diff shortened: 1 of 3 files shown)


brianpm:
Just to note that #240 is a similar modification to allow RESTOM to be derived. The implementation there is quite different. I'd recommend we encourage doing these kinds of simple derivations in a consistent manner. I like @bstephens82 's approach of just having a `derive_variable` method. I'm not sure how I feel about doing the derivation with `ncap2`, but I'm open to it if others think this is better than using python tools.

bstephens82:
> Just to note that #240 is a similar modification to allow RESTOM to be derived. The implementation there is quite different. I'd recommend we encourage doing these kinds of simple derivations in a consistent manner. I like @bstephens82 's approach of just having a `derive_variable` method. I'm not sure how I feel about doing the derivation with `ncap2`, but I'm open to it if others think this is better than using python tools.

Ah, I had heard RESTOM might be another variable this could be applied to, but didn't realize it had already been implemented.  I figured for each variable that could be derived, we would need to add another "recipe" in `derive_variable`.  In the case of PRECT using `ncap2` was just one of the shortest routes I could come up with.  Feel free to modify or suggest other methods. Is the idea to avoid using NCO operators where possible?

brianpm:
Yeah, I think adding "recipes" for each variable is the right way to go. 

We do use NCO to make time series files, but we have discussed that having NCO as a dependency might not be great for portability. I'm not opposed to using it in the short term, but we might need to revisit in the future. 

I guess my point was mostly to note that we should somehow combine this PR and #240 to allow both `PRECT` and `RESTOM` to be generated when the requirements are met.

brianpm:
I also notice that #232 mentions similar functionality, so should also be harmonized with this one and #240.

justin-richling:
Also I believe this is mentioned (PRECT specifically) in Julie's issue #228, so hopefully this can close that issue out as well

lib/adf_diag.py
```diff
@@ -379,6 +379,9 @@ def call_ncrcat(cmd):
             hist_str = 'cam.h0'
         #End if
 
+        # get info about variables defaults
```
nusbaume, line 382: minor typo:

```suggestion
        # get info about variable defaults
```

lib/adf_diag.py
```diff
@@ -510,12 +513,24 @@ def call_ncrcat(cmd):
...
+            vars_to_derive = []
+            #create copy of var list that can be modified for derivable variables
+            diag_var_list = self.diag_var_list
+            for var in diag_var_list:
                 if var not in hist_file_var_list:
-                    msg = f"WARNING: {var} is not in the file {hist_files[0]}."
-                    msg += " No time series will be generated."
-                    print(msg)
-                    continue
+                    vres = res[var]
```
nusbaume, line 521: I think you'll want to check here if `var` is actually in the `variable_defaults` dictionary, and if not then return an empty dictionary.  So something like this:

```suggestion
                    vres = res.get(var, {})
```

lib/adf_diag.py
```diff
@@ -915,4 +933,25 @@ def setup_run_cvdp(self):
         print('For CVDP information visit: https://www.cesm.ucar.edu/working_groups/CVC/cvdp/')
         print('   ')
 
+    #########
+
+    def derive_variables(self,vars_to_derive=None,ts_dir=None):
+
+        """
+        Derive variables acccording to steps given here.  Since derivations will depend on the 
+        variable, each variable to derive will need its own set of steps below.
```
nusbaume, line 942: I think this is totally fine for now, but long-term I wonder if we can figure out a way to add the derivation equations to the variable defaults YAML file itself, and then simply have this function properly parse those equations to produce the derived variable.  That way future users won't have to modify the core ADF python code if they want to add a new derived variable.

Thoughts?  @brianpm @justin-richling

lib/adf_diag.py
```diff
@@ -915,4 +933,25 @@ def setup_run_cvdp(self):
...
+        """
+
+        for var in vars_to_derive:
+            if var == "PRECT":
+                 # PRECT can be found by simply adding PRECL and PRECC
+                 # grab file names for the PRECL and PRECC files from the case ts directory
```
nusbaume, line 948: We should probably have a check here that skips trying to calculate `PRECT` and prints out a warning if it can't find both `PRECC` and `PRECL`.

lib/adf_diag.py
```diff
@@ -915,4 +933,25 @@ def setup_run_cvdp(self):
...
+                 constit_files = sorted(glob.glob(ts_dir+"/*PREC*"))
```
nusbaume, line 949: I believe you'll want to do the following here in order to avoid potential non-POSIX OS issues:

```suggestion
                 constit_files = sorted(glob.glob(os.path.join(ts_dir,"*PREC*")))
```

nusbaume (changes requested):
Thanks for working on this @bstephens82!  It looks good, but I do have a few changes requests, mostly to avoid possible errors showing up if certain variables aren't present.

lib/adf_diag.py
bstephens82: I like that idea...I am not sure how to implement it though.

lib/adf_diag.py
bstephens82: Done.

lib/adf_diag.py
bstephens82: Done.

lib/adf_diag.py
bstephens82: Ok so the way I put in a check was as follows (I'm not sure this is the best way to do it but it seems pretty explicit and somewhat simple):
~~~
if len(glob.glob(os.path.join(ts_dir,"*PRECC*"))) and len(glob.glob(os.path.join(ts_dir,"*PRECL*"))):
    constit_files=sorted(glob.glob(os.path.join(ts_dir,"*PREC*")))
else:
    ermsg = "PRECC and PRECL were not both present; PRECT cannot be calculated."
    raise FileNotFoundError(ermsg)
~~~
What do you think of this? I like that it explicitly looks for both PRECC and PRECL, and I've tested and it works (including with failure if it doesn't find both files), just wondering if there's an even simpler way.

lib/adf_diag.py
```diff
@@ -946,7 +946,11 @@ def derive_variables(self,vars_to_derive=None,ts_dir=None):
             if var == "PRECT":
                  # PRECT can be found by simply adding PRECL and PRECC
                  # grab file names for the PRECL and PRECC files from the case ts directory
-                 constit_files = sorted(glob.glob(ts_dir+"/*PREC*"))
+                 if len(glob.glob(os.path.join(ts_dir,"*PRECC*"))) and len(glob.glob(os.path.join(ts_dir,"*PRECL*"))):
+                     constit_files=sorted(glob.glob(os.path.join(ts_dir,"*PREC*")))
+                 else:
+                     ermsg = "PRECC and PRECL were not both present; PRECT cannot be calculated."
+                     raise FileNotFoundError(ermsg)
```
nusbaume, line 953: Thanks for this!  I might just change this code slightly to something like the following:

```suggestion
                 if glob.glob(os.path.join(ts_dir,"*PRECC*")) and glob.glob(os.path.join(ts_dir,"*PRECL*")):
                     constit_files=sorted(glob.glob(os.path.join(ts_dir,"*PREC*")))
                 else:
                     ermsg = "PRECC and PRECL were not both present; PRECT cannot be calculated."
                     ermsg += " Please remove PRECT from diag_var_list or find the relevant CAM files."
                     raise FileNotFoundError(ermsg)
```

nusbaume (changes requested):
Almost there!  Just a little bit of extra clean-up on your new file-checking code.

lib/adf_diag.py
bstephens82: Done.

lib/adf_diag.py
```diff
@@ -946,10 +946,16 @@ def derive_variables(self,vars_to_derive=None,ts_dir=None):
             if var == "PRECT":
                  # PRECT can be found by simply adding PRECL and PRECC
                  # grab file names for the PRECL and PRECC files from the case ts directory
-                 if len(glob.glob(os.path.join(ts_dir,"*PRECC*"))) and len(glob.glob(os.path.join(ts_dir,"*PRECL*"))):
+#                 if len(glob.glob(os.path.join(ts_dir,"*PRECC*"))) and len(glob.glob(os.path.join(ts_dir,"*PRECL*"))):
+#                     constit_files=sorted(glob.glob(os.path.join(ts_dir,"*PREC*")))
+#                 else:
+#                     ermsg = "PRECC and PRECL were not both present; PRECT cannot be calculated."
+#                     raise FileNotFoundError(ermsg)
```
nusbaume, line 953: Is there a reason you commented this out instead of just deleting it?

nusbaume (commented):
Super-close, just one last question.

lib/adf_diag.py
bstephens82: Done.

nusbaume (approved):
Looks good to me now, thanks!  Once @justin-richling also reviews and approves it then it will be ready to be merged.

justin-richling (approved):
This looks good on my end, thanks @bstephens82! This will help get us in the right direction for more derived quantities in the future.

PR #270, opened 2023-12-04
making annual mean plots display by default.

bstephens82:


nusbaume (commented):
I think that with this change we no longer need the "no-plot" variable landing page represented by `template_var.html`.  Thus my vote would be to go ahead and delete `template_var.html` and the associated code in `adf_web.py`, which I believe are just lines 666 to 688, in this same PR.  What are your thoughts @justin-richling  and @brianpm?

justin-richling:
@nusbaume Yeah I don't see that being an issue if the `template_var.html` file is literally just a place holder page. The only issue I can see with this is that for the QBO the "seasons" are the time series and amplitude plots. I have an idea that I was working on in a different branch that might help

nusbaume:
@justin-richling good catch!  We'll need some way to handle "special" seasons before this PR can be merged.  Happy to look over your fix whenever it is ready.  Otherwise I think I have a couple ideas myself that we can discuss if need be.

nusbaume (approved):
This all looks good to me now.  Thanks @bstephens82 and @justin-richling!

justin-richling (approved):
Looks good to me, thanks all!

PR #274, opened 2024-01-08
Fix header bug in html webpages

justin-richling:
Climo years in the webpage header were being based off config file declared years. This will take it from the curated climo years done in `adf_info.py`.

This also fixes the bug that shows up when viewing the tables page where both test and baseline cases in html header display the same case name.

### DIFF
--- a/lib/adf_web.py
+++ b/lib/adf_web.py
@@ -535,7 +535,7 @@ class AdfWeb(AdfObs):
                 if web_data.multi_case:
                     case1 = "Listed in tables"
                 else:
-                    case1 = web_data.case
+                    case1 = case_names[0]
                 #End if
 
                 #Write table dataframe HTML as a string:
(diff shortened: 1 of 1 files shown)


lib/adf_web.py
```diff
@@ -343,46 +343,23 @@ def create_website(self):
...
+        #Grab baseline years (which may be empty strings if using Obs):
+        syear_baseline = self.climo_yrs["syear_baseline"]
+        eyear_baseline = self.climo_yrs["eyear_baseline"]
 
         #Set name of comparison data, which depends on "compare_obs":
         if self.compare_obs:
             data_name = "Obs"
             baseline_yrs = ""
+            syear_baseline = ""
+            eyear_baseline = ""
```
nusbaume, line 359: I don't think these lines are needed, as I believe the `climo_yrs` method will return empty strings for the baseline case automatically if `compare_obs` is `True`.

nusbaume (approved):
Nice cleanup!  Just one minor request.

PR #275, opened 2024-01-08
Add flexibility for globbing different h-numbers

justin-richling:
This will add some flexibility in creating time series files from history files that may have extra time info in history file name, ie if the history file is like: $CASENAME.cam.h3.0033-03-02-10800.nc. Currently the ADF uses glob that would ignore these history files.

Also move where ADF appends case climo years in case the ADF has to update years from what user requested.

@bitterbark Could you check this out and see if it works with your MDTF/ADF tests?

### DIFF
--- a/lib/adf_info.py
+++ b/lib/adf_info.py
@@ -177,13 +177,16 @@ class AdfInfo(AdfConfig):
             if baseline_hist_locs:
 
                 starting_location = Path(baseline_hist_locs)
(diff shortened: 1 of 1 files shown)


bitterbark (commented):
I merged this with my hist_str changes and it worked flawlessly on two different atm h files. I'm not going to test it on ocean etc at this time; we'll get to that eventually. Thanks for doing it!

PR #279, opened 2024-01-22
Update TEM configuration

justin-richling:
It seems like it might make more sense to move the TEM configuration into the case subsections as requested by @cecilehannay.

This will move each TEM configuration into case and baseline sections of config file and will update the TEM files to account for the configuration change.

### DIFF
--- a/scripts/averaging/create_TEM_files.py
+++ b/scripts/averaging/create_TEM_files.py
@@ -29,6 +29,13 @@ def create_TEM_files(adf):
     start_years   = adf.climo_yrs["syears"]
     end_years     = adf.climo_yrs["eyears"]
 
+    res = adf.variable_defaults # will be dict of variable-specific plot preferences
+
+    if "qbo" in adf.plotting_scripts:
+        var_list = ['uzm','epfy','epfz','vtem','wtem',
+                    'psitem','utendepfd','utendvtem','utendwtem']
+    else:
+        var_list = ['uzm','epfy','epfz','vtem','wtem','psitem','utendepfd']
 
     tem_locs = []
     
@@ -171,14 +178,6 @@ def create_TEM_files(adf):
             print("\t 'cam_tem_loc' not found in 'diag_cam_baseline_climo', so no baseline files/diagnostics will be generated.")
 
     #End if (check for obs)
-
-    res = adf.variable_defaults # will be dict of variable-specific plot preferences
-
-    if "qbo" in adf.plotting_scripts:
-        var_list = ['uzm','epfy','epfz','vtem','wtem',
-                    'psitem','utendepfd','utendvtem','utendwtem']
-    else:
-        var_list = ['uzm','epfy','epfz','vtem','wtem','psitem','utendepfd']
 
     #Loop over cases:
     for case_idx, case_name in enumerate(case_names):
--- a/scripts/averaging/create_TEM_files.py
+++ b/scripts/averaging/create_TEM_files.py
@@ -14,9 +14,13 @@ def create_TEM_files(adf):
 
     """
 
+    #Notify user that script has started:
+    print("\n  Generating CAM TEM diagnostics files...")
+
     #Special ADF variables
     #CAM simulation variables (these quantities are always lists):
     case_names    = adf.get_cam_info("cam_case_name", required=True)
+    base_name     = adf.get_baseline_info("cam_case_name")
 
     #Grab h4 history files locations
     cam_hist_locs = adf.get_cam_info("cam_hist_loc", required=True)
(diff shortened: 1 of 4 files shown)


scripts/averaging/create_TEM_files.py
```diff
@@ -50,45 +75,60 @@ def create_TEM_files(adf):
         #If dictionary is empty, then there are no observations, so quit here:
         if not var_obs_dict:
             print("No observations found to plot against, so no TEM will be generated.")
-            return
+            pass
```
nusbaume, line 78: If this is now a `pass` statement then I might vote to change the `print` statement to something like:

`"No observations found to plot against, so no obs-based TEM file will be generated".`

This is assuming that the standard CAM case TEM files can still be generated without the obs data.  If not then this should probably be turned back into a `return` statement.

scripts/averaging/create_TEM_files.py
```diff
@@ -50,45 +75,60 @@ def create_TEM_files(adf):
...
-    else:
-        #Notify user that script has started:
-        print("\n  Generating CAM TEM diagnostics files...")
+            #Set default to h4
+            hist_num = adf.get_baseline_info("hist_num")
+            if hist_num is None:
+                hist_num = "h4"
+            hist_nums.append(hist_num)
+        else:
+            print("\t 'cam_tem_loc' not found in 'diag_cam_baseline_climo', so no TEM files/diagnostics will be generated.")
```
nusbaume, line 129: I might update this print statement to be:

```suggestion
            print("\t 'cam_tem_loc' not found in 'diag_cam_baseline_climo', so no baseline TEM files/diagnostics will be generated.")
```

scripts/averaging/create_TEM_files.py
```diff
@@ -102,18 +142,13 @@ def create_TEM_files(adf):
...
-
         #Set baseline file name as full path
-        tem_fil = output_loc_idx / f'{base_name}.TEMdiag.nc'
+        tem_fil = tem_base_loc / f'{base_name}.TEMdiag.nc'
 
-        #Get baseline case tem over-write boolean
-        overwrite_tem = tem_opts.get("overwrite_tem_base")
+        #Get obs case tem over-write boolean
+        #QUESTION: should we make these and put in ADF obs location so we don't have to 
+        #worry about making these??
```
nusbaume, line 150: If the obs data generally isn't changing then the answer to your question is YES.

scripts/averaging/create_TEM_files.py
```diff
@@ -189,15 +224,16 @@ def create_TEM_files(adf):
         #End if
 
         #Check if history files actually exist. If not then kill script:
-        hist_str = '*.cam.'+hist_num
+        #hist_str = '*.cam.'+hist_nums[case_idx]
```
nusbaume, line 227: Remove commented-out line here?

scripts/plotting/tem.py
```diff
@@ -71,26 +68,34 @@ def tem(adf):
...
+    #Extract TEM file save locations
+    tem_case_locs = adf.get_cam_info("cam_tem_loc",required=True)
+    tem_base_loc = adf.get_baseline_info("cam_tem_loc")
 
-    #Location of saved TEM netCDF files
-    tem_loc = tem_opts.get("tem_loc")
+    #THSI IS IF IT IS MISSING!
+    #if tem_case_locs is not None:
+    #    if (len(tem_case_locs) == 1) and (any(tem_case_locs) is None):
+    #        tem_case_locs = None
```
nusbaume, line 82: I would vote to either un-comment or remove this code block, depending on whether you think this check is needed (or not).

scripts/plotting/tem.py
```diff
@@ -150,7 +162,6 @@ def tem(adf):
             # Check redo_plot. If set to True: remove old plot, if it already exists:
             if (not redo_plot) and plot_name.is_file():
                 #Add already-existing plot to website (if enabled):
-                adf.debug_log(f"'{plot_name}' exists and clobber is false.")
```
nusbaume, line 153: Is there a reason you removed this `debug_log` statement?

scripts/plotting/tem.py
```diff
@@ -107,20 +112,27 @@ def tem(adf):
...
+    if adf.compare_obs:
+        obs = True
         #Set TEM file for observations
-        base_file_name = f'{base_name}.TEMdiag.nc'
+        base_file_name = 'Obs.TEMdiag.nc'
     else:
-        #Set TEM file for baseline
-        base_file_name = f'{base_name}.TEMdiag_{syear_baseline}-{eyear_baseline}.nc'
+        obs = False
+        base_name = adf.get_baseline_info("cam_case_name", required=True)
```
nusbaume, line 122: I might put these lines underneath the `else` block of the `if tem_base_loc is None` check below.

scripts/plotting/tem.py
```diff
@@ -189,8 +200,6 @@ def tem(adf):
         #Add plot to website (if enabled):
         adf.add_website_data(plot_name, "TEM", case_name, season=s)
 
-    print("  ...TEM plots have been generated successfully.")
```
nusbaume, line 192: Is there a reason you got rid of this `print` statement?

nusbaume (changes requested):
Looks good, but did have some questions and a few change requests.

scripts/averaging/create_TEM_files.py
```diff
@@ -50,45 +75,60 @@ def create_TEM_files(adf):
...
-        start_years.append(syear_baseline)
-        end_years.append(eyear_baseline)
-        overwrite_tem_cases.append(tem_opts.get("overwrite_tem_base", False))
-    #End if
+        tem_base_loc = adf.get_basic_info('obs_tem_loc')
+        
+        #If path not specified, skip TEM calculation?
+        if tem_base_loc is None:
+            print("\t 'obs_tem_loc' not found in config file, so no TEM files/diagnostics will be generated.")
+            return
```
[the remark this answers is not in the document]
justin-richling: This is an oversight, thanks

scripts/averaging/create_TEM_files.py
```diff
@@ -50,45 +75,60 @@ def create_TEM_files(adf):
...
-    if hist_num is None:
-        hist_num = "h4"
+    else:
+        if tem_base_loc:
+            cam_hist_locs.append(adf.get_baseline_info("cam_hist_loc", required=True))
+
+            #Set default to h4
+            hist_num = adf.get_baseline_info("tem_hist_str")
+            if hist_num is None:
+                hist_num = "h4"
```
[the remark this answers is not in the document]
justin-richling: Good catch, this is the true hist_num section. The other one above has been removed

scripts/averaging/create_TEM_files.py
justin-richling: That was my intention, if the obs don't exist, the test case TEM plots should still be generated. Thanks

scripts/averaging/create_TEM_files.py
justin-richling: Giving it more thought, we should just remake these files each time in case the TEM obs files we have do change over time.

scripts/plotting/tem.py
justin-richling: I probably accidentally over-rid this line, thanks

scripts/plotting/tem.py
justin-richling: I left the `base_name` line so that the it's available in the print statement for the 
```python
if tem_base_loc is None
```
But I moved the `obs = False` to the `else` block

scripts/plotting/tem.py
justin-richling: I probably accidentally over-rid this line, thanks

justin-richling:
@nusbaume I think I addressed your comments/suggestions,. Thanks for looking this over, the changes are in when you get a chance to review it again. Thanks!

scripts/averaging/create_TEM_files.py
```diff
@@ -146,8 +127,6 @@ def create_TEM_files(adf):
         tem_fil = tem_base_loc / f'{base_name}.TEMdiag.nc'
```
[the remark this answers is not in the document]
justin-richling: @nusbaume Since we don't have a dedicated obs location for saved files, the best I could come up with quickly was just to save the obs TEM file to the same location as the first test case TEM file location. Let me know if you have thoughts for or against this for now.

justin-richling:
@nusbaume There was also a bug for the TEM plots if the file exists and redo plot is false, it created an empty plot and saved that to the web.

I've tested this with model vs model and model vs obs. If you don't mind just looking at these two changes again that'd be great.

nusbaume (approved):
Everything looks good to me now.  Thanks @justin-richling!

PR #282, opened 2024-01-30
Pre-made timeseries bug fix

justin-richling:
Currently if the user wants to use pre-made time series files, the ADF is trying to grab climo years from the given history files even if `cam_ts_done` variable is true because `cam_hist_loc` is always required. This will cause the ADF to fail as it tries to read from files that don't exist or have a different filename format that doesn't match CAM hist files.

This fix removes the requirement of `cam_hist_loc` so the user can now use pre-made time series files and not have to specify the `cam_hist_loc`. 

This also introduces a check in `adf_diag.py` if `cam_ts_done` is true then it will make `cam_hist_loc` none if it was provided, so that the climo years are not trying to be read in from the `cam_hist_loc`.

This will also change the ADF behavior that requires the user to specify climo years for pre-made ts files. Now if the years are not specified, they are read in via xarrray in `adf_info.py` via new function `get_climo_yrs_from_ts`. This will print a warning message if the year range is larger than 100 because the ADF has a hard time creating climo files on the order of hundreds of years.

Additionally, this addresses the scenario if any of the test cases in a multi-case run have pre-made time series files. The user still must have the same number of entries for `cam_hist_loc` as the rest of variables, but they can be left blank (but not commented out) for cases that are using pre-made ts files:

<h3>Works</h3>

```
cam_hist_loc:
        - /glade/derecho/scratch/tilmes/archive/b.e21.B1850WmaCARMA.f19_g17.carmats16_cesm2.2.001/atm/hist/
        - 
   .
   .
   .

cam_ts_done:
        - false
        - true
```

<h3>Does NOT Work</h3>

```
cam_hist_loc:
        - /glade/derecho/scratch/tilmes/archive/b.e21.B1850WmaCARMA.f19_g17.carmats16_cesm2.2.001/atm/hist/
        #- 
   .
   .
   .

cam_ts_done:
        - false
        - true
```

Finally, instead of killing the ADF if the tables script encounters CMIP-like files or multiple time series files in the same directory, it just quits the tables script and continues the ADF flow.

### DIFF
--- a/lib/adf_info.py
+++ b/lib/adf_info.py
@@ -166,12 +166,36 @@ class AdfInfo(AdfConfig):
 
             #Attempt to grab baseline start_years (not currently required):
             syear_baseline = self.get_baseline_info('start_year')
-            syear_baseline = int(f"{str(syear_baseline).zfill(4)}")
             eyear_baseline = self.get_baseline_info('end_year')
-            eyear_baseline = int(f"{str(eyear_baseline).zfill(4)}")
-
+            
             #Get climo years for verification or assignment if missing
             baseline_hist_locs = self.get_baseline_info('cam_hist_loc')
+
+            #Check if any time series files are pre-made
+            baseline_ts_done   = self.get_baseline_info("cam_ts_done")
+
+            #Check if time series files already exist, if so don't rely on climo years from history location
+            if baseline_ts_done:
+                baseline_hist_locs = None
+
+                #History file path isn't needed if user is running ADF directly on time series.
+                #So make sure start and end year are specified:
+                if syear_baseline is None or eyear_baseline is None:
+                    msg = f"No given start or end year for {data_name}"
+                    print(msg)
+
+                    #Grab baseline time series file location
+                    input_ts_baseline = self.get_baseline_info("cam_ts_loc", required=True)
+                    input_ts_loc = Path(input_ts_baseline)
+
+                    #Get years from pre-made timeseries file(s)
+                    syear_baseline, eyear_baseline = self.get_climo_yrs_from_ts(input_ts_loc, data_name)
+
+                    msg = f"Using first found year: {syear_baseline}\n"
+                    msg += f"and last found year: {eyear_baseline}"
+                    print(msg)
+                #End if
+            #End if
 
             #Check if history file path exists:
             if baseline_hist_locs:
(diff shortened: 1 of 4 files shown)


lib/adf_info.py
```diff
@@ -270,61 +296,84 @@ def __init__(self, config_file, debug=False):
...
                     print(msg)
-                    syear = int(case_climo_yrs[0])
+                    syear = case_found_syr
+                else:
+                    syear = int(syear)
                 #End if
                 if eyear is None:
-                    print(f"No given end year for {case_name}, using last found year...")
-                    eyear = int(case_climo_yrs[-1])
+                    msg = f"No given start year for {case_name}, "
```
nusbaume, line 366: I think this should be `end year`:

```suggestion
                    msg = f"No given end year for {case_name}, "
```

lib/adf_info.py
```diff
@@ -567,6 +616,62 @@ def add_diag_var(self, var_str):
...
+        for pre-made time series file(s)
+
+        Return
+        ------
+          - start year
+          - end year
+        """
+
+        #Grab variable list
+        var_list = self.__diag_var_list.copy()
```
nusbaume, line 634: I _think_ you can use the standard `diag_var_list` property here:

```suggestion
        var_list = self.diag_var_list
```

lib/adf_info.py
```diff
@@ -567,6 +616,62 @@ def add_diag_var(self, var_str):
...
+        input_location  = Path(input_ts_loc)
+
+        #Check that time series input directory actually exists:
+        if not input_location.is_dir():
+            errmsg = f"Time series directory '{input_ts_loc}' not found.  Script is exiting."
+            raise AdfError(errmsg)
+
+        #Search for first variable in var_list to get a time series file to read
+        #NOTE: it is assumed all the variables have the same dates!
+        ts_files = sorted(list(input_location.glob(f"{case_name}*.{var_list[0]}.*nc")))
```
nusbaume, line 646: Don't need to convert to a list (as `sorted()` does that automatically):

```suggestion
        ts_files = sorted(input_location.glob(f"{case_name}*.{var_list[0]}.*nc"))
```

scripts/analysis/amwg_table.py
```diff
@@ -204,7 +199,8 @@ def amwg_table(adf):
             if len(ts_files) != 1:
                 errmsg =  "Currently the AMWG table script can only handle one time series file per variable."
                 errmsg += f" Multiple files were found for the variable '{var}'"
-                raise AdfError(errmsg)
+                print(errmsg)
+                return
```
nusbaume, line 203: Might want to update this `errmsg` here to explicitly state that the tables will be skipped.

nusbaume (changes requested):
Thanks for tackling this bug @justin-richling!  I have some change requests, although the "bigger" requests can be saved for a future PR.  Of course if you have any questions or concerns with my comments just let me know.  Thanks!

justin-richling:
@nusbaume I think I've made the suggested changes, thanks for the attention to detail! 

One thing I did was expand a little more on the tables script. I've update the script so it will continue to the next variable if multiple time series files are found. I thought this would be good in an edge case where there are some variables that don't have multiple time series files, then the table can still be created. 

With that, if it is a case where all variables have multiple time series files, the script will continue and still attempt to create the test case table. This will be similar to comparing against obs; the website will still produce the test case table.

I also cleaned up the case comparison section for cmip like scenario where the table is skipped, it will search for the appropriate number of tables and if there aren't enough, the case comparison table is skipped too.

I also addressed the scenario if given pre-made ts climo years are not in the found range.

If you wouldn't mind reviewing it again that would great! I tested it for all these ADF scenarios:
* model vs model
* model vs pre-made ts
* model vs obs
* multi-case with pre-made ts as one of the test cases

lib/adf_info.py
```diff
@@ -310,22 +328,36 @@ def __init__(self, config_file, debug=False):
...
-                    #Grab case time series file location(s)
-                    input_ts_locs = self.get_cam_info("cam_ts_loc", required=True)
-                    input_ts_loc = Path(input_ts_locs[case_idx])
-
-                    #Get years from pre-made timeseries file(s)
-                    syear, eyear = self.get_climo_yrs_from_ts(input_ts_loc, case_name)
-
-                    msg = f"Using first found year: {syear}\n"
-                    msg += f"and last found year: {eyear}"
+                    syear_baseline = found_syear
```
nusbaume, line 344: I think this is supposed to be `syear` here:

```suggestion
                    syear = found_syear
```

lib/adf_info.py
```diff
@@ -310,22 +328,36 @@ def __init__(self, config_file, debug=False):
...
+                if syear not in found_yr_range:
+                    msg = f"Given start year '{syear}' is not in current dataset "
+                    msg += f"{case_name}, using first found year: {found_syear}\n"
                     print(msg)
+                    syear_baseline = found_syear
```
nusbaume, line 349: I think this is supposed to be `syear` here:

```suggestion
                    syear = found_syear
```

lib/adf_info.py
```diff
@@ -310,22 +328,36 @@ def __init__(self, config_file, debug=False):
...
+                #End if
+                if eyear is None:
+                    msg = f"No given end year for {case_name}, "
+                    msg += f"using last found year: {found_eyear}"
+                    print(msg)
+                    eyear = found_eyear
+                if eyear not in found_yr_range:
+                    msg = f"Given end year '{eyear}' is not in current dataset "
+                    msg += f"{case_name}, using first found year: {found_eyear}\n"
```
nusbaume, line 358: I think you meant to say `last` here:

```suggestion
                    msg += f"{case_name}, using last found year: {found_eyear}\n"
```

lib/adf_info.py
```diff
@@ -354,30 +386,31 @@ def __init__(self, config_file, debug=False):
...
-                elif (eyear) not in case_climo_yrs:
+                if eyear not in case_climo_yrs:
                     msg = f"Given end year '{eyear}' is not in current dataset "
                     msg += f"{case_name}, using last found year: {case_climo_yrs[-1]}\n"
                     print(msg)
                     eyear = case_found_eyr
-                else:
-                    eyear = int(eyear)
+                #else:
+                #    eyear = int(eyear)
```
nusbaume, line 407: I would remove this commented-out code.

scripts/analysis/amwg_table.py
```diff
@@ -287,39 +290,45 @@ def amwg_table(adf):
...
+            if 'RESTOM' in table_df['variable'].values:
+                table_df = pd.concat([table_df[table_df['variable'] == 'RESTOM'], table_df]).reset_index(drop = True)
+                table_df = table_df.drop_duplicates()
+                table_df.to_csv(output_csv_file, header=cols, index=False)
 
-        # last step is to add table dataframe to website (if enabled):
-        table_df = pd.read_csv(output_csv_file)
-        adf.add_website_data(table_df, case_name, case_name, plot_type="Tables")
+            # last step is to add table dataframe to website (if enabled):
+            table_df = pd.read_csv(output_csv_file)
```
nusbaume, line 303: I think `table_df` is already read in above, so I believe this line isn't needed (?).

nusbaume (changes requested):
Thanks @justin-richling, almost there!  I just found a few last possible bugs.

justin-richling:
@nusbaume Thanks again, these have been fixed!

nusbaume (approved):
Everything looks good to me now!

PR #283, opened 2024-02-08
Add scale factor to LWP obs to convert to g m-2.

nusbaume:
Adds an obs scale factor to the `TGCLDLWP` variable to ensure that both the model and the observations are in units of g m-2.  

Credit goes to @bstephens82 for finding and reporting the bug!


bstephens82:
Thanks @nusbaume for creating this PR.  Just wanted to add the same change should probably be applied to `TGCLDIWP` as it has the same problem.

nusbaume:
@bstephens82 Good catch!  The scale factor has now been added to `TGCLDIWP` as well.

justin-richling (approved):
Looks good, thanks for the additional plot details @bstephens82!

PR #286, opened 2024-02-23
New adf updates

justin-richling:
### Generalize load dataset

Make `load_dataset` function in `plotting_functions.py` that replaces `_load_dataset` function in various plotting scripts.
Now all plotting scripts will call this function from `plotting_functions.py` to get datasets.

### Website

In `adf_web.py`, add a list of defualt plot types and loop over any additional plot types. The point of this is to show all plot types on the landing page (`index.html`) and grey out any unavailable types. This will keep the main page consistent as well as always showing potential ADF plot types.

Also in `adf_web.py`, the check for mean_diag html file existence is removed. The reasoning is if a certain plot type (ie LatLon) is created from separate plotting scripts, it will only produce the plots from the last instance on the webpage.

In `template_index.html` there is code now that will grey out any unavailable plot types.

### Tape Recorder

Currently there is a bug that is not producing the correct html plot page for the tape recorder. The plot name and web object has been updated to reflect the variable Q specifically in the case a different variable is desired for this plot in the future, like temperature, etc.

### DIFF
--- a/lib/adf_variable_defaults.yaml
+++ b/lib/adf_variable_defaults.yaml
@@ -63,6 +63,13 @@
 #+++++++++++
 
 #+++++++++++++
+# Available ADF Default Plot Types
+#+++++++++++++
+
+default_ptypes: ["Tables","LatLon","LatLon_Vector","Zonal","Meridonal",
+                  "NHPolar","SHPolar","Special"]
+
+#+++++++++++++
 # Category: Microphysics
 #+++++++++++++
 
--- a/lib/adf_variable_defaults.yaml
+++ b/lib/adf_variable_defaults.yaml
@@ -66,7 +66,7 @@
 # Available ADF Default Plot Types
 #+++++++++++++
 
-default_ptypes: ["Tables","LatLon","LatLon_Vector","Zonal","Meridonal",
+default_ptypes: ["Tables","LatLon","LatLon_Vector","Zonal","Meridional",
                   "NHPolar","SHPolar","Special"]
 
 #+++++++++++++
--- a/lib/adf_web.py
+++ b/lib/adf_web.py
@@ -652,21 +652,20 @@ class AdfWeb(AdfObs):
 
                 #Check if the mean plot type page exists for this case:
                 mean_ptype_file = img_pages_dir / f"mean_diag_{web_data.plot_type}.html"
-                if not mean_ptype_file.exists():
-                    #Construct individual plot type mean_diag html files, if they don't
-                    #already exist:
-                    mean_tmpl = jinenv.get_template('template_mean_diag.html')
 
-                    #Remove keys from main dictionary for this html page
-                    templ_rend_kwarg_dict = {k: rend_kwarg_dict[k] for k in rend_kwarg_dict.keys() - {'imgs', 'var_title', 'season_title'}}
-                    templ_rend_kwarg_dict["list"] = jinja_list
-                    mean_rndr = mean_tmpl.render(templ_rend_kwarg_dict)
+                #Construct individual plot type mean_diag html files, if they don't
+                #already exist:
+                mean_tmpl = jinenv.get_template('template_mean_diag.html')
 
-                    #Write mean diagnostic plots HTML file:
-                    with open(mean_ptype_file,'w', encoding='utf-8') as ofil:
-                        ofil.write(mean_rndr)
-                    #End with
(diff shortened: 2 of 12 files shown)


lib/plotting_functions.py
```diff
@@ -123,6 +133,34 @@
...
+    When just one entry is provided, use `open_dataset`, otherwise `open_mfdatset`
+    """
+    if len(fils) == 0:
+        warnings.warn(f"Input file list is empty.")
+        return None
+    elif len(fils) > 1:
+        return xr.open_mfdataset(fils, combine='by_coords')
+    else:
+        sfil = str(fils[0])
+        return xr.open_dataset(sfil)
```
nusbaume, line 160: Does `fils` have to be converted to a string here?  I thought the list was already assumed to contain either strings or path objects?  If not then we can simplify these lines to be:

 ```suggestion
        return xr.open_dataset(fils[0])
```

File: scripts/analysis/amwg_table.py [hunk 2]
```diff
@@ -206,8 +206,9 @@ def amwg_table(adf):
                 continue
             #End if
 
-            #Load model data from file:
-            data = _load_data(ts_files[0], var)
+            #Load model variable data from file:
+            ds = pf.load_dataset(ts_files)
```
nusbaume, line 210: Do we want to only grab the first entry in `ts_files` here, or was the original way incorrect for using `ts_files[0]`?

Also with this change can we now delete the `_load_data` function from this python file?

scripts/plotting/qbo.py
```diff
@@ -98,7 +99,7 @@ def qbo(adfobj):
 
     #----Read in the case data and baseline
     ncases = len(case_loc)
-    casedat = [ _load_dataset(case_loc[i], case_names[i],'U') for i in range(0,ncases,1) ]
+    casedat = [pf.load_dataset(sorted(Path(case_loc[i]).glob(f"{case_names[i]}.*.U.*.nc"))) for i in range(0,ncases,1)]
```
nusbaume, line 102: Can we now remove the `_load_dataset` function from this python file?

nusbaume (changes requested):
Very nice cleanup!  Just had a few questions/requests.

scripts/analysis/amwg_table.py
justin-richling: It should still be good. The old way of reading in files in this script was just `open_dataset` so the file needed to be pulled from the glob list. In the new `load_dataset` in `plotting_functions`, it has that built in in the function (your comment just before this one).

lib/adf_web.py
```diff
@@ -681,6 +680,16 @@ def jinja_list(seas_list):
             plot_types = plot_type_html
             #End if
 
+            #List of ADF default plot types
+            avail_plot_types = ["Tables","LatLon","LatLon_Vector","Zonal","Meridonal",
+                                "NHPolar","SHPolar","Special"]
```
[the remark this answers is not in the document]
justin-richling: That's good idea, especially in the viewpoint of CUPiD

justin-richling:
@nusbaume I think I got these taken care of. Let me know if you see or think of anything else. Thanks!

lib/adf_variable_defaults.yaml
```diff
@@ -62,6 +62,13 @@
 #
 #+++++++++++
 
+#+++++++++++++
+# Available ADF Default Plot Types
+#+++++++++++++
+
+default_ptypes: ["Tables","LatLon","LatLon_Vector","Zonal","Meridonal",
```
nusbaume, line 69: Typo here:

```suggestion
default_ptypes: ["Tables","LatLon","LatLon_Vector","Zonal","Meridional",
```

lib/adf_web.py
```diff
@@ -653,8 +656,7 @@ def jinja_list(seas_list):
                 #Check if the mean plot type page exists for this case:
```
nusbaume, line 656: I might change this comment too given that we aren't actually checking if the plot type page exists anymore.

scripts/analysis/amwg_table.py
nusbaume: Right!  That makes sense.  Thanks for the explanation!

nusbaume (approved):
Looks good!  Just had a couple of final suggestions (but I don't need to re-review).

PR #280, opened 2024-01-23
Add temp try/except for plotting errors

justin-richling:
Placeholder until the random bug is fixed.

I'm working through several suggested fixes and also reaching out to cartopy and matplotlib. But this will at least allow the ADF to continue to run.

PR #287, opened 2024-02-26
Fix bad polygon error in plotting

justin-richling:
Add `transform_first` call to cartopy contour plots that need transforming.


nusbaume:
@justin-richling Does this PR mean PR #280 is no longer needed?

PR #293, opened 2024-03-20
Call MDTF from ADF, and allow list of hist_str to be processed for timeseries

bitterbark:
1. Call MDTF from ADF

This adds a section in the config and run files so the ADF can directly call the [MDTF (Model Diagnostics Task Force (MDTF)-Diagnostics package) ](https://www.gfdl.noaa.gov/mdtf-diagnostics/). 

This is the simplest possible implementation that calls an existing, installed version of the MDTF using existing, installed conda envs.  For nowl, timeseries files will be copied to the requested MDTF directory and file names. Imminent data catalogue functionality in MDTF will change this.

Basic documentation of new functionality:
in config.yaml, there is a new section:
> diag_mdtf_info:
>     # Run the MDTF on the model cases
>     mdtf_run: false
> 

If `mdtf_run:true`:

> run_adf_diag
>     #Call the MDTF:
>     if diag.get_mdtf_info('mdtf_run'):
>        diag.setup_run_mdtf()
> 

[Detailed documentation of implementation/call tree]
(https://www2.cgd.ucar.edu/cms/bundy/Projects/diagnostics/mdtf/notes/run_mdtf_from_adf.html#Documentation)
More [documentation for MDTF](https://mdtf-diagnostics.readthedocs.io/en/main/)

2.  Allow `hist_str` to be a list 
The config.yaml file variable can now also be a list:
    hist_str: [cam.h2,cam.h0]

These timeseries files are all placed in the same ts/ directory, which may need to be modified for multiple component hist strings. (The MDTF checks file types for time frequency for its own purposes).


bitterbark:
@nusbaume  I'm trying to fix the errors/warnings but I'm confused by the very first 
> lib/adf_diag.py:373:8: W0601: Global variable 'call_ncrcat' undefined at the module level (global-variable-undefined)

There is no between mine and the upstream repo.

> 
>     def create_time_series(self, baseline=False):
>         """                                                                                                                             
>         Generate time series versions of the CAM history file data.                                                                     
>         """
> 
>         global call_ncrcat
> 
>         def call_ncrcat(cmd):
> 

Do I need to do something?

lib/adf_diag.py
```diff
@@ -1080,3 +1108,227 @@ def derive_variables(self, vars_to_derive=None, ts_dir=None, overwrite=None):
...
+                        adf_file, decode_cf=False, decode_times=False
+                    )
+                    if "time_period_freq" in hist_file_ds.attrs:
+                        dataset_freq = hist_file_ds.attrs["time_period_freq"]
+                        if verbose > 2:
+                            print(f"time_period_freq attribute found: {dataset_freq}")
+                    else:
+                        if verbose > 0:
+                            print(
+                                f"ERROR: Necessary 'time_period_freq' attribute missing from {adf_file}"
```
nusbaume, line 1182: I would change this message to the following:

```suggestion
                                f"WARNING: Necessary 'time_period_freq' attribute missing from {adf_file}. Skipping file."
```

lib/adf_diag.py
```diff
@@ -1080,3 +1108,227 @@ def derive_variables(self, vars_to_derive=None, ts_dir=None, overwrite=None):
...
+                    elif len(found_strings) > 1:
+                        if verbose > 0:
+                            print(
+                                f"WARNING: Found dataset_freq {dataset_freq} matches multiple string possibilities:{', '.join(found_strings)}"
+                            )
+                    else:
+                        if verbose > 0:
+                            print(
+                                "ERROR: None of the frequency options {freq_string_options} are present in the time_period_freq attribute {time_perido_freq}"
+                            )
```
nusbaume, line 1203: I would change this message to `WARNING`, as `ERROR` usually implies that the ADF must stop completely.  Also I believe you are missing the `f` string here:

```suggestion
                            print(
                                f"WARNING: None of the frequency options {freq_string_options} are present in the time_period_freq attribute {time_perido_freq}"
                            )
```

lib/adf_diag.py
```diff
@@ -1080,3 +1108,227 @@ def derive_variables(self, vars_to_derive=None, ts_dir=None, overwrite=None):
...
+                            print("Skipping {adf_file}")
```
nusbaume, line 1204: Missing `f` label here:

```suggestion
                            print(f"Skipping {adf_file}")
```

lib/adf_diag.py
```diff
@@ -1080,3 +1108,227 @@ def derive_variables(self, vars_to_derive=None, ts_dir=None, overwrite=None):
...
+                            freq = "frequency_missing"
+                        continue
+                    freq = found_strings[0]
+
+                    #
+                    # Destination file is MDTF directory and name structure
+                    #
+                    mdtf_dir = os.sep.join([mdtf_model_data_root, case_name, freq])
```
nusbaume, line 1212: I might change this line to the following:

```suggestion
                    mdtf_dir = os.path.join(mdtf_model_data_root, case_name, freq)
```

lib/adf_diag.py
```diff
@@ -1080,3 +1108,227 @@ def derive_variables(self, vars_to_derive=None, ts_dir=None, overwrite=None):
...
+                        mdtf_file
+                    )  # Check if files already exist in MDTF directory
+                    if (
+                        mdtf_file_list
+                    ):  # If files exist, then should check if over-writing is allowed:
+                        if verbose > 0:
+                            print(
+                                f"\t   INFO: not clobbering existing mdtf file {mdtf_file_list}"
+                            )
+                        # logic to allow override#  if not overwrite_mdtf[case_idx]:
```
nusbaume, line 1228: Is there supposed to be overwrite logic here, or is this a future effort?

lib/adf_diag.py
```diff
@@ -1080,3 +1108,227 @@ def derive_variables(self, vars_to_derive=None, ts_dir=None, overwrite=None):
...
+                    if verbose > 1:
+                        print(f"copying {adf_file} to {mdtf_file}")
+                    shutil.copyfile(adf_file, mdtf_file)
+                # end for hist_str
+            # end for var
+        # end for case
+
+    def setup_run_mdtf(self):
+        """
+        Create MDTF directory tree, generate input setttings jsonc file
```
nusbaume, line 1240: Typos here:

```suggestion
        Create MDTF directory tree, generate input settings json file
```

lib/adf_diag.py
```diff
@@ -1080,3 +1108,227 @@ def derive_variables(self, vars_to_derive=None, ts_dir=None, overwrite=None):
...
+        Submit MDTF diagnostics.
+
+        """
+
+        # import needed standard modules:
+        import json
```
nusbaume, line 1246: I might move this to the top of `adf_diag.py` where the rest of the import statements are.

lib/adf_diag.py
```diff
@@ -1080,3 +1108,227 @@ def derive_variables(self, vars_to_derive=None, ts_dir=None, overwrite=None):
...
+
+        test = False  # True (copy files but don't run), False (copy files and run MDTF).  Could make this a yaml variable
+        verbose = 2  # 0 errors and major sections, 1 warnings, 2 detailed, 3 excessive
```
nusbaume, line 1249: Aren't these both YAML config variables already?

lib/adf_diag.py
```diff
@@ -1080,3 +1108,227 @@ def derive_variables(self, vars_to_derive=None, ts_dir=None, overwrite=None):
...
+        #     This could be changed by inputing them into ADF with other MDTF-specific variables
+        #
+        case_list_keys = ["CASENAME", "FIRSTYR", "LASTYR", "model", "convention"]
+
+        # Casenames, paths and start/end years come through the ADF
+        case_names = self.get_cam_info("cam_case_name", required=True)
+        start_years = self.climo_yrs["syears"]
+        end_years = self.climo_yrs["eyears"]
+
+        case_list_all = list(())
```
nusbaume, line 1267: Too many parentheses here, also better to just use the symbolic list representation:

```suggestion
        case_list_all = []
```

lib/adf_diag.py
```diff
@@ -1080,3 +1108,227 @@ def derive_variables(self, vars_to_derive=None, ts_dir=None, overwrite=None):
...
+            case_list_all.append(dict(zip(case_list_keys, case_list_values)))
+        mdtf_info["case_list"] = (
+            case_list_all  # this list of dicts is the format wanted by MDTF
+        )
+
+        # The plot_path is given by case in ADF but MDTF needs one top dir, so use case 0
+        # Working dir and output dir can be different. These could be set in config.yaml
+        # but then we don't get the nicely formated plot_location
+        case_idx = 0
+        plot_path = self.plot_location[case_idx] + "/mdtf"
```
nusbaume, line 1285: Probably safer to just use `os.path.join` here:

```suggestion
        plot_path = os.path.join(self.plot_location[case_idx], "mdtf")
```

lib/adf_diag.py
```diff
@@ -1080,3 +1108,227 @@ def derive_variables(self, vars_to_derive=None, ts_dir=None, overwrite=None):
...
+        for var in ["WORKING_DIR", "OUTPUT_DIR"]:
+            if mdtf_info[var] == "default":
+                mdtf_info[var] = plot_path
```
nusbaume, line 1288: What should the ADF do if these config variables are not set to `default`?

lib/adf_diag.py
```diff
@@ -1080,3 +1108,227 @@ def derive_variables(self, vars_to_derive=None, ts_dir=None, overwrite=None):
...
+        #    model_input_data/case/freq/case.VAR.freq.nc
+
+        self.move_tsfiles_for_mdtf(verbose)
+
+        #
+        # Submit the MDTF script in background mode, send output to mdtf.out file
+        #
+        mdtf_log = os.path.join(
+            "./mdtf.out"
+        )  # maybe set this to cam_diag_plot_loc: /glade/scratch/${user}/ADF/plots
```
nusbaume, line 1317: If `mdtf.out` is being written to the current local directory then this can just be:

```suggestion
        mdtf_log = "mdtf.out" # maybe set this to cam_diag_plot_loc: /glade/scratch/${user}/ADF/plots
```

lib/adf_diag.py
```diff
@@ -1080,3 +1108,227 @@ def derive_variables(self, vars_to_derive=None, ts_dir=None, overwrite=None):
...
+        mdtf_exe = mdtf_codebase + "/mdtf -f " + mdtf_input_settings_filename
```
nusbaume, line 1318: I would use `os.sep` here:

```suggestion
        mdtf_exe = mdtf_codebase + os.sep + "mdtf -f " + mdtf_input_settings_filename
```

nusbaume (changes requested):
Thanks for the new features and all the extra clean-up @bitterbark!  However, I did have some questions and change requests, which would be good to resolve, or at least discuss, before this PR is merged in.  Please let me know if you have any questions or concerns with any of my requests!

lib/adf_diag.py
bitterbark: These are set in the input jsonc file to the MDTF
> 	mdtf_info = self.__mdtf_info
And so their values are read in.  I added the default keyword so the directory doesn't have to be copied into the input file.

lib/adf_diag.py
bitterbark: We do use jsonc format.

lib/adf_info.py
```diff
@@ -207,136 +220,144 @@ def __init__(self, config_file, debug=False):
...
-                    base_nickname = data_name
-            #End if
-
-            #Grab baseline nickname
-            base_nickname = self.get_baseline_info('case_nickname')
+            # End if
+
+            # Check if history file path exists:
+            if any(baseline_hist_locs):
+                for hist_str in baseline_hist_str:
```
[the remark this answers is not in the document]
bitterbark: You're right in that the values aren't retained. It turns out it doesn't need                                                               
to be a loop because we assume all the files in a case have the same dates,                                                         
and we are only using one syear/eyear per case, not per hist_str.                                                                   
So I've changed this to                                                                                                             
 ```
               hist_str = baseline_hist_str[0]                                                                                     
                hist_str = hist_str_case[0]                
```                                                                         
and removed the for-loop over hist_str.                                                                                             
(Will commit with the other changes)

File: lib/adf_info.py [hunk 15]
```diff
@@ -358,142 +379,158 @@ def __init__(self, config_file, debug=False):
...
+                # End except
+            # End if
+        # End if
+        # Print number of processors being used to debug log (if requested):
         self.debug_log(f"ADF is running with {self.__num_procs} processors.")
-        #-----------------------------------------
+        # -----------------------------------------
+
+    #########
+    def hist_str_to_list(self, conf_var, conf_val):
```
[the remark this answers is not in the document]
bitterbark: done

lib/adf_diag.py
bitterbark: Good point. The verbose is so I modified this to use it instead. There is a test_mode in MDTF but that is different from what I'm doing here. I made a comment as such in the file and changed the name of this 'test' to be more clear:

> copy_files_only = False  # True (copy files but don't run), False (copy files and run MDTF)
        # Note that the MDTF variable test_mode (set in the mdtf_info of the yaml file) 
        # has a different meaning: Data is fetched but PODs are not run.
...
        verbose = mdtf_info["verbose"]

lib/adf_diag.py
```diff
@@ -170,6 +171,13 @@ def __init__(self, config_file, debug=False):
             self.expand_references(self.__cvdp_info)
         # End if
 
+        # Add MDTF info to object:
+        self.__mdtf_info = self.read_config_var("diag_mdtf_info")
+
+        if self.__mdtf_info is not None:
+            self.expand_references(self.__mdtf_info)
+        # End if
```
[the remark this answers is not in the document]
bitterbark: Should I move the CVDP calls too?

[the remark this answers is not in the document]
bitterbark: I moved the MDTF and CVDP calls. There was a remaining reference in adf_diag.py in which I need to just get the whole mdtf_info dict in order to dump it all into the json for MDTF. So I added a small change in 
> adf_config.py: read_config_var():                                                                                                                                           
>         if varname == "ALL":                                                                                                                                         
>             return var_dict    
I hope this is okay. It speaks (again) to the great design of this code that I could add two lines to do this and like so many thing in ADF, it just worked!

lib/adf_diag.py
```diff
@@ -382,12 +406,9 @@ def call_ncrcat(cmd):
...
 
         # Read hist_str (component.hist_num) from the yaml file, or set to default
-        hist_str = self.get_basic_info("hist_str")
-        # If hist_str is not present, then default to 'cam.h0':
-        if not hist_str:
-            hist_str = "cam.h0"
-        # End if
-
+        hist_str_list = self.get_cam_info("hist_str")
+        print(f"reading from {hist_str_list} files")
```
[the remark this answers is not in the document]
bitterbark: I chose the adf.debug_log option.

lib/adf_diag.py
bitterbark: Am I wrong that there *is* overwrite logic?
if the file is present
then continue 
else copy file

> mdtf_file_list = glob.glob(mdtf_file)  # Check if files already exist in MDTF directory
                    if (mdtf_file_list):  # If files exist, then should check if over-writing is allowed:
                        if verbose > 0:
                            print(
                                f"\t   INFO: not clobbering existing mdtf file {mdtf_file_list}"
                            )
                        # logic to allow override#  if not overwrite_mdtf[case_idx]:
                        continue  # simply skip file copy for this variable:

bitterbark: Oh, I see where I was misunderstanding. Yes, there should be logic saying over-write or don't. At the moment, the user can just remove the files if they want them re-copied.

bitterbark:
Github and I did not get along very well through this process, so in order to remove the final conflicts, I had to use the web version of the file. This might have been better if it hadn't involved indentations, but I did what I could. However, I was unable to test it. (I merged the code over on my branch but again, something went wonky, so I couldn't even just throw in a new commit). So the problem is: I literally haven't been able to run this code to test it before hitting commit.

Another issue: the reformatting I had done in order to pass the linting tests was sometimes thrown out in the merge. I guess Justin's code is so clean that he can get away with these problems? I am reluctant to delay getting this in any longer as re-merging can take a week of my time. Can I get a pass on those?

nusbaume:
Hi @bitterbark, it looks like the Github test failure is due to an actual python syntax error:

```
lib/adf_diag.py:491:20: E0001: Parsing failed: 'unindent does not match any outer indentation level (adf_diag, line 491)' (syntax-error)
```

So we'll have to fix it in order for this PR to be accepted (otherwise the ADF itself would break).

After that do you think this PR is ready for a re-review?  If so I'll check it over and hopefully have @justin-richling run a test to make sure it runs as expected.  If everything goes well we can hopefully get this PR merged in soon!

bitterbark:
Unfortnately the linter found an indent error. I am going back and merging by hand. Will update with a fresh pull request.

bitterbark:
Closing this in lieu of new [PR](https://github.com/NCAR/ADF/pull/310)

PR #313, opened 2024-06-24
Small fixes

### LINKED ISSUE #289: model data paths in example config files are out of date.
### ADF run type

Model vs. Model

### What happened?

As found by @bitterbark, many of the paths in `config_cam_baseline_example.yaml` and `config_amwg_default_plots.yaml` are using `/glade/p`, which no longer exists on Derecho.  Instead we need to update those paths to point to files somewhere in `/glade/campaign`.

### ADF Hash you are using

b961522

CISL machine

NPL (CISL machines only)

### Extra info

This would also be a good time to update the example simulation data itself to point to some newer simulations, possible with model output that allows for the full range of diagnostics currently in the ADF?

justin-richling:
Update the config yaml files for updated example CAM cases and history files. 

&nbsp; &nbsp; There are now 4 example coupled cases in: `/glade/campaign/cgd/amp/amwg/ADF_test_cases`:
&emsp;&nbsp;&nbsp;  - 3 recent cases with h0a history files
&emsp;&nbsp;&nbsp;  - 1 semi-recent case with old h0 history files

&nbsp; &nbsp; Each with 10 years of data

&nbsp; &nbsp; Additionally, update paths for Derecho scratch locations and update `hist_str` argument to default to `cam.h0a`.

In `tape_recorder.py`, add check for scenario where no test case data is available and exit script if so.

Closes #289

### DIFF
--- a/config_amwg_default_plots.yaml
+++ b/config_amwg_default_plots.yaml
@@ -66,8 +66,8 @@ diag_basic_info:
     #History file string to match (eg. cam.h0 or ocn.pop.h.ecosys.nday1)
     # Only affects timeseries as everything else uses timeseries
     # Leave off trailing '.'
-    #Default: cam.h0
-    hist_str: cam.h0
+    #Default: cam.h0a
+    hist_str: cam.h0a
 
     #Is this a model vs observations comparison?
     #If "false" or missing, then a model-model comparison is assumed:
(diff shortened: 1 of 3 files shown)


nusbaume (approved):
Looks good, just two (optional) change requests.  Thanks @justin-richling!

PR #316, opened 2024-07-01
Update history string objects

### LINKED ISSUE #315: Tape Recorder plots failing after recent history string update
### ADF run type

Model vs. Model

### What happened?

After PR #310 the Tape Recorder plots are failing due to incorrect ADF object history string returns.

```python
File "/glade/work/richling/ADF/ADF/lib/../scripts/plotting/tape_recorder.py", line 40, in tape_recorder
    hist_str = adfobj.hist_string
  File "/glade/work/richling/ADF/ADF/lib/adf_info.py", line 636, in hist_string
    return self.__hist_str
AttributeError: 'AdfDiag' object has no attribute '_AdfInfo__hist_str'
```

### ADF Hash you are using

c715fc11

CISL machine

NPL (CISL machines only)

### Extra info

I have a fix, will create PR

justin-richling:
Clean up the history string objects in `adf_info.py`:
 - clean up naming of history string variables

 - Make baseline history info more robust; set default if not provided
 - Give full dictionary of test and baseline case(s) history string lists to adf object 

Update `tape_recorder.py` for new history string format:
 - Add check in history string lists for h0 time series files

Closes #315

### DIFF
--- a/scripts/plotting/tape_recorder.py
+++ b/scripts/plotting/tape_recorder.py
@@ -46,16 +46,15 @@ def tape_recorder(adfobj):
     cam_hist_strs = adfobj.hist_string["test_hist_str"]
 
     # Filter the list to include only strings that are exactly in the possible h0 strings
-    # - Search for either h0 of h0a
-    substrings = ["cam.h0","cam.h0a"]
+    # - Search for either h0 or h0a
+    substrings = {"cam.h0","cam.h0a"}
     case_hist_strs = []
-    for i,cam_case_str in enumerate(cam_hist_strs):
-        cam_hist_str = []
+    for cam_case_str in cam_hist_strs:
         # Check each possible h0 string
         for string in cam_case_str:
             if string in substrings:
-                cam_hist_str.append(string)
-        case_hist_strs.append(cam_hist_str[0])
+               case_hist_strs.append(string)
+               break
 
     #Grab test case climo years
     start_years = adfobj.climo_yrs["syears"]
--- a/scripts/plotting/tape_recorder.py
+++ b/scripts/plotting/tape_recorder.py
@@ -36,14 +36,26 @@ def tape_recorder(adfobj):
     plot_location = adfobj.plot_location
     plot_loc = Path(plot_location[0])
 
-    #Grab history string:
-    hist_str = adfobj.hist_string
-
     #Grab test case name(s)
     case_names = adfobj.get_cam_info('cam_case_name', required=True)
 
     #Grab test case time series locs(s)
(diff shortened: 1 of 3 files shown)


scripts/plotting/tape_recorder.py
```diff
@@ -36,15 +36,27 @@ def tape_recorder(adfobj):
...
     case_names = adfobj.get_cam_info('cam_case_name', required=True)
 
     #Grab test case time series locs(s)
     case_ts_locs = adfobj.get_cam_info("cam_ts_loc", required=True)
 
+    #Grab history strings:
+    cam_hist_strs = adfobj.hist_string["test_hist_str"]
+
+    # Filter the list to include only strings that are exactly in the possible h0 strings
+    # - Search for either h0 of h0a
```
nusbaume, line 49: Typo:

```suggestion
    # - Search for either h0 or h0a
```

scripts/plotting/tape_recorder.py
```diff
@@ -36,15 +36,27 @@ def tape_recorder(adfobj):
...
+    substrings = ["cam.h0","cam.h0a"]
+    case_hist_strs = []
+    for i,cam_case_str in enumerate(cam_hist_strs):
```
nusbaume, line 52: I believe `substrings` can be a set datatype, also it doesn't look like there is a need to `enumerate`:

```suggestion
    substrings = {"cam.h0","cam.h0a"}
    case_hist_strs = []
    for cam_case_str in cam_hist_strs:
```

scripts/plotting/tape_recorder.py
```diff
@@ -36,15 +36,27 @@ def tape_recorder(adfobj):
...
+        cam_hist_str = []
+        # Check each possible h0 string
+        for string in cam_case_str:
+            if string in substrings:
+                cam_hist_str.append(string)
+        case_hist_strs.append(cam_hist_str[0])
```
nusbaume, line 58: If you are just trying to find the first match then I believe this code block can be simplified to:

```suggestion
        # Check each possible h0 string
        for string in cam_case_str:
            if string in substrings:
               case_hist_strs.append(string)
               break
```

nusbaume (changes requested):
Looks good, but I did have a few questions and (hopefully minor) change requests.

justin-richling:
@nusbaume I got your suggestions done, but it looks like we still have some issues with the new history string implementation; the baseline history strings were not being brought over from the yaml file correctly. 

I have a quick addition in `adf_diag` that should take care of this bug, so if you wouldn't mind taking another quick look, thanks!

nusbaume (approved):
Everything looks great now, thanks @justin-richling!

PR #317, opened 2024-07-11
Dataclass fixes

justin-richling:
Fix issues from recent data class PR #269:

`adf_info.py`

* Add check for existing variable file for climo year info.
* Update hist_strs object.

`global_latlon_map.py`
* Fix check for vertical levels, initially based on `pres_levs` instead of `has_lev`.
* Update data grab to include obs.

`zonal_mean.py`
* Fix check for vertical levels, and reorganize plotting block. Only log-p plots were being made for variables with vertical levels.
* Update data grab to include obs.

`adf_dataset.py`
* Reorganize methods into file categories: time series, climo, regrid and general data load methods.
* These methods weren't working with obs, fix reference methods to correctly gather obs data.

`tape_recorder.py`:

* Bring in changes from PR #316.
* Update time fix for old CESM time series files via `adf_dataset.py` structure.

### DIFF
--- a/lib/adf_dataset.py
+++ b/lib/adf_dataset.py
@@ -1,3 +1,4 @@
+from dataclasses import asdict
 from pathlib import Path
 import xarray as xr
 
@@ -59,9 +60,11 @@ class AdfData:
         self.ref_nickname = self.base_nickname
 
         # define reference data
-        self.set_reference() # specify "ref_labels" -> called "data_list" in zonal_mean (name of data source)
+        self.set_reference(init=True) # specify "ref_labels" -> called "data_list" in zonal_mean (name of data source)
 
-    def set_reference(self):
+    # Reference case setup (baseline/obs)
+    #------------------------------------
+    def set_reference(self, init=False):
         """Set attributes for reference (aka baseline) data location, names, and variables."""
         if self.adf.compare_obs:
             self.ref_var_loc = {v: self.adf.var_obs_dict[v]['obs_file'] for v in self.adf.var_obs_dict}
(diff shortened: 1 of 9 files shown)


justin-richling:
@brianpm I needed to make some adjustments to the methods in `adf_dataset.py`, let me know if you want to go over what I changed, thanks.

lib/adf_dataset.py
```diff
@@ -59,9 +59,11 @@ def __init__(self, adfobj):
         self.ref_nickname = self.base_nickname
 
         # define reference data
-        self.set_reference() # specify "ref_labels" -> called "data_list" in zonal_mean (name of data source)
+        self.set_reference(init=True) # specify "ref_labels" -> called "data_list" in zonal_mean (name of data source)
```
justin-richling, line 62: I added `init` argument here to avoid printing that the climo files don't exist on the initial run of the ADF.

brianpm:
@justin-richling -- Let's discuss the changes. I think they are mostly fine. 

There are two things that I'm not liking at first glance. First is the method `load_obs_climo_dataset` -- I don't see why we need to make "obs" a special method instead of only using "ref". If possible, I'd like to only ever deal with "ref" (and especially, I'd like script-writers to only have to worry about ref and never distinguish between obs and ref).

Second is related, and is the logic in `load_da`. My thinking is that `load_da` shouldn't care what kind of data is getting loaded. I'm thinking this logic should be pushed into the methods that call `load_da`.  But it looks like you made the opposite determination and moved this logic into `load_da`... let's talk about it and decide how to deal with it.

lib/adf_dataset.py
```diff
@@ -76,87 +78,19 @@ def set_reference(self):
             # when using a reference simulation, allow a "special" attribute with the case name:
             self.ref_case_label = self.adf.get_baseline_info("cam_case_name", required=True)
             for v in self.adf.diag_var_list:
+                self.ref_var_nam[v] = v
```
justin-richling, line 81: I pulled these out the else statement because initially these weren't being populated because the check: 

```
f = self.get_reference_climo_file(v)
if f is None:
```

`f` will be none until the climo files are made, but we need ``` self.ref_var_nam``` and ```self.ref_labels``` before we get to climo file generation.

lib/adf_info.py
```diff
@@ -773,16 +776,18 @@ def get_climo_yrs_from_ts(self, input_ts_loc, case_name):
...
-            hist_str = 'cam.h0'
+        for var in var_list:
+            ts_files = sorted(input_location.glob(f"{case_name}*h0*.{var}.*nc"))
+            if ts_files:
+                print(var)
+                break
+            else:
+                logmsg = "get years for time series:"
+                logmsg = f"\tVar '{var}' not in dataset, skip to next to try and find climo years..."
+                self.debug_log(logmsg)
```
justin-richling, line 790: This is in case the first variable in the yaml file doesn't have a history file, we need to keep looking for the next available one.

justin-richling:
@brianpm I've added the changes we discussed and cleaned up the methods. If you wouldn't mind giving `adf_dataset.py` another quick glance, we can probably pass this to @nusbaume for a final review. Thanks.

lib/adf_dataset.py
```diff
@@ -245,14 +279,50 @@ def load_dataset(self, fils):
...
         if ds is None:
             warnings.warn(f"ERROR: Load failed for {variablename}")
             return None
         da = (ds[variablename]).squeeze()
-        if variablename in self.adf.variable_defaults:
-            vres = self.adf.variable_defaults[variablename]
-            da = da * vres.get("scale_factor",1) + vres.get("add_offset", 0)
-            da.attrs['units'] = vres.get("new_unit", da.attrs.get('units', 'none'))
+
+        da = da * kwargs["scale_factor"] + kwargs["add_offset"]
```
brianpm, line 291: I think we probably just need to make this a bit more robust with something like
`scale_factor = kwargs.get('scale_factor', 1)`
`add_offset = kwargs.get('add_offset', 0)`
Just in case it gets used without the kwargs being passed.

brianpm: Actually, maybe use `get_defaults` to do this?

justin-richling: That's what I was attempting to do with `get_defaults`, it should be setting those to the defaults before it becomes the `kwargs`

I have the offset set to 0 and scale factor to 1 initially, then the check `if variablename in res` should grab the values and if not it will default to 0/1 respectively with `vres.get`. 

And if the variable is not in `self.adf.variable_defaults`, the original values for `new_unit`, `add_offset` , and `scale_factor` will take on the original set values of none, 0, and 1.

justin-richling:
@nusbaume I think this is all good for your final review., let me know if you see anything that needs addressing.

nusbaume (changes requested):
Looks good, but do have some (hopefully easy) change requests.

lib/adf_dataset.py
```diff
@@ -203,32 +142,127 @@ def load_timeseries_dataset(self, fils):
...
+
+
+    # Climatology files
+    #------------------
+
+    # Test case(s)
+    def load_climo_da(self, case, variablename):
+        """Return DataArray from climo file"""
+        add_offset, scale_factor = self.get_defaults(case, variablename)
+        fils = self.get_climo_file(case, variablename)
```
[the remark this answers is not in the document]
justin-richling: The next line in `load_climo_da` calls ```self.load_da``` which should be catching this scenario and returning `None` and a warning message:

```python
    # Load DataArray
    def load_da(self, fils, variablename, **kwargs):
        """Return xarray DataArray from files(s) w/ optional scale factor, offset, and/or new units"""
        ds = self.load_dataset(fils)
        if ds is None:
            warnings.warn(f"ERROR: Load failed for {variablename}")
            return None
        da = (ds[variablename]).squeeze()

        da = da * kwargs["scale_factor"] + kwargs["add_offset"]
        if variablename in self.adf.variable_defaults:
            vres = self.adf.variable_defaults[variablename]
            da.attrs['units'] = vres.get("new_unit", da.attrs.get('units', 'none'))
        else:
            da.attrs['units'] = 'none'
        return da
```

Which will first check if there are any files using `self.load_dataset`

lib/adf_dataset.py
```diff
@@ -203,32 +142,127 @@ def load_timeseries_dataset(self, fils):
...
-        return self.load_da(fils, field)
+        return self.load_da(fils, field, add_offset=add_offset, scale_factor=scale_factor)
+
+
+    # Reference case (baseline/obs)
+    def get_ref_regrid_file(self, case, field):
+        """Return list of reference regridded files"""
+        if self.adf.compare_obs:
+            obs_loc = self.ref_var_loc.get(field, None)
+            fils = [str(obs_loc)]
```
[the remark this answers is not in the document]
justin-richling: Ah, good catch!

lib/adf_dataset.py
```diff
@@ -244,15 +278,50 @@ def load_dataset(self, fils):
...
         if variablename in self.adf.variable_defaults:
             vres = self.adf.variable_defaults[variablename]
-            da = da * vres.get("scale_factor",1) + vres.get("add_offset", 0)
             da.attrs['units'] = vres.get("new_unit", da.attrs.get('units', 'none'))
+        else:
+            da.attrs['units'] = 'none'
         return da
+
+    # Get vairable defaults, if applicable
+    def get_defaults(self, case, variablename):
```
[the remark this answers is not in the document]
justin-richling: I like that name for the function better, done. 

I forgot that `load_da` can be called outside of the normal ADF usage, so the kwargs might not be declared (as you pointed out), so I added:

```python
        scale_factor = kwargs.get('scale_factor', 1)
        add_offset = kwargs.get('add_offset', 0)
        da = da * scale_factor + add_offset
```

justin-richling:
@nusbaume I believe I addressed all the suggestions. If you wouldn't mind just giving it one last glance, I think it's on its way, thanks!

lib/adf_dataset.py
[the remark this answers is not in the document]
nusbaume: Good point!  I definitely missed that (and now think this code is fine as-is).

PR #324, opened 2024-08-22
Add capability for ADF to run MDTF with hourly data  

bitterbark:
Added a dict to translate the CESM output frequency strings (eg. hour_1) to MDTF file naming conventions (eg 1hr).


justin-richling (approved):
Looks good to me, thanks @bitterbark

PR #326, opened 2024-09-03
Derived variables structure update

justin-richling:
Move the calculation of derived variables out of `adf_diag.py` into an external script. This is a placeholder until a more robust variable derivations scheme is implemented.

This will help declutter `adf_diag.py` in the hopes of keeping non-essential methods separate. 

`adf_derive.py` has two main methods:  `check_derive` and `derive_variable`

* `check_derive` - called in variable loop; will check on each available variable
     For incoming variable, look for list of constituents if available as a list in variable defaults yaml file

     If the variable does not have the argument `derivable_from` or `derivable_from_cam_chem`,
     then it will be assumed not to be a derivable variable, just missing from history file

     If the variable does have the argument `derivable_from` or `derivable_from_cam_chem`,
     first check cam-chem, then regular cam and add all constituents to `diag_var_list` for time series generation. 

     NOTE: this will only modify `diag_var_list` for use in `adf_diag.py`, it will not modify the yaml variable list used in other scripts.

* `derive_variable` - called after variable loop and time series generation; loop over dictionary containing derived variables as keys and list of constituents as values.
     This will then be called in a loop of accepted derived variables and will be generated from constituent time series files.


lib/adf_derive.py
```diff
@@ -0,0 +1,274 @@
...
+        print(exit_msg)
+        self.debug_log(exit_msg)
+    # End if
+
+    return diag_var_list, constit_dict
+
+########
+
+def derive_variable(self, case_name, var, res=None, ts_dir=None,
+                         constit_list=None, overwrite=None):
```
nusbaume, line 148: I think all of these variables should be required except `overwrite`, as the code will fail as-is if those quantities aren't provided.  Also we should rename `self` to something like `adfobj`, at least until it becomes part of the offficial ADF class structure:

```suggestion
def derive_variable(adfobj, case_name, var, res, ts_dir,
                         constit_list, overwrite=False):
```

lib/adf_derive.py
```diff
@@ -0,0 +1,274 @@
...
+        else:
+            dmsg = f"derived time series for {case_name}:"
+            dmsg += f"\n\tneeded constituents for derivation of "
+            dmsg += f"{var}:\n\t\t- {constit_list}\n"
+            dmsg += f"\tNo constituent(s) found in history files"
+            self.debug_log(dmsg)
+        # End if
+    else:
+        # Open a new dataset with all the constituent files/variables
+        ds = self.data.load_dataset(constit_files)
```
nusbaume, line 194: Just a reminder to replace `self` here and below with the new object name (e.g. `adfobj`).

lib/adf_derive.py
```diff
@@ -0,0 +1,274 @@
...
+        derived_file = constit_files[0].replace(constit_list[0], var)
+
+        # Check if clobber is true for file
+        if Path(derived_file).is_file():
+            if overwrite:
+                Path(derived_file).unlink()
+            else:
+                msg = f"[{__name__}] Warning: '{var}' file was found "
+                msg += "and overwrite is False. Will use existing file."
+                print(msg)
```
nusbaume, line 214: Should we `return` here?  Otherwise it seems like we are doing calculations for no reason.

nusbaume (changes requested):
Nice cleanup!  Just have a few suggestions.

lib/adf_derive.py
justin-richling: Yeah, good catch!

brianpm (commented):
Once Jesse's comments are addressed, I think this looks good. Claude helped me with the review, here's what it says. 

Lines 92-127:
The error message can be written twice in certain scenarios:

Line 104: If CAM-CHEM branch finds an empty constit_list, it logs constit_errmsg
Line 111: Then the code might overwrite constit_list with vres["derivable_from"]
Lines 126-127: If the new constit_list is also empty, the same error is logged again
Plus, the isinstance(constit_list, list) check at line 126 is redundant since constit_list is initialized as a list at line 80 and should always be a list.

@nusbaume  suggestion is valid. Here's the fix:


    # Initialize tracking for error messages
    constit_errmsg_written = False
    
    try_cam_constits = True
    # Try finding info from variable defaults yaml file
    try:
        vres = res[var]
    except KeyError:
        print(exit_msg)
        self.debug_log(exit_msg)
        return diag_var_list, constit_dict

    # Check first if variable is potentially part of a CAM-CHEM run
    if "derivable_from_cam_chem" in vres:
        constit_list = vres["derivable_from_cam_chem"]

        if constit_list:
            if all(item in hist_file_ds.data_vars for item in constit_list):
                # Set check to look for regular CAM constituents in variable defaults
                try_cam_constits = False
                msg = f"derive time series for {case_name}:"
                msg += "\n\tLooks like this a CAM-CHEM run, "
                msg += f"checking constituents for '{var}'"
                self.debug_log(msg)
        else:
            # Only write error if we haven't already written it
            self.debug_log(constit_errmsg)
            constit_errmsg_written = True
        # End if
    # End if
    
    # If not CAM-CHEM, check regular CAM runs
    if try_cam_constits:
        if "derivable_from" in vres:
            constit_list = vres["derivable_from"]
        else:
            # Missing variable or missing derivable_from argument
            der_from_msg = f"derive time series for {case_name}:"
            der_from_msg += f"\n Can't create time series for {var}.\n\tEither "
            der_from_msg += "the variable is missing from CAM output or it is a "
            der_from_msg += "derived quantity and is missing the 'derivable_from' "
            der_from_msg += "config argument.\n\tPlease add variable to CAM run "
            der_from_msg += "or set appropriate argument in variable "
            der_from_msg += "defaults yaml file."
            self.debug_log(der_from_msg)
            constit_errmsg_written = True
        # End if
    # End if

    # Log if this variable can be derived but is missing list of constituents
    # (only if not already logged above)
    if not constit_list and not constit_errmsg_written:
        self.debug_log(constit_errmsg)
Key changes:
Add constit_errmsg_written flag at the top
Set flag to True whenever we log an error (lines 104→105, 121→122)
Remove the isinstance() check at line 126 and replace with the simpler if not constit_list and not constit_errmsg_written:
This eliminates redundant error logging and makes the logic clearer.

justin-richling:
Thanks @brianpm (and Claude) I was just looking through this PR, this will help!

lib/adf_derive.py
```diff
@@ -0,0 +1,274 @@
...
+        if constit_list:
+            if all(item in hist_file_ds.data_vars for item in constit_list):
+                # Set check to look for regular CAM constituents in variable defaults
+                try_cam_constits = False
+                msg = f"derive time series for {case_name}:"
+                msg += "\n\tLooks like this a CAM-CHEM run, "
+                msg += f"checking constituents for '{var}'"
+                self.debug_log(msg)
+        else:
+            self.debug_log(constit_errmsg)
```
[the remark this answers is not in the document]
justin-richling: I think the changes I brought in should help with the confusing logic, thanks @brianpm

lib/adf_diag.py
```diff
@@ -512,6 +522,8 @@ def call_ncrcat(cmd):
 
                 # Intitialize list for NCO commands
                 list_of_commands = []
+                list_of_ncattend_commands = []
```
nusbaume: I'm assuming this is supposed to be `ncatted` instead of `ncattend`?

lib/adf_diag.py
```diff
@@ -524,8 +536,24 @@ def call_ncrcat(cmd):
...
+                        + os.sep
+                        + ".".join([case_name, hist_str, var, time_string, "nc"])
+                    )
+
+                    # Check if clobber is true for file
+                    if Path(ts_outfil_str).is_file():
+                        if overwrite_ts[case_idx]:
+                            Path(ts_outfil_str).unlink()
+                        else:
+                            #msg = f"[{__name__}] Warning: '{var}' file was found "
```
nusbaume: Remove commented-out line?

lib/adf_diag.py
```diff
@@ -610,20 +629,57 @@ def call_ncrcat(cmd):
...
+                    ]
+
                     # Add to command list for use in multi-processing pool:
+                    # -----------------------------------------------------
+                    # generate time series files
                     list_of_commands.append(cmd)
+                    # Add global attributes: user, original hist file loc(s) and all filenames
+                    list_of_ncattend_commands.append(cmd_ncatted)
+                    # Remove the `history` attr that gets tacked on (for clean up)
+                    # NOTE: this may not be best practice, but it the history attr repeats
```
nusbaume: grammar fix:

`but it the` -> `but the`

nusbaume (approved):
Looks great now, just had a couple final (non-required) cleanup requests!

Issue #211, opened 2022-10-13
generated files need provenance information


bitterbark:
### New feature type

New infrastructure or infrastructure enhancement

### What is this new feature?

Each file that is created by the ADF during processing (eg timeseries, climo, regrid) should have meta data indicating
how it was made.

My suggestions are to put in the history attribute:
- made by user 
- ADF (version & link)
- which script/routine
- from which files (summarize if necessary, eg. 24 files (first $case.h0.0001-01.nc ... last $case.h0.0002-12.nc)

### Assistance required?

Yes, I will need some assistance before making a PR

### Extra info

I can work on this but if someone else wants to jump on it first, go for it. Also I welcome input regarding what is included and how.

PR #333, opened 2024-09-24
Add global attrs to all ADF generated files

justin-richling:
This will add attributes for:
- timeseries files: ADF user, history files location, and history file name(s)
- climo files: ADF user, climo years, and time series file name(s)
- regridded files: ADF user, climo years, and climo file name(s). The regridded files will have both the test and baseline climo years in the atributes


justin-richling:
Closes #211

PR #334, opened 2024-10-15
Global mean time series plotting script

brianpm:
Add a script to plot global mean time series (annual mean only right now).

Uses time series files directly. To do this, I added some functionality to `adf_dataset.py` to get time series data.

Adds a TimeSeries type of plot category to `adf_variable_defaults.yaml`. 

Modifies `plotting_functions.py` to try to keep units in tact while doing spatial and annual averages. 

Changes in `adf_info.py` is incidental (two print statements), and can be ignored. 
The change in `adf_web.py` is literally whitespace and can be ignored.

### DIFF
--- a/lib/adf_web.py
+++ b/lib/adf_web.py
@@ -684,7 +684,7 @@ class AdfWeb(AdfObs):
 
             #List of ADF default plot types
             avail_plot_types = res["default_ptypes"]
-            
+           
             #Check if current plot type is in ADF default.
             #If not, add it so the index.html file can include it
             for ptype in plot_types.keys():
--- a/scripts/plotting/global_mean_timeseries.py
+++ b/scripts/plotting/global_mean_timeseries.py
@@ -29,31 +29,38 @@ def global_mean_timeseries(adfobj):
     Include the CESM2 LENS result if it can be found.
     """
 
-    plot_loc = get_plot_loc(adfobj)
+    #Notify user that script has started:
+    print("\n  Generating global mean time series plots...")
 
+    # Gather ADF configurations
+    plot_loc = get_plot_loc(adfobj)
     plot_type = adfobj.read_config_var("diag_basic_info").get("plot_type", "png")
 
+    # Loop over variables
     for field in adfobj.diag_var_list:
         # reference time series (DataArray)
         ref_ts_da = adfobj.data.load_reference_timeseries_da(field)
+        
+        # Check to see if this field is available
         if ref_ts_da is None:
             print(
                 f"\t Variable named {field} provides Nonetype. Skipping this variable"
             )
-            continue
-        # check if this is a "2-d" varaible:
-        has_lat_ref, has_lev_ref = pf.zm_validate_dims(ref_ts_da)
-        if has_lev_ref:
-            print(
-                f"Variable named {field} has a lev dimension, which does not work with this script."
-            )
-            continue
+        else:
+            # reference time series global average
+            ref_ts_da_ga = pf.spatial_average(ref_ts_da, weights=None, spatial_dims=None)
 
-        # reference time series global average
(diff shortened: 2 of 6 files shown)


scripts/plotting/global_mean_timeseries.py
```diff
@@ -0,0 +1,199 @@
...
+
+    plot_loc = get_plot_loc(adfobj)
+
+    plot_type = adfobj.read_config_var("diag_basic_info").get("plot_type", "png")
+
+    for field in adfobj.diag_var_list:
+        # reference time series (DataArray)
+        ref_ts_da = adfobj.data.load_reference_timeseries_da(field)
+        # check if this is a "2-d" varaible:
+        has_lat_ref, has_lev_ref = pf.zm_validate_dims(ref_ts_da)
```
adagj, line 40: If the field doesn't exist (for some reason), the adfobj.data.load_reference_timeseries_da(field) will return a Nonetype, and 
  has_lat_ref, has_lev_ref = pf.zm_validate_dims(ref_ts_da)
fails, and ADF crashes. 

I added 
ref_ts_da = adfobj.data.load_reference_timeseries_da(field)
        if ref_ts_da is None:
            print(
                f"\t Variable named {field} provides Nonetype. Skipping this variable"
            )
            continue
        # check if this is a "2-d" varaible:
        has_lat_ref, has_lev_ref = pf.zm_validate_dims(ref_ts_da)
        
        which seems to work

lib/adf_web.py
```diff
@@ -684,7 +684,7 @@ def jinja_list(seas_list):
 
             #List of ADF default plot types
             avail_plot_types = res["default_ptypes"]
-            
+           
             #Check if current plot type is in ADF default.
             #If not, add it so the index.html file can include it
             for ptype in plot_types.keys():
```
adagj, line 690: There was a problem causing ADF to crash in the for loop starting on line 687 in adf_web.py

I needed to change 

```
 for ptype in plot_types.keys():
                if ptype not in avail_plot_types:
                    avail_plot_types.append(plot_types)
                    
```
to


```
 for ptype in plot_types.keys():
                if ptype not in avail_plot_types:
                    avail_plot_types.append(ptype)
                    
```

justin-richling: @adagj Thanks for catching this! Although this should've caught the `TimeSeries` plot type since it is included in `lib/adf_variable_defaults.yaml`. I'm curious why this crashed for you, would you be able to provide any further details when this crashed? Thanks!

scripts/plotting/global_mean_timeseries.py
justin-richling: @adagj You're right, thanks for finding this!  I can confirm this fix works

Additionally, if the ADF is comparing against obs, it will also trigger the `Nonetype`.

@brianpm We can just add this check and not have the time series plot made if comparing against obs, or I have coded up a quick fix to allow the time series plots to be created in the model vs obs case, but will just plot the test time series. I'm happy to share if this desired.

justin-richling: Screenshot 2024-11-18 at 3 13 33 PM (3)

brianpm: @justin-richling -- I think your solution to keep making the plot (just skip the reference) is the desired behavior. Can you incorporate your fix and merge it in?

justin-richling: @brianpm Will do!

justin-richling (approved):

justin-richling:
@brianpm This all looks good to me. now. I've tested with obs and tested with with fake variables and it works now as intended.

@adagj Thanks again for the review and changes, we appreciate your time in helping!

PR #337, opened 2024-10-23
adding support for variables on the ilev vertical grid.

bstephens82:
The one-line change to adf_diag.py ensures that PS, hyam, hybm, hyai, and hybi will be added to the time series and climatology files for variables that have an `ilev` dimension, and the changes to regrid_and_vert_interp.py sets `has_lev=True` if `ilev` is present in the regrid file, and calls the vertical interpolation function for those files.


brianpm:
Just to chime in that this looks good to me, and seems like the most straightforward way to handle ilev variables.

justin-richling:
@bstephens82 This looks good, thanks for the enhancement!

PR #340, opened 2024-10-31
fix MDTF implementation

bitterbark:
Improves the functionality of MDTF
- ADF now waits for the MDTF to be done before exiting
- The default settings for MDTF to save_nc = false and make_tar = false, so that the output can just be tarred with ADF output
- The file permissions are fixed so that figures are viewable from the webpag


justin-richling:
Thanks for the update @bitterbark

PR #349, opened 2024-12-12
Update glob for variables with underscore

justin-richling:
There was an issue with the wildcard in globbing regridded files that would cause issues if two variable names were similar but one had an underscore, ie vpwp_mfl and vpwp.


justin-richling:
@brianpm I think this fix will address the problem of variables with underscores. It worked on the test cases Ben sent me. Can you just take a quick look and see if I might be missing any unintended consequences?

brianpm:
Yeah, that does seem good! I sent you an alternative using a regex, but this seems nice and simple. Maybe we should have Ben test it?

PR #352, opened 2025-01-08
Update global_mean_timeseries.py

justin-richling:
Currently the time series plot is crashing when trying to check dimensions of the zm data. This is because the data was averaged before checking dims with `pf.zm_validate_dims`, which will fail because the 'lat' dim will no longer be present after spatial average. This PR will change order of checking zm dims first then it will average the data if it passes the dim check.

Also update LENS2 file grab since the files in the directory are not named exactly the same so add wildcard for searching.

### DIFF
--- a/lib/adf_dataset.py
+++ b/lib/adf_dataset.py
@@ -149,6 +149,9 @@ class AdfData:
         """
         add_offset, scale_factor = self.get_value_converters(case, variablename)
         fils = self.get_timeseries_file(case, variablename)
+        if not fils:
+            warnings.warn(f"\t    WARNING: Did not find case time series file(s), variable: {variablename}")
+            return None
         return self.load_da(fils, variablename, add_offset=add_offset, scale_factor=scale_factor)
     
     def load_reference_timeseries_da(self, field):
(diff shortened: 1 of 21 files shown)


brianpm:
I think this line:
`has_lat_ref, has_lev_ref = pf.zm_validate_dims(ref_ts_da)`
could still be problematic if zm_validate_dims returns None. 

We can either fix zm_validate_dims (return (None, None), for example), or we might just change this to:
```python
valdims = pf.zm_validate_dims(ref_ts_da)
if valdims is not None:
  has_lat_ref, has_lev_ref = valdims
else:
  has_lat_ref, has_lev_ref = False, False # I *think* this is what we'd expect here ... double check
```

justin-richling:
I think your'e right, if there is a scenario where there is not a `lat` dim, this would be a problem. I guess either change is sufficient because we want to shut it down if it's missing `lat` dim. @nusbaume any thoughts on which would be preferable? I can't seem to see an advantage of one over the other.

Either way, we should also probably put a check for when `has_lat` is `None` or `False` with a `continue` call too so that it doesn't try to get to `ref_ts_da_ga = pf.spatial_average(ref_ts_da, weights=None, spatial_dims=None)`

nusbaume:
@justin-richling If other scripts that are calling `zm_validate_dims` are checking for just a single `None` then I would vote to implement @brianpm's option.  However, if none of them are doing this check then I would probably vote to modify `zm_validate_dims` itself to just return `False` for both variables in this case, as that should keep the interface the same everywhere while still protecting us from the situation where the latitude dimension or input data is invalid.

Anyways, that's my two cents.  Hope that helps!

justin-richling (commented):
All test cases work as intended

PR #353, opened 2025-01-10
Update python version list for unit tests.

nusbaume:
This PR removes end-of-life versions of python (3.7 and 3.8) from the automated unit testing workflow, and adds two newer language versions that are currently missing from the test list (3.12 and 3.13).


justin-richling:
Sounds good, thanks for addressing this!

justin-richling (approved):

PR #354, opened 2025-01-23
Land Diagnostics using ADF

wwieder:
This is a draft PR, but hopefully offers a place to start working on using the ADF framework to make land model diagnistics


wwieder:
@justin-richling I realize this PR inherited a bunch of your older work that may not be necessary.

I also was wondering if you could help me with aspects of this PR that are still a little clunky. I've tried marking these with TODOs in the modified files.

Finally, I'm pretty lost by what the plotting functions are doing, but I have a mock up of lat-lon plots using uxarray in plot_uxarray_test.ipynb that needs to be migrated into the new global_unstructured_latlon_map.py code. 

I'm planning on coming the the SEWG hackathon if that's a convenient time to work on this?  I also appreciate you may be over subscribed on that day.

wwieder:
OK, as a baseline, it seems like this is ready for review.  I do worry that some of the changes we made may break ADF functionality on the CAM side, so maybe it's worth testing / trying this out with this concern in mind?

brianpm:
Just a note that I've been experimenting this week with sending CAM-SE native files through ADF. At this point I have some prototype code that does a *very* similar xESMF regridding step as is included here. So it there could be some potential iteration on that to generalize the regridding steps so both CAM and CLM can use the same functions.

wwieder:
@brianpm that would be excellent.  There are some particular things that we have to do to properly weight grids by their landfraction for CLM output, but hopefully much of the basic regridding functionality can be more generalized.

PR #357, opened 2025-02-14
Address python generated warnings

justin-richling:
This PR will clean up some warnings that python prints to console (not ADF warnings). 

amwg_table: - Add check if case has only one year of data. If so, ignore statistics and just add means to table.
* python warning was justified, but now is good

plotting_functions: - Change global average function from `global_average` to  `spatial_average` to suppress the warnings that come from the spurious underflow error bug in numpy's mean calculation. 
* python warning is due to bug; can't fix. This is a warning that shows up if there is a very small number involved in the calculation. Verified that was not the case in the ADF run. 
* Instead xarray will be used in `spatial_average()` method in`lib/plotting_functions.py`
* 10+ year old issue https://github.com/numpy/numpy/issues/4895

Thanks

### DIFF
--- a/lib/plotting_functions.py
+++ b/lib/plotting_functions.py
@@ -1218,7 +1218,7 @@ def plot_map_and_save(wks, case_nickname, base_nickname,
 
     # get statistics (from non-wrapped)
     fields = (mdlfld, obsfld, diffld, pctld)
-    area_avg = [global_average(x, wgt) for x in fields]
+    [spatial_average(x, weights=wgt, spatial_dims=None) for x in fields]
 
     d_rmse = wgt_rmse(mdlfld, obsfld, wgt)  # correct weighted RMSE for (lat,lon) fields.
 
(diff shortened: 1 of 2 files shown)


lib/plotting_functions.py
```diff
@@ -1218,7 +1218,7 @@ def plot_map_and_save(wks, case_nickname, base_nickname,
 
     # get statistics (from non-wrapped)
     fields = (mdlfld, obsfld, diffld, pctld)
-    area_avg = [global_average(x, wgt) for x in fields]
+    [spatial_average(x, weights=wgt, spatial_dims=None) for x in fields]
```
mnlevy1981, line 1221: Does this still need to be saved as `area_avg`?

justin-richling: Nice catch, this was declared in my branch. I wonder what happened. Anyways, just updated it thanks!

PR #358, opened 2025-03-07
Better "pre-flight" checking and refactoring of polar maps

brianpm:
This PR contains a very tiny change in the global maps script. It is a bug fix to reverse the logic when checking whether all the plots for a given variable have been made. This will speed up ADF runs that don't need to re-make those plots.

It does the same for the polar plots, but with a much more substantial refactoring of the script. I think that it produces all the same plots, but the code is more efficient and easier to read. Essentially it has a more elaborate "pre-flight" part of the code, where it figures out all the plots that will be made and keeps all the information about them in a dictionary, and then iterates through a list of these dictionaries to do the calculations. This reduces the complexity of the code a lot because it removes nested loops for all the variations. It also updates the polar plots script to use the adf_dataset module, which also reduces code. 

If we confirm this works and we like the approach, I think we should open an issue to refactor the other standard plotting scripts to use this structure.

### DIFF
--- a/lib/plotting_functions.py
+++ b/lib/plotting_functions.py
@@ -2007,7 +2007,7 @@ def prep_contour_plot(adata, bdata, diffdata, pctdata, **kwargs):
 
 def plot_zonal_mean_and_save(wks, case_nickname, base_nickname,
                              case_climo_yrs, baseline_climo_yrs,
-                             adata, bdata, has_lev, log_p, obs=False, **kwargs):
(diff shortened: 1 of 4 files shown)


brianpm:
@justin-richling -- Could you test this and confirm that it is generating the same plots as the current ADF script? I'm kind of excited to reduce the size of the script by this much, I hope I'm not missing anything.

PR #359, opened 2025-03-12
ADF MOPITT CO plotting script addition 

shawnusaf:
Adds the capability to compare test case Carbon Monoxide (CO) to MOPITT satellite observed CO.  Plots include monthly and annual mean comparisons.


justin-richling (approved):
@shawnusaf I apologize for the super late review, thanks for the work on these diagnostics

PR #360, opened 2025-03-28
add requirement for xesmf

YanchunHe:
update the environment requirement file for `xesmf` in relation to commit 17524b6.


justin-richling:
Thanks for for the addition @YanchunHe

PR #374, opened 2025-04-07
Add ENSO analysis and plot scripts

megandevlan:
Create an analysis script that computes a variety of ENSO metrics (western extent, duration, etc.) for a new case, adding that to a pre-created netCDF file containing the same variables across previous simulations. Those are then plotted and placed in a new "ENSO" folder on the ADF website.


brianpm (approved):
Looks good to me.

Issue #368, opened 2025-04-06
Enable regional subsetting


wwieder:
### What is this new feature?

Write or adapt python function to make regional timeseries or climo plots.
- [x] Create dictionary of regions to plot
- [x] Need workflow for unstructured and regular grids (uxarry and xarray approaches)
   - build on / adapt existing `region_multicase` examples in ADF
   - see also [this notebook, cell 22](https://github.com/wwieder/ADF/blob/clm-test/scripts/regridding/regrid_conservative.ipynb) for a uxarray example.  
- [x] Develop web interface for displaying results (again maybe build on region_multicase?)
  - For example see set_3 and set_6 from [current land diags](https://webext.cgd.ucar.edu/BLT1850/b.e30_alpha06b.B1850C_LTso.ne30_t232_wgx3.143/lnd/b.e30_alpha06b.B1850C_LTso.ne30_t232_wgx3.143_7_16-b.e30_alpha06b.B1850C_LTso.ne30_t232_wgx3.139_21_40/setsIndex.html)

wwieder:
Another helpful example of regional hydrology from @olyson [is on these slides](https://docs.google.com/presentation/d/1L2HsGVlRD1Rt-6iuQGPivwERuM9w0P2a/edit?usp=sharing&ouid=110501992574177180138&rtpof=true&sd=true)

olyson:
The current NCL-based diagnostics package gets the regions information from this file:

/glade/campaign/cgd/tss/people/oleson/FROM_LMWG/diag/lnd_diag4.2/code/resources/region_definitions.nc

ncdump -h region_definitions.nc

...
        int region(region) ;
        float BOX_S(region) ;
                BOX_S:long_name = "Southern Boundary" ;
                BOX_S:_FillValue = -9999.f ;
        float BOX_N(region) ;
                BOX_N:long_name = "Northern Boundary" ;
                BOX_N:_FillValue = -9999.f ;
        float BOX_W(region) ;
                BOX_W:long_name = "Western Boundary" ;
                BOX_W:_FillValue = -9999.f ;
        float BOX_E(region) ;
                BOX_E:long_name = "Eastern Boundary" ;
                BOX_E:_FillValue = -9999.f ;
        char PS_ID(region, ncl1) ;
                PS_ID:long_name = "Postscript File Id" ;
                PS_ID:_FillValue = "" ;
        char PTITSTR(region, ncl2) ;
                PTITSTR:long_name = "Region Title" ;
[...]

PR #386, opened 2025-04-17
Add regional plot climatology capability

megandevlan:
As described in https://github.com/NCAR/ADF/issues/368#issuecomment-2797830149, enables the ability to look at regional climatologies. 

Admittedly there's still more to do, but this is a start! A few notes: 

- On the website, I'd like the regions to show multi-row, rather than extending the website so far horizontally; not sure how to do that! 
- I've commented out some regions in config_clm_unstructured_plots.yml; they break the script and I haven't tracked down why. But since we want to move to yml file for region definition anyway, maybe we wait for that?


brianpm:
Hi @megandevlan -- I haven't looked through your code yet. I'm not sure if this will be helpful at this point, but I had previously added similar capability, I think. Take a look at: scripts/plotting/regional_map_multicase.py

megandevlan:
Thanks Brian! I took a quick look initially and opted to diverge slightly, but it's a good point that there might be something more efficient. I can dig into this a bit more (though I might be a bit slow :( )

Issue #376, opened 2025-04-10
Land regional subsetting: Convert netCDF dictionary to YAML


samsrabin:
The existing land diagnostics package uses this netCDF file to define regions:
```
/glade/campaign/cgd/tss/people/oleson/FROM_LMWG/diag/lnd_diag4.2/code/resources/region_definitions.nc
```

Here's the structure:
```
variables:
	int region(region) ;
	float BOX_S(region) ;
		BOX_S:long_name = "Southern Boundary" ;
		BOX_S:_FillValue = -9999.f ;
	float BOX_N(region) ;
		BOX_N:long_name = "Northern Boundary" ;
		BOX_N:_FillValue = -9999.f ;
	float BOX_W(region) ;
		BOX_W:long_name = "Western Boundary" ;
		BOX_W:_FillValue = -9999.f ;
	float BOX_E(region) ;
		BOX_E:long_name = "Eastern Boundary" ;
		BOX_E:_FillValue = -9999.f ;
	char PS_ID(region, ncl1) ;
		PS_ID:long_name = "Postscript File Id" ;
		PS_ID:_FillValue = "" ;
	char PTITSTR(region, ncl2) ;
		PTITSTR:long_name = "Region Title" ;
		PTITSTR:_FillValue = "" ;
	char BOXSTR(region, ncl3) ;
		BOXSTR:long_name = "Region Lat/Lon Title" ;
		BOXSTR:_FillValue = "" ;
```

For git-trackability and ease of working with the region definitions, this should be converted to a YAML dictionary in the ADF repo.

wwieder:
done with #387

PR #387, opened 2025-04-17
yaml and python options to define regions

wwieder:
I pulled the regions and lat-lon boundaries from the .nc file used by the current land diagnostics package to address [#376](https://github.com/NCAR/ADF/issues/376)  This was done: 
- manually (maybe introducing errors) and 
- naively (perhaps without the right formatting)

@megandevlan & @samsrabin please see if this yaml file can be read in and used as expected in the regional subsetting you're working on.  

**Note:** there are regions that span 0 longitude.  Not sure what the best way to handle this is?
I've also provided lats -180 to 180, as I think this is what uxarray needs, but would have to be modified for xarray datasets that are 0-360. 

Finally, I tried using copilot to see if the yaml file was formatted correctly to create a python dictionary and it just created the python dictionary for me!  I've included that here too, but both files don't need to be merged.  Dealers choice on which you prefer


samsrabin:
I ended up getting distracted from regions and working on #382 instead, so I'll let @megandevlan handle this. I vote for the actual YAML file though.

megandevlan:
I vote for the actual yml file as well, though I don't know that I have the bandwidth to test things out today. I can get to this early next week though!

PR #388, opened 2025-04-21
Separate time bounds for time series and climo generation

nmizukami:
**Descriptions:**

This PR is to enable configuring start year and end year for climatology computation.

New yaml variables are added: `climo_start_year`  and `climo_end_year`


nmizukami:
TO-DO: 
- Need to change a text for climatology years in plots.

PR #404, opened 2025-08-01
Add CAMS SO4 aerosol burden reanalysis to variable defaults

brianpm:
Adding CAMS reanalysis sulfate aerosol burden to adf_variable_defaults.yaml.


brianpm:
I added the CAMS dust burden, too. File put into standard ADF location, and this PR just changes variable defaults.

brianpm:
I have also added sea salt burden.

PR #411, opened 2025-08-24
WIP: latest updates for LDF

wwieder:
Updating to a newer version of uxarry seems to cause plotting issues...

### DIFF
--- a/lib/adf_diag.py
+++ b/lib/adf_diag.py
@@ -744,6 +744,10 @@ class AdfDiag(AdfWeb):
                         # End if cam
                     # End if has_lev
 
+                    # Optional, add additional time varying variables to clm2.h1 files
+                    if ("clm" in hist_str) and ("h1" in hist_str):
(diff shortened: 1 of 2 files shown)


TeaganKing:
It's on the [agenda](https://docs.google.com/presentation/d/1kF1qoVEkVvLggz1MSgnUiJ_AYm405oEvGiMh6qq6orA/edit?slide=id.g35fc035607e_0_43#slide=id.g35fc035607e_0_43)!

wwieder:
> Somewhat relevant to this issue ticket, can we update the `LDF/lib/scripts/plotting/regional_climatology.py` file line 87 to not use a hard-coded `ymlFilename`? This causes some issues when running from a different location than the LDF directory-- eg when running through CUPID. I'm thinking something in the config file similar to the `defaults_file` would probably work.

I also noticed that LDF has some infrastructure for reading in default.yaml files around line 112 of the same regional_climatology file that could be helpful for reading in the location of the `regions_lnd.yaml` file in question here.

PR #415, opened 2025-09-03
Updates to Chemistry tables

Behrooz-Roozitalab:
* the html page was not created, it is fixed.
            * 4 digits are included for years, to use it for historical periods.
            * added "hm" as a case to enable using annual averaged files in addition to monthly files.
            * if config arg "trop_val" is pres500 : This version uses 500hPa as the tropopause threshold.
            * if config arg "trop_val" is tropopause (realistic case) : New method of defining troposphere, use  ozone (150ppb) or Trop_P. If not found, calculate total column
            * Added DMS to gases list - reported as DMS not S
            * Automatic addition of gaseous compounds even when not defined in the default list, 
            *   based on Carbon MW (12). It still needs ADF modification to read a list from yaml file.

### DIFF
--- a/lib/adf_web.py
+++ b/lib/adf_web.py
@@ -117,7 +117,6 @@ class AdfWeb(AdfObs):
 
         #Extract needed variables from yaml file:
         case_names = self.get_cam_info('cam_case_name', required=True)
-        print(case_names)
         #Also extract baseline case (if applicable), and append to case_names list:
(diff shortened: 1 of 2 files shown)


Behrooz-Roozitalab:
@justin-richling Thanks for the comments. I resolved the edits I guess. Please check and let me know if something is missing!

Issue #407, opened 2025-08-07
Add git branch print to ADF


justin-richling:
### New feature type

New infrastructure or infrastructure enhancement

### What is this new feature?

It seems useful to document which branch the ADF is run from. There are a couple ways to do this, either print it at the start of the diagnostics and/or add it to the log file.

### Assistance required?

No, I will make a PR when the feature is ready

PR #416, opened 2025-09-09
Update ADF to save Git, Conda, and config yaml info and display to webpage (if applicable)

justin-richling:
Address #407 

Add infrastructure to gather and save/display information related to ADF runs:

Grab and store:
* Current config yaml file name and path and all info from config file to dictionary
* Conda environment
* Git info: branch, commit, remote url, repo name, and if has unstaged changes

If website generation is requested, this information will appear as a new webpage in the table of contents at the top of the webpage called `Case Setup Info` and will generate a markdown file that will be added to the `website` directory. If website generation is not requested, then the user will need to run the ADF with the `--debug` flag. This info will be written to the log file that gets stored in the user's ADF root directory.

### DIFF
--- a/lib/adf_web.py
+++ b/lib/adf_web.py
@@ -803,12 +803,10 @@ class AdfWeb(AdfObs):
             #Also check if index page exists for this case:
             index_html_file = \
                 self.__case_web_paths[web_data.case]['website_dir'] / "index.html"
-            print("index_html_file",index_html_file)
 
             # Create run info web page
             run_info_md_file = \
                 self.__case_web_paths[web_data.case]['website_dir'] / self.run_info
-            print("run_info_md_file",run_info_md_file)
 
             # Read the markdown file
             with open(run_info_md_file, "r", encoding="utf-8") as mdfile:
@@ -816,7 +814,7 @@ class AdfWeb(AdfObs):
 
             # Convert markdown to HTML
             run_info_html = markdown.markdown(md_text)
-            index_title = "AMP Diagnostics Prototype"
+            index_title = "CAM Diagnostics"
             run_info_html_file = self.__case_web_paths[web_data.case]['website_dir'] / "run_info.html"
             run_info_tmpl = jinenv.get_template('template_run_info.html')
             run_info_rndr = run_info_tmpl.render(title=index_title,


lib/adf_base.py
```diff
@@ -49,22 +49,22 @@ def __init__(self, debug = False):
             raise TypeError("'debug' must be a boolean type (True or False)")
 
         self.__debug_fname = ''
+        # Get the current date and time
+        current_timestamp = datetime.now()
+        # Format the datetime object to a string without microseconds
+        dt_str = current_timestamp.strftime('%Y-%m-%d %H:%M:%S')
+        ext = f'{str(dt_str).replace(" ","-")}'
+        debug_fname = f"ADF_debug_{ext}.log"
+        self.__debug_fname = debug_fname
```
brianpm, line 58: seems like this could be condensed to
`self.__debug_fname = f"ADF_debug_{datetime.now().strftime('%Y-%m-%d-%H:%M:%S')}.log"`

lib/adf_info.py
```diff
@@ -614,9 +611,85 @@ def __init__(self, config_file, debug=False):
         #End if
         #Print number of processors being used to debug log (if requested):
         self.debug_log(f"ADF is running with {self.__num_procs} processors.")
-        # -----------------------------------------
 
+        active_env = self.get_active_conda_environment()
```
nusbaume, line 615: Hi @justin-richling, as discussed in the hackathon, I might recommend the following changes:

1.  It might be good to collect all of the git/conda/config information in an internal ADF object or dictionary as early as possible in the ADF run, maybe even in the `AdfConfig` class?

2.  I might then move all of this code into two functions in `AdfWeb`.  One function can be for when `create_html` is True, which uses this new config info obj/dict to create the new markdown file, and another for when `create_html` is False but the debug log is active, in which case it just writes everything to the logger.

Of course, please let me know if you find that this doesn't actually work, or causes more headaches then benefits.  Thanks!

nusbaume (commented):
Thanks for bringing this feature in @justin-richling!  However, as discussed in the hackathon I wonder if it would help to rearrange some of this code so that it fits better within the ADF's infrastructure.

lib/adf_info.py
justin-richling: This worked perfectly, thanks!

justin-richling:
Thanks for the suggestions @brianpm and @nusbaume, these helped out a lot. I have moved all the logic out of `adf_info.py` to `adf_config.py` and `adf_web.py`. Additionally, I created two functions in `adf_web.py` to create the log output and html page for the run info, if either or both of `create_html` argument is set in the config yaml file and `--debug` is called at command line.

This should be good to review again.

lib/adf_web.py
```diff
@@ -695,6 +697,32 @@ def jinja_enumerate(arg):
             #Also check if index page exists for this case:
             index_html_file = \
                 self.__case_web_paths[web_data.case]['website_dir'] / "index.html"
+            print("index_html_file",index_html_file)
+
+            # Create run info web page
+            run_info_md_file = \
+                self.__case_web_paths[web_data.case]['website_dir'] / self.run_info
+            print("run_info_md_file",run_info_md_file)
```
nusbaume: Are these just debugging `print` statements?  If so then I might vote to remove them.

lib/adf_web.py
```diff
@@ -695,6 +697,32 @@ def jinja_enumerate(arg):
...
+
+            # Read the markdown file
+            with open(run_info_md_file, "r", encoding="utf-8") as mdfile:
+                md_text = mdfile.read()
+
+            # Convert markdown to HTML
+            run_info_html = markdown.markdown(md_text)
+            index_title = "AMP Diagnostics Prototype"
```
nusbaume: Not directly relevant for this PR, but at what point should we drop the word `Prototype` from the website title?  It feels like the ADF is used in production-like settings now.

PR #251, opened 2023-07-25
Cloud regime analysis

Isaaciwd:
Add a cloud regime analysis script to the ADF.

PR #417, opened 2025-09-15
Cloud regime analysis

brianpm:
This PR is supposed to replace #251. 

It is a substantial refactoring of Isaac's original script, but accomplishes the same thing: plots of the cloud regimes and maps of frequency of occurrence of cloud regimes. This works with ISCCP, MODIS, and MISR cloud regimes. There are a number of options available, and they should be set using the argument dictionary approach in the configuration YAML file.

I've done testing for ISCCP histograms, but haven't repeated for MODIS and MISR so far. 

There are quite a few helper functions included. I have tried to organize them into the categories:
- local functions
  + `_validate_user_inputs`
  + `_get_ref_cluster_labels`
  + `_preprocess_data`
  + `compute_cluster_labels`
  + `precomputed_clusters`
  + `_compute_distances_pot`
  + `_compute_distances_wasserstein`
  + `_calculate_rfo`
  + `load_reference_data`
  + `load_cluster_centers`
- plotting functions
  + `_plot_map`
  + `_configure_map_axes`
  + `_add_colorbar` 
  + `_add_colorbar2`
  + `plot_rfo_maps`
  + `plot_hists_baseline`
  + `plot_hists_obs`
  + `_prepare_plot_data`
  + `_plot_cloud_regimes`
  + `_plot_observation_column`
  + `_plot_baseline_column`
  + `_plot_test_case_column`
  + `_create_colormap`
  + `_configure_axes`
  + `_configure_modis_axes`
  + `_configure_misr_axes`
  + `_configure_isccp_axes`
  + `_add_figure_labels`
  + `_save_figure`
- Data handling
  + `spatial_subset`
  + `temporal_subset`
  + `select_valid_tau_height`
  + `_calculate_weighted_mean_xr`
- Land/Ocean masking
  + `apply_land_ocean_mask`
  + `_make_mask_broadcastable`
  + `create_land_mask`
  + `_find_land_points`
- Regridding
  + `make_se_regridder`
  + `regrid_se_data_bilinear`

This can probably be simplified more, but I thought it was worth getting it into ADF now, and modifying it incrementally as possible. In particular, the land/ocean masking and the regridding should probably be removed for more generalized versions of those functions that can be used across ADF scripts.


brianpm:
Using Claude Code to bring this PR up to current main branch. Claude found several issues to fix, and now more of the code has been tested (model-vs-model) (MISR hasn't been validated still, but I'll probably try to do that with new runs soon.) Here's Claude's comment on the audit and fixes:

Going in, only one code path had ever executed: obs comparison with pre-computed ISCCP labels, which returns early before reaching most of the script. Everything else raised on first call. Details below for reviewers.

### Crashes fixed

| Where | Problem |
|---|---|
| `_get_ref_cluster_labels` | `ot_library` undefined — it is a parameter of `cloud_regime_analysis` and was never passed down. `NameError` on every baseline run, and on any obs run without pre-computed labels. |
| `_plot_baseline_column` | `cluster_labels_b` undefined (flagged in-code by a `# Global variable` comment). Baseline mode always failed here. |
| `_preprocess_data` | `ALL_VARS[field_name]` `KeyError` — for observations `field_name` is the obs variable (`n_pctaudist`), not a key in `ALL_VARS`. |
| plotting columns | The obs column transposed its histogram for `pcolormesh`; the test/baseline columns did not. |
| baseline plotting | Averaged the **raw** reference data while the labels came from **preprocessed** data, so the two disagreed on time range and binning. `_get_ref_cluster_labels` now returns both. |

On the transpose: this does **not** affect native CAM output, which is `(time, cosp_prs, cosp_tau, ncol)` — height before tau — so existing ISCCP figures were fine. It raises for the opposite dimension order and would silently plot transposed for a square histogram. `_calculate_weighted_mean_xr` now returns `(ht, tau)` explicitly rather than depending on input ordering.

### Observation metadata was wrong

`obs_ht_var`/`obs_tau_var` were swapped for all three products. The obs files label their own axes, and the corrected assignment is what makes the valid bin counts match the supplied cluster centers:

```
ISCCP  levtau "Optical thickness levels" (6) x levpc "Pressure levels" (7)  = 42
MISR   tau "cloud optical depth" (7 valid) x cth "altitude" (15 valid)      = 105
MODIS  COT optical thickness (7) x PRES pressure (7)                        = 49
```

These never ran before because the `KeyError` above fired first. Also skipped the obs-comparison bin trimming for observational input — trimming exists to bring *model* histograms onto the centers' binning, and the obs files already match.

### MODIS units

`MODIS_CLD_HISTO` is a cloud fraction (profile totals ~0.5) while CAM's `CLMODIS` and the MODIS centers are percent (~60). Euclidean distance is not scale invariant, so every profile collapsed into a single cluster while looking like a valid result.

Fixed through the existing ADF mechanism: `obs_scale_factor: 100` on `CLMODIS`. CRA loads the whole obs `Dataset` via `load_dataset` (it needs the stored labels alongside the histogram), which bypasses the scaling `load_da` would apply, so it now applies it explicitly. Also added a check in `compute_cluster_labels` that warns when histogram and center magnitudes differ by more than 10x, so this class of problem can't pass silently again. ISCCP and MISR obs are already percent and untouched.

### Unrelated regression reverted

The branch's `adf_variable_defaults.yaml` had `'$\mu$g/m3'` → `'$\\mu$g/m3'` for BC, POM, SO4, SOA, DUST, SeaSalt. In single-quoted YAML the backslash is literal, so mathtext gets two and raises `ValueError` at draw time. Reverted to `main`'s form. This would have broken aerosol plots well outside this diagnostic.

### Portability and cleanup

- Hardcoded `/glade/work/brianpm/mapping_ne30pg3_to_fv09_esmfbilin.nc` replaced with a documented `regrid_weights_file` argument; unstructured cases are skipped with a clear message when it is absent.
- `n_jobs = 36` replaced with the `num_procs` value the script already loaded but never used for the POT path.
- `emd_method` now defaults to `"exact"`, matching the docstring. It had defaulted to sinkhorn, which the same docstring calls experimental and not recommended.
- RFO map counted NaN points as "not in this cluster" while the scalar total excluded them, so the two disagreed whenever land/ocean masking was on.
- Year ranges are per-case rather than case `[0]` applied to every case.
- `create_land_mask` built the land-point list, discarded it, then re-derived the mask by O(n_points x n_land) coordinate matching under numba; it now records the result directly. `_find_land_points` and the numba dependency for masking are gone.
- Removed dead `plot_data['weights']` (and its expensive stack), unused imports/locals, a leftover "identical to your original" comment, and a comment describing panel removal as setting ticks. In baseline mode the centers column no longer reports an RFO that just duplicated the baseline column's.

### Verification

- **Model-vs-model ADF run**: CESM2 amip vs CESM2 amip-p4K, CFmon `clisccp` (which is `FISCCP1_COSP` renamed), 2000–2001. Completes and produces all 8 figures. Physically sensible: CR4 sits in the storm tracks and is absent over subtropical subsidence, expanding under +4K (42.7% → 47.1%), while low-cloud CR7 drops 2.1% → 0.4%. This is the path that previously died on two `NameError`s.
- **ISCCP obs clustering** reproduces the file's stored `euclidean_cluster_labels` **exactly** (100% agreement).
- **MODIS obs clustering**, with the scale factor, likewise reproduces its stored labels.
- Static: clean compile and no undefined names; `pytest lib/test/unit_tests` passes; all CI checks green.

### Not covered — worth knowing before merge

- **MISR has never been run end-to-end.** The mapping and bin counts check out on paper (105 = 7 x 15) but no run has exercised it. I could not find a `CLD_MISR` source: the CMIP6 CFmon archive has `clisccp` and the CALIPSO variables but no MISR/MODIS histograms, and the one CAM case I found with `CLD_MISR` has a single month on an unstructured grid.
- **The Wasserstein path is untested.** The `ot_library` threading is fixed and `exact` is now the default, but neither the POT nor the `wasserstein` backend has been run on real data.
- The land/ocean masking and SE regridding helpers are still inline, as the PR description notes. Those belong in shared utilities, which I'd suggest as a follow-up issue rather than growing this PR.

brianpm:
Merged up from current `main` and closed out both of the gaps flagged above: **MISR now runs end-to-end, and the Wasserstein path runs.** Diff against `main` is still the same 5 files.

### MISR, end-to-end

The blocker was data, not code. The CERESMIP run `ceresmip_amip02` turns out to carry `CLD_MISR` in its `h0` (monthly) stream alongside `CLMODIS` and `FISCCP1_COSP`, on a regular 192x288 grid — so no regrid weights needed. Tested against `MISR_obs_data.nc` + `MISR_6C_weather_state_centers.npy`, 1995-2004, 6.16M valid profiles.

Bin counts and units line up on both sides, as the table above predicted: model `cosp_tau` (7) x `cosp_htmisr` (15, after the lowest-bin drop) = 105, obs `tau` (7) x `cth` (15) = 105, centers `(6, 105)`, everything in percent. No units warning fires. The `Dropping lowest height bin (cosp_htmisr)` trim executed for the first time.

Results are physically sensible and reproduce known CAM biases in MISR-simulator space:

| | Obs RFO | CAM RFO |
|---|---|---|
| CR5 low thin (trade Cu / Sc) | 17.7% | **0.4%** |
| CR6 high thin (cirrus) | 32.5% | **51.3%** |

The CR5 map puts the observed regime exactly where trade cumulus and stratocumulus live (subtropical eastern ocean basins, ITCZ flanks) and CAM has essentially none of it. RFOs sum to 100% on both sides.

This run also exercised the `interp_like` regrid in `plot_rfo_maps` for the first time, since the obs labels are 180x360 and the model is 192x288.

### Three things that run exposed

1. **MISR height axis was mislabelled by 1000x.** `_add_figure_labels` used `ht_unit = "m"` while `_configure_misr_axes` writes its ticks in kilometres. Only MISR can reach that branch — ISCCP and MODIS use `cosp_prs` and take the hPa branch — which is why it survived until now.

2. **The euclidean distance OOM-killed a 16 GB job.** `np.sum((mat[:, :, None] - cl.T[None, :, :]) ** 2, axis=1)` materialises an `(n_profiles, n_bins, n_centers)` temporary: 6.2M x 105 x 6 is ~17 GB in float32, for a result that is only `(n_profiles, 6)`. Computing one center at a time peaks at a single `(n_profiles, n_bins)` array — 5.9x less memory, and bit-identical output (same argmin, max abs diff 0.0; the full run's RFOs are unchanged to the decimal). Worth knowing regardless: a decade of MISR at 1 degree peaked at 45 GB before this change.

3. **`lat_range` / `only_ocean_or_land` were not applied to pre-computed obs labels.** The test case gets them via `_preprocess_data`; the obs branch returned the labels straight from the file, so the reference RFO covered the globe while the test RFO covered only the requested region. The example config in this PR advertises `lat_range: [-30, 30]`, so this was reachable exactly as documented. Confirmed on a +/-15 degree run — observation RFOs move from `9.3/16.0/15.6/16.4/19.2/23.5` to `18.4/20.1/18.9/3.1/6.7/32.7`: the mid-level and stratocumulus regimes drop out and trade cumulus grows, which is what restricting to the tropics should do. Time is deliberately left alone, since the obs record is its own climatology and need not overlap the model years.

### Wasserstein / EMD

Runs, via POT + joblib with the `exact` algorithm and the MISR `emd` centers and `emd_cluster_labels`. Tested on 1995-1996, +/-15 degrees: 211,968 profiles in **52 s on 8 cores**. Extrapolating, the full 10-year global field would be roughly 25 minutes on 8 cores — slower than euclidean but not prohibitive. Note that `_compute_distances_pot` dispatches one joblib task per profile; batching would help if anyone wants to run this routinely, but I have left that alone.

The cost matrix bin ordering is tau-major, matching the `tau_ht=(tau_var, ht_var)` stacking used to flatten the histograms, so the ground metric and the histograms agree.

### Cloud regimes now get their own website tab

The merge brings in #463, which fixes `adf_web.py` to append the plot *type* rather than the whole `plot_types` dict. That was the blocker recorded in the `WEB_PLOT_TYPE` comment, so this diagnostic no longer has to borrow the generic `Special` tab. Verified with `create_html: true`: `mean_diag_CloudRegimes.html` is generated, `index.html` links to it, and all 7 plot pages match their figure names with no broken hrefs or missing images anywhere in the site.

### Still not covered

- `regrid_weights_file` / the unstructured-grid path. The CERESMIP data is a regular grid, so this remains unexercised.
- MODIS model-vs-obs. `CLMODIS` is in the same `h0` stream, so this is now easy to run if a reviewer wants it — I just have not.
- The `wasserstein` library backend (only `pot` has been run).
- Land/ocean masking and SE regridding are still inline, as noted before; I would still suggest a follow-up issue rather than growing this PR.

`pytest lib/test/unit_tests` passes (17); clean compile, no undefined names.

brianpm-ucar:
## Review notes

Reviewed against `AGENTS.md`, on the branch **merged with current `main`** (`9c24110`, which now contains #435 and #467). **No blockers.** The items below are worth fixing, but none of them stops this merging.

### Re-tested end to end on the merged code

This branch was based on `7b19576`, before both #435 and #467, and #435 reworked `adf_dataset.py` — which this script reads through. So I merged current `main` into the head (clean, no conflicts) and re-ran both configurations against real data on Casper.

**Model vs. obs — MISR**, `ceresmip_amip02` 1995-2004 against `MISR_obs_data.nc`, 6.16M valid histograms, euclidean:

- Exit 0, 7 figures, website built, no dead links.
- **All seven PNGs are byte-identical to the pre-merge run of 2026-08-21.** Merging #435 and #467 changes nothing about this diagnostic's output.
- RFOs sum to 100.1% (obs) and 100.0% (test); the MISR height axis is labelled km, matching its ticks.

**Model vs. model — ISCCP**, `CESM2_amip` vs `CESM2_amip_p4K` 2000-2001, 1.24M histograms:

- Exit 0, 8 figures, three-column layout, RFOs sum to 100.0% on both sides.
- This path matters because it goes through `load_reference_timeseries_da`, which #435 touched and #467 re-pointed at `find_ts_files`. It is unaffected.

Also confirmed: 71 unit tests pass on the merged tree, `py_compile` is clean, and the website wiring is right — the `CloudRegimes` tab links `CLD_MISR`, and the plot page's regime selector reaches all seven pages (Centers, CR1-CR6). Nothing is orphaned.

### Worth fixing

- **`compute_cluster_labels` can return `None`, and the test-case caller does not check it** (`cloud_regime_analysis.py:265`). `test_labels.attrs['k'] = ...` raises `AttributeError` and ends the whole run with a traceback, which AGENTS §4.5 rules out for `scripts/`. Two ways to get there: no valid histograms after masking (line 445) — reachable with `only_ocean_or_land: "L"` and a `lat_range` that is all ocean — and `precomputed_clusters` returning `None`, which happens if `wasserstein` is requested in an environment without POT, or `ot_library` is misspelled. The reference path is already guarded (`if ref_labels is None: continue`); the test path needs the same two lines.
- **Same root cause one level down**: when `precomputed_clusters` returns `None`, `cluster_labels_flat[is_valid] = labels_valid` raises `TypeError` before the caller ever sees it. An early `if labels_valid is None: return None` makes both failures land as a warning-and-skip.
- **`LongitudeFormatter(zero_direction_label=True)`** (line 784) labels the prime meridian `0°W`, visible on every RFO map. Dropping the kwarg gives the correct `0°`.
- **`central_longitude` from `diag_basic_info` is ignored.** My config sets 180 and the maps still come out centred on 0. Other ADF plot scripts honour it.
- **pylint E0606** at line 520: `distances` is possibly unbound if `wasserstein_or_euclidean` is neither `"euclidean"` nor `"wasserstein"`. Unreachable through the entry point, since `_validate_user_inputs` coerces the value, but a `distances = None` before the branch closes it.
- **Style (AGENTS §6.1)**: 55 added lines carry trailing whitespace and 42 exceed 100 columns. `darker --revision main...HEAD` clears both without reformatting the untouched parts of the file.
- **`shapely` is imported at module scope** (lines 17-18) but is not in `env/conda_environment.yaml`. It arrives transitively through cartopy, so nothing breaks — worth listing it explicitly alongside the new `pot` and `joblib` entries, or dropping the direct import.
- Minor: `except Exception as e` in `load_cluster_centers` wants the one-line "why" that AGENTS §6.3 asks for; `_save_figure` takes an unused `fig`; `im` at line 1012 is unused.

### Follow-up / not verified

- The **`wasserstein` library backend** (as opposed to `pot`) and the **`regrid_weights_file` / unstructured-grid path** are still untested — no suitable data on GLADE for the latter beyond a single month on an `ncol` grid.
- `_land_mask_values` pulls cartopy's Natural Earth 110m coastline, which downloads on first use. Fine on Casper; worth knowing if this ever runs somewhere without outbound network.
- MODIS model-vs-obs is now cheap to exercise (`CLMODIS` is in the same CERESMIP h0 stream) and would close the last product gap.

### Things I checked and found correct

Entry-point name matches the module; the example-config entry is commented out with the kwargs dict form spelled out, which is the established opt-in pattern; `add_website_data` is called for every figure with `plot_type=WEB_PLOT_TYPE`; `spatial_subset` handles descending latitude properly; `_calculate_rfo` masks to valid labels so the map and the scalar agree; the euclidean distance is computed one centre at a time to avoid the 17 GB temporary; and the deliberate shortcuts carry `ponytail:` comments naming their upgrade path.

brianpm-ucar:
## Review fixes applied (`6a2f02a` → `8708355`)

Four fixes from the review above, in two commits. All three CI workflows pass on the new head, and it merges cleanly into current `main` (`9c24110`).

### ⚠️ One user-visible change: the maps now default to 180-centred

`plot_rfo_maps` ignored the run's `central_longitude` and always drew maps centred on 0. It now uses `plotting_utils.get_central_longitude`, the same helper `global_latlon_map` uses, with the x ticks placed relative to whatever it returns. **That helper's fallback is the ADF-wide default of 180**, so a run that does not set `central_longitude` gets Pacific-centred RFO maps where it previously got 0-centred ones. Consistent with the rest of the ADF, but worth a release note.

RFO values are unaffected — MISR CR3 reports 13.3% / 24.2% / 11.0% before and after. Only the framing moved.

### The other three

- **`compute_cluster_labels` could end the whole run with a traceback.** It returns `None` in two situations — no valid histograms survive the masking, and `precomputed_clusters` could not compute distances (POT missing while `wasserstein` is requested, or a misspelled `ot_library`). The reference path already checked for that; the test-case path did not, so `test_labels.attrs['k']` raised `AttributeError`. It could not even get that far in the second case: assigning `None` into the label array raised `TypeError` first. Both now warn and skip the variable, per AGENTS.md §4.5. Checked directly: an all-NaN field and a bad `ot_library` each return `None` cleanly, and the normal path is unchanged.
- **`LongitudeFormatter(zero_direction_label=True)`** labelled the prime meridian `0°W` on every RFO map. Keyword dropped.
- **A blank meridian down the difference panel.** Recentring exposed a seam that had been hidden at the map edge. It is real: the MISR obs grid runs −179.5 to 179.5 while CAM has a column at exactly −180, so `interp_like` left that column NaN. Now `kwargs={"fill_value": "extrapolate"}` — for nearest-neighbour that takes the closest cell, which across the dateline is the right one.

### Re-verified after the fixes

Both configurations re-run on the branch merged with current `main`:

| run | result |
| --- | --- |
| MISR model-vs-obs, `ceresmip_amip02` 1995-2004, 6.16M histograms | exit 0, 7 figures, seam gone, RFOs sum to 100% |
| ISCCP model-vs-model, `CESM2_amip` vs `_p4K` 2000-2001 | exit 0, 8 figures, RFOs sum to 100% on both sides |

### Still open, deliberately not fixed here

- pylint E0606: `distances` is possibly unbound in `precomputed_clusters` if the metric is neither `"euclidean"` nor `"wasserstein"`. Unreachable through the entry point, since `_validate_user_inputs` coerces the value.
- 55 added lines with trailing whitespace, 42 over 100 columns. `darker --revision main...HEAD` clears both without touching the rest of the file.
- `shapely` is imported at module scope but is not in `env/conda_environment.yaml`; it arrives transitively through cartopy, so nothing breaks.
- Untested paths: the `wasserstein` library backend (as opposed to `pot`), and `regrid_weights_file` on unstructured grids. MODIS model-vs-obs is cheap to add now that `CLMODIS` is known to be in the same CERESMIP h0 stream.

PR #419, opened 2025-10-03
Add regional map for structured grids

megandevlan:
Add capability to plot the regional definition in a map when using structured grids. Currently, only worked for unstrucutred grids.


megandevlan:
I tested it with the cases in config_clm_structured (ctsm53065_54surfdata_PPEcal115_115_HIST), which I believe is regular FV grid? Assuming I'm right on that - yes! It works and should be good to merge :)

PR #420, opened 2025-10-15
Create external scripts to break up `lib/plotting_functions` into organized helper methods

justin-richling:
This PR will organize helper methods that currently live in `lib/plotting_functions.py` with the intent of bringing any ADF-wide utilities (non plotting related) to an external script `lib/adf_untils.py` so methods such as validating dimensions like `zm_validate_dims`, or averaging schemes like `seasonal_mean`, or any other non-plotting helpers like `mask_land_or_ocean`, etc would now fall under this script.

A second external script `lib/plotting_utils.py` is created for helper methods related to plotting such as `prep_contour_plot` and `get_central_longitude`, etc. This script would be called in `plotting_functions.py` so that  `plotting_functions.py` can be specific to only plotting schemes for diagnostic plotting scripts. 

Breaking these methods up into external scripts will not only organize helper functions better, but it will eliminate roughly 1000 lines from `plotting_functions.py`, specifically ones that are not related to plotting.

Additionally, this PR will create a warnings function in `lib/adf_utils.py` 

```python
#Format warning messages:
def my_formatwarning(msg, *args, **kwargs):
    """Issue `msg` as warning."""
    return str(msg) + '\n'
```

that is defined in most scripts currently so it can be centralized and called in scripts via `adf_utils`:
```python
import warnings  # use to warn user about missing files.
import adf_utils as utils
warnings.formatwarning = utils.my_formatwarning
```

I'm open to also making a plotting directory in `lib` to organize plotting scripts since we will probably want to extend color maps to NCL soon thanks to the work of @brianpm

### DIFF
--- a/lib/adf_web.py
+++ b/lib/adf_web.py
@@ -717,7 +717,7 @@ class AdfWeb(AdfObs):
             avail_external_packages = {'MDTF':'mdtf_html_path', 'CVDP':'cvdp_html_path'}
             
             #Construct index.html
-            index_title = "AMP Diagnostics Prototype"
+            index_title = "CAM Diagnostics"
             index_tmpl = jinenv.get_template('template_index.html')
             index_rndr = index_tmpl.render(title=index_title,
                                             case_name=web_data.case,


scripts/analysis/amwg_table.py
```diff
@@ -224,7 +225,7 @@ def amwg_table(adf):
             #End if
 
             #Load model variable data from file:
-            ds = pf.load_dataset(ts_files)
+            ds = utils.load_dataset(ts_files)
```
nusbaume: Just double-checking if importing `plotting_functions` is needed in this script anymore?  If not then I would probably recommend removing the `import plotting_functions as pf` line above.

lib/plotting_functions.py
```diff
@@ -134,519 +94,11 @@ def my_formatwarning(msg, *args, **kwargs):
...
-                    difference:Optional[xr.DataArray]=None,
-                    pctchange:Optional[xr.DataArray]=None,
-                    domain:Optional[list]=None, 
-                    hemisphere:Optional[str]=None, 
-                    obs=False, 
-                    **kwargs):
+def make_polar_plot(adfobj, wks, case_nickname, base_nickname,
+                    case_climo_yrs, baseline_climo_yrs,
+                    d1:xr.DataArray, d2:xr.DataArray, difference:Optional[xr.DataArray]=None,pctchange:Optional[xr.DataArray]=None,
+                    domain:Optional[list]=None, hemisphere:Optional[str]=None, obs=False, **kwargs):
```
nusbaume: Was there a reason the argument spacing here was changed?  If not then I might request changing it back to a "one argument per line" style like it was previously.

lib/plotting_functions.py
```diff
@@ -741,7 +193,7 @@ def make_polar_plot(wks, case_nickname,
     absmaxdif = np.max(np.abs(dif))
     absmaxpct = np.max(np.abs(pct))
 
-    if 'colormap' in kwargs:
+    '''if 'colormap' in kwargs:
```
nusbaume: Is there a reason to keep this entire section of commented-out code?  If not then I might recommend just removing it entirely.

lib/plotting_functions.py
```diff
@@ -1122,9 +601,9 @@ def plot_map_vect_and_save(wks, case_nickname, base_nickname,
     #Set stats: area_avg
     ax[0].set_title(f"Mean: {mdl_mag.weighted(wgt).mean().item():5.2f}\nMax: {mdl_mag.max():5.2f}\nMin: {mdl_mag.min():5.2f}", loc='right',
                        fontsize=tiFontSize)
-    ax[1].set_title(f"Mean: {obs_mag.weighted(wgt).mean().item():5.2f}\nMax: {obs_mag.max():5.2f}\nMin: {obs_mag.min():5.2f}", loc='right',
+    ax[1].set_title(f"Mean: {obs_mag.weighted(wgt).mean().item():5.2f}\nMax: {obs_mag.max():5.2f}\nMin: {mdl_mag.min():5.2f}", loc='right',
```
nusbaume: Just double-checking that replacing `obs_mag` with `mdl_mag` is the correct thing to do here?  The same question holds for the `diff_mag` change below as well (and of course if I am just mis-understanding this change then please let me know!)

lib/plotting_functions.py
```diff
@@ -1160,7 +639,10 @@ def plot_map_vect_and_save(wks, case_nickname, base_nickname,
     fig.colorbar(img2, cax=cb_c2_ax)
 
     # Plot vector differences:
-    img3 = ax3.contourf(lons, lats, diff_mag, transform=ccrs.PlateCarree(), transform_first=True, norm=normdiff, cmap='PuOr', alpha=0.5)
+    img3 = ax3.contourf(lons, lats, diff_mag, transform=ccrs.PlateCarree(), transform_first=True, norm=normdiff,
+                        cmap='PuOr', alpha=0.5,
+                        #extend=cp_info["extend_diff"]
```
nusbaume: Remove or un-comment `extend` line here?

lib/plotting_functions.py
```diff
@@ -1311,21 +794,31 @@ def plot_map_and_save(wks, case_nickname, base_nickname,
...
         levs = np.unique(np.array(levels))
         if len(levs) < 2:
             img.append(ax[i].contourf(lons,lats,a,colors="w",transform=ccrs.PlateCarree(),transform_first=True))
             ax[i].text(0.4, 0.4, empty_message, transform=ax[i].transAxes, bbox=props)
         else:
-            img.append(ax[i].contourf(lons, lats, a, levels=levels, cmap=cmap, norm=norm, transform=ccrs.PlateCarree(), transform_first=True, **cp_info['contourf_opt']))
+            #print("******* GLOBAL LAT LON levels:",levels,"\n *******\n")
+            if mdlfld.name == "PRECT":
+                #print("cp_info['pct_colorbar_opt']",cp_info['pct_colorbar_opt'])
+                img.append(ax[i].contourf(lons, lats, a, levels=levels, cmap=cmap, transform=ccrs.PlateCarree(), transform_first=True, **cp_info['contourf_opt']))#extend='max'
```
nusbaume: Remove commented-out `extend` code here?

lib/plotting_utils.py
```diff
@@ -0,0 +1,830 @@
...
+            else:
+                msg += f"\n\t\tFile will be downloaded and saved as {filename}"
+                download_ncl_colormap(fil, str(filename))
+        else:
+            is_url = False
+            filename = Path(fil)
+    elif isinstance(fil, Path):
+        filename = fil
+    else:
+        raise ValueError(f"\t\tERROR: what to do with type {type(fil)}")
```
nusbaume: Instead of killing the ADF run here, would it be better to just send this error message to the debug log and then return `None` like what is done below?  That would in general fit better with the ADF design philosophy of trying to keep the ADF run going even when errors occur.

lib/plotting_utils.py
```diff
@@ -0,0 +1,830 @@
...
+                        #adfobj.debug_log(msg)
+                        return np.linspace(*entry), msg
+                    else:
+                        msg += f"\n\tLevels specified for {plotty} as list of 3 values, please add more values."
+                        msg += " Will get contrours from data range."
+                        #adfobj.debug_log(msg)
+                        return None, msg
+                elif len(entry) < 3:
+                    msg += f"\n\tNot enough {kind} entries for {plotty} (<3) — ambiguous"
+                    #adfobj.debug_log(msg)
```
nusbaume: Remove or un-comment the `debug_log` call here?  This same request holds for the other commented-out usages of `debug_log` below as well.

lib/plotting_utils.py
```diff
@@ -0,0 +1,830 @@
...
+
+        url = guess_ncl_url(cmappct)
+        locfil = "." / f"{cmappct}.rgb"
+        if locfil.is_file():
+            data = read_ncl_colormap(adfobj, locfil)
+        else:
+            data = read_ncl_colormap(adfobj, url)
+        cm, cmr = ncl_to_mpl(data, cmappct)
+        #ncl_colors[cm.name] = cm
+        #ncl_colors[cmr.name] = cmr
```
nusbaume: Remove or un-comment the two lines here?

nusbaume (changes requested):
Thanks for the great cleanup @justin-richling!  I have some change requests, but hopefully they will all be easy to handle (it is mostly about commented-out lines of code).

justin-richling:
Hi @nusbaume, it looks like almost all of these requested changes are resolved; I had some erroneous code that got committed in to this branch. My apologies for the extra work, but could you double check the `plotting_functions.py` and `plotting_utils.py` again when you get a chance? Thanks!

nusbaume (commented):
Thanks for removing the erroneous code @justin-richling!  I believe that has resolved all of my concerns except for the one request related to `amwg_table.py`.  Once that is resolved this PR will be good to go (at least for me)!

nusbaume (approved):
Everything looks great to me now.  Thanks again!

Issue #426, opened 2025-11-13
Add observations to Taylor diagrams


justin-richling:
### New feature type

New plot and/or plot enhancement

### What is this new feature?

We need to build the infrastructure for Taylor Diagrams to plot when comparing against observations.

### Assistance required?

Yes, I am requesting a new feature but won't have time to develop it

brianpm:
One step is complete. I just generated TAUX and RELHUM files for the ADF_obs database. Both based on ERA5.

brianpm-ucar:
Closing based on https://github.com/NCAR/ADF/pull/435

PR #435, opened 2026-01-15
Taylor Diagram with observations and multi-case

brianpm:
The main objective of this PR is to provide Taylor Diagrams with observation/reanalysis reference data, and thus to address #426. 

Since Taylor Diagrams are especially useful for comparing multiple simulations to a common reference, I also prioritized being able to run with multiple test cases. 

Refactoring the original `cam_taylor_diagram.py` mostly needed moving to use more of `adf_dataset`. The other key part was allowing for the potential need to regrid obs-to-obs for derived variables that the Taylor diagram uses. I also **refactored the `regrid_and_vert_interp.py` module**.  Both use **xESMF** for regridding, defaulting to conservative methods. The regridding weights are currently being stored in a subdirectory of the regrid location. 

For cases when we need to regrid from pressure levels to pressure levels, I added a plev_to_plev method in `adf_utils`. Elsewhere in that module I made some modifications to try to keep metadata.

I was running in to a hang when creating climo files, and I tracked it down to our use of multiprocessing. I think I've implemented a fix for that (which I think we've had to work around before). The main thing seems to be `mp.get_context('spawn')` but there's also some dask stuff to try to prevent zombie processes being made. This seems to work, but isn't crucial for this PR.

There are some updates to `adf_info.py` that help to deal with the multi-case configurations. This has to do with handling the specification of history files from the YAML file. *This could break previously working configurations* because it might make some assumptions about how those history files are specified. 

To read nested lists correctly (i.e., [['h0a', 'h1a'],['h0a', 'h1a']]), I also needed to modify `adf_config.py`.

I added the ERA5 land fraction to the observations repository and included it in the variable defaults file. 

A few modifications are included in `adf_dataset`, but mainly just trying to apply unit transformations more carefully. These might not be super necessary, but I don't think they will hurt anything.

---

## Added: vertical interpolation is now decided by the variable, not its file

Found while running this branch together with #467 (GenTS time series) on real
data. Neither branch shows it alone, so it is worth explaining.

`_determine_vertical_coord_type` inspected the **Dataset's** dimensions. ADF's own
`ncrcat` time series copy `hyam`/`hybm` only into files for variables that are
genuinely on model levels, so the dataset dims and the variable's dims agree and
nothing goes wrong. GenTS copies *every* non-time-varying variable — `lev`, `ilev`,
`hyam`, `hybm` — into *every* file, so a 2-D field such as `TS` arrives in a dataset
carrying a `lev` dimension it does not use:

| climo file | dataset has `lev` dim | variable on `lev` | detected type |
| --- | --- | --- | --- |
| ncrcat `TS` | False | False | `'none'` |
| GenTS `TS` | **True** | False | **`'hybrid'`** |
| GenTS `ICEFRAC` | **True** | False | **`'hybrid'`** |
| GenTS `PS` | **True** | False | **`'hybrid'`** |
| GenTS `Q` | True | True | `'hybrid'` |

The mis-detection sent 2-D fields into hybrid interpolation, where
`_handle_vertical_interpolation` does `da[lev_coord_name]` on a DataArray that has no
`lev`, raising `KeyError: 'lev'` out of `create_time_series` and ending the run with a
traceback — which AGENTS.md §4.5 rules out for framework code.

The Dataset is still passed, because the `hyam`/`hybm` fallback check needs the other
variables in the file; the DataArray now decides whether the variable needs
interpolating at all. With ncrcat output the answer is unchanged for every variable,
so this is a no-op for the existing back end.

Verified on real output from both back ends:

```
ncrcat   TS='none'  ICEFRAC='none'  PS='none'  Q='hybrid'
GenTS    TS='none'  ICEFRAC='none'  PS='none'  Q='hybrid'
```

29 unit tests pass on this branch.

### Note on the regridding fix in this PR

The same runs confirmed the problem this PR set out to fix. On `main`, the test-case
climo lookup globbed `{case}_{var}_*.nc` while files are named
`{case}_{hist_str}_{var}_climo.nc`, so nothing matched — three full ADF runs produced
**zero plots** and still printed "ADF diagnostics has completed successfully", because
with no variables surviving regridding the plotters' "all plots already exist" check is
vacuously true. With this branch merged in, the same configs produce 280 plots each.


brianpm:
_This comment authored by Claude, and so is very verbose. But it is comprehensive about what was changed in this latest update. At this point, the code seems to run and is improved. I might do another round of refactoring, but at this time I think it is ready for review/testing._

Branch `taylor_with_obs` was 42 commits behind `main` and conflicting. Merged
`main` in, reviewed the code, and verified with a full model-vs-model ADF run
(`b.e23_alpha17f.BLT1850.ne30_t232.098` vs `.093`, years 10–11, 16 variables).
That run now reports **"ADF diagnostics has completed successfully."** Unit
tests pass (18); the two CI-gated files clear the 9.5 pylint bar
(`adf_config.py` 9.52, `adf_info.py` 9.71).

## The blocker: the rewritten regridder stopped producing the reference

The old driver wrote two products per variable; the rewrite wrote only one.

| output | old (`main`) | new (as submitted) |
|---|---|---|
| `{target}_{case}_{var}_regridded.nc` | yes | yes |
| `{base}_{var}_baseline.nc` | yes | **never** |

The baseline file is **not** limited to 3-D fields — in the old driver a 2-D
variable passes straight through and is still written, so the file existed for
every variable in non-obs mode. Five scripts read it and nothing else produces
it: `zonal_mean`, `global_latlon_map` and `polar_map` via
`load_reference_regrid_da`; `meridional_mean` and `global_latlon_vect_map` by
globbing it directly. All five degrade with a warning rather than crashing, so
a model-vs-model run would have silently lost the reference and difference
panels from **every** plot.

Fixed by adding `_write_reference_files`, run once before the test-case loop
(the reference does not depend on the test case, and doing it first also fixes
an old-code quirk where the baseline was skipped whenever the test regridded
file already existed). It reuses the module's existing
`_determine_vertical_coord_type` / `_handle_vertical_interpolation` /
`_apply_ocean_mask` / `save_to_nc`. No horizontal regrid — the reference *is*
the target grid. Gated on `not compare_obs`, as the old driver was.

## Bugs the end-to-end run exposed — all on paths shared with the test case

These were latent in the PR; the reference pass just hit them first.


1. **`attrs['transformed'] = True` could not be written.** netCDF4 rejects a
   Python bool, so *any* variable carrying a scale factor failed to save. Now `1`.
2. **`DEFAULT_PLEVS_Pa` was a Python list.** geocat's
   `interp_hybrid_to_pressure` needs `.size` and `utils.vert_remap` needs
   `.shape`, so every hybrid- and height-coordinate variable crashed. Now an
   ndarray.
3. **`_create_clean_grid` clipped the target grid *centers* to ±89.999999.**
   Those centers become the output coordinate, so regridded files disagreed
   with the reference at the poles by 1e-6°, and xarray alignment silently
   dropped both polar rows from every difference field — surfacing as
   `IndexError: shape mismatch ... (190,1) (192,289)` in cartopy. Only the
   bounds are clamped now, to ±90.
4. **`load_climo_ds` dropped attrs.** `ds[v] = ds[v] * scale + offset` does not
   preserve them, so every regridded file lost `units` and `polar_map` raised
   `KeyError: 'units'`. (`main`'s `load_da` always forced a units attribute;
   this PR removed that safety net, so the loss became visible.)
5. **PS resolution.** ADF climo files for a 3-D variable carry `PS` alongside
   `hyam`/`hybm`, and the old driver preferred it. The rewrite only looked for
   a standalone PS climo, which exists only when PS is in `diag_var_list` — it
   is not in the standard config, so every 3-D variable was skipped. Added
   `_find_surface_pressure` to match the old precedence.
6. **`cam_taylor_diagram` could not honour its own requirement check.** It
   accepts "PRECT **or** PRECC+PRECL", but both the reference and test branches
   raised `KeyError: 'PRECT'` instead of returning `None` when PRECT was
   absent. Added `_load_field`/`_load_field_ds`, which return `None` for a
   variable that is not part of the run, and routed every loader through them.

## Other defects fixed while reviewing

Regridder: `source_grid` was undefined for unstructured input (`NameError`,
caught by pylint E0606); `load_climo_da(case,'PS')['PS']` indexed a DataArray;
the PS fallback passed a DataArray where a grid Dataset was expected;
conservative regridding had no bounds to work with because CAM climo files
carry none, so cell edges are now inferred from centers; two no-op
`final_ds = final_ds` blocks and the committed 753-line
`regrid_and_vert_interp.py.OLD` removed.

`cam_taylor_diagram`: reference data was recognized only when the label was
literally `'Obs'`, so model-vs-model looked for the baseline among test-case
regrid files; the obs weights directory was built from `Path(case_list[0])`,
a case *name* rather than a path; `get_var_at_plev`'s hybrid branch was
unreachable (`'hyam' not in dataarray` tests *values*, not variable names) and
was collapsed into the correct pressure-level path.

`adf_utils`: **`plev_to_plev` was silently wrong for surface-first data** —
`vert_remap` uses `np.interp`, which needs increasing x, so a descending
pressure coordinate (the CMIP convention) returned a constant field equal to
the topmost level. Verified both orderings now agree.

`adf_dataset`: None-guards on the new Dataset loaders and an unguarded
`variable_defaults[var]` lookup; `new_unit` must exist before a units match
counts as "already transformed".

`adf_config`: collapsed `__expand_dict_refs` + `__expand_list_refs` (35 largely
duplicated lines) into one recursive helper, with a unit test for nested-list
keyword expansion — the actual new capability, previously uncovered.

**`cam_regrid_loc` reverted to a single directory.** The per-case list broke
`meridional_mean.py:28` and `global_latlon_vect_map.py:62` (`Path()` of a list)
and `_regrid_loc_for_case` (the baseline is not in `case_names`). Regridded
filenames already carry the case name, so one directory suffices, and it is
what `config_cam_baseline_example.yaml:84` documents.

Housekeeping: removed leftover `CHECK ON INPUT` and `HIST_STR_USE` debug
prints, and repaired the comment glued onto the assignment at the end of
`hist_str_to_list`.

## Verified by the run

1. 16/16 variables have a `{base}_{var}_baseline.nc`, 2-D included.
2. Every 3-D reference is on the **identical** 21 pressure levels (hPa) as its
   test counterpart; difference fields keep all 192 latitudes.
3. Exactly one "did not find reference" warning in the whole log — the `U,V`
   vector pair, because `V` is not in this config's `diag_var_list`. Correct
   behaviour, not a regression.
4. Spot-checked plots (`PSL_ANN_LatLon`, `U_ANN_Zonal`) show all four panels —
   test, baseline, %diff, difference — with correct units and full pole-to-pole
   coverage.
5. Ocean masking reaches the reference: all 27,237 land cells in
   `{base}_SST_baseline.nc` are set to −999 with `missing_value` recorded.
   (Check this with `decode_cf=False`; xarray's CF decoding turns −999 into NaN
   and strips the attribute, which makes a correctly masked file look unmasked.)
6. `zonal_mean`, `global_latlon_map`, `global_latlon_vect_map`,
   `meridional_mean`, `polar_map`, `cam_taylor_diagram` and `qbo` all complete;
   the Taylor diagram derives PRECT from PRECC+PRECL and places all 10
   variables at ratio ≈ 1, correlation ≈ 0.99+, as expected for two near-identical runs.

## Worth deciding before merge

1. **`hist_str_to_list` is ambiguous.** With N cases, `["h0","h1"]` means "one
   stream per case" when N==2 and "both streams for every case" otherwise — the
   interpretation flips on the case count. Also `[conf_val for _ in range(n)]`
   hands every case the *same* list object; use `list(conf_val)`.
2. **The `cam_climo_info` scalar broadcast** (`[conf_val]` →
   `[conf_val] * num_cases`) is the change the PR description warns could break
   existing configs. Worth a note in the example config / release notes.
3. **`load_reference_climo_da` picks the wrong scale factors for obs.**
   `get_value_converters` compares `case == ref_labels[var]`; called with
   `ref_case_label` ("Obs") while `ref_labels[var]` is the source name (e.g.
   "ERA5"), it selects the model `scale_factor` rather than `obs_scale_factor`.
4. **`taylor_stats_single` reports bias as a percentage of the reference mean.**
   For `EquatorialPacificStress` the 5°S–5°N TAUX mean can be near zero, so the
   percentage — and hence the marker size — becomes meaningless.
5. **`weighted_correlation` assumes matching NaN masks.** Land/ocean-masked
   fields are masked with each case's *own* LANDFRAC, so case and reference can
   have different valid points. Relatedly, the test side masks with its
   conservatively-regridded OCNFRAC, which perturbs values near the 0.9
   threshold: ~205 of 27,237 land cells classify differently from the
   reference. Pre-existing behaviour, but a common mask would be cleaner.
6. **`_handle_vertical_interpolation` sources PMID from the climo Dataset
   itself**, whereas the old driver could read a separate `{target}_PMID_climo.nc`.
   Same limitation already applies to the test side; `height` coordinates are
   rare.
7. `polar_map.py:131-134` picks `base_name` with the `compare_obs` test inverted
   relative to `zonal_mean`/`global_latlon_map`. Harmless today, but asymmetric.
8. `meridional_mean` and `global_latlon_vect_map` still glob files by hand
   rather than using `adf.data`. Pre-existing; not this PR's job.


## Model-vs-obs run — the PR's headline feature

Also run end to end (`compare_obs: true`, same test case against ERA5 / CERES /
ERAI), on a Casper node: **completes successfully, Taylor diagram included.**
All 12 variables that have obs regridded onto three different observational
grids — ERA5 721×1440, CERES 180×360, ERAI 121×240 — none of which carry
lat/lon bounds, so this is a real workout for the inferred-bounds conservative
regridding. The four variables with no obs (PRECC, PRECL, TAUX, TAUY) skip with
warnings rather than raising, which is `_load_field` doing its job.

Three further defects it exposed, all fixed:

1. **`seasonal_mean` crashed on the PR's own ERA5 land mask, aborting every
   model-vs-obs run.** `ERA5_LSM_1deg_conservativeregrid.nc` is a
   single-timestamp file (`2013-11-29`, `hours since 1900-01-01`), not the
   12-month climatology (`time = 1..12`) every other file in `ADF_obs` uses.
   `load_da`'s `squeeze()` leaves a *scalar* time coordinate, so
   `data.time.dt.days_in_month` succeeds and sails past the existing guard,
   then `.sel(time=...)` raises `KeyError: no index found for coordinate
   'time'`. Time-invariant fields now pass through unchanged.
   **Consider regenerating that obs file as a proper 12-month climatology** so
   it matches the convention of every other file in the directory; the code fix
   is worth keeping either way.
2. **The Taylor diagram mislabelled its points whenever a variable was
   missing.** Points were numbered by a counter that advanced only for points
   actually drawn, while the figure legend numbers the full 10-variable list.
   In the obs run, 3 of 10 variables have no counterpart, so the 7 plotted
   points were numbered 1–7 and four carried the wrong name — the figure read
   as if `TropicalLandPrecip` were present when point 4 was really
   `Land2mTemperature`. Silent, and it only appears when variables are missing,
   which is the normal case in obs mode (model-vs-model had all 10, so it
   looked fine). Now numbered by position, leaving gaps; this also retires the
   "requires cases to have same order in DataFrame" caveat.
3. **`cam_taylor_diagram` carried a duplicate `_create_clean_grid`** that still
   clipped grid *centers* to ±89.999999 and never inferred bounds, so the
   regridding behind its derived variables ran without bounds (the `Latitude is
   outside of [-90, 90]` warnings) and would have reintroduced the polar
   misalignment fixed in the regridder. One implementation now lives in
   `adf_utils.create_clean_grid`.

Two obs-mode limitations that are **not** bugs but constrain the feature:

- ~~`EquatorialPacificStress` can never work against obs~~ — **resolved.** ERA5
  surface wind stress is now wired in: the TAUX climatologies already existed in
  `ADF_obs` but were referenced only by the `era5-1deg` defaults, and TAUY did
  not exist. TAUY was built from ERA5 `inss` (RDA ds633.1, 1979-2022) at 0.25°
  and 1° by `AMP_toolbox/make_era5_tau_obs.py`, and both defaults files now
  point at the matching resolution. The obs Taylor diagram went from 7 to 8 of
  10 variables; `EquatorialPacificStress` scores correlation 0.712, stddev ratio
  1.347, bias +26.2% (model -0.0287 vs ERA5 -0.0228 N/m² over 5°S-5°N — the
  model's easterly trades are ~26% stronger).

  Two things that fix uncovered, both worth knowing:
  - **The pre-existing TAUX obs file stored latitude north-to-south**, unlike
    every other file in `ADF_obs`. `adf_utils.domain_stats` (`lib/adf_utils.py:435`)
    slices with `sel(lat=slice(lo, hi))`, which returns an **empty** selection on
    a descending axis, so `polar_map` died in `nanmin` with
    `a.size==0`. Fixed in the data (a lossless sort, verified statistics
    bit-identical), but `domain_stats` silently mis-slicing any north-to-south
    dataset remains a live footgun in ADF.
  - **The percentage bias is ambiguous for a signed quantity.** +26.2% here means
    the model stress is 26% *larger in magnitude* in the same (negative)
    direction, not weaker. See the `taylor_stats_single` item below.
- ~~The two tropical-precip entries need `PRECT` in `diag_var_list`~~ —
  **resolved.** The derivation pathway itself works: ADF derives PRECT from
  PRECC + PRECL when building the time series (verified numerically identical
  to their sum), and it regrids onto ERAI's grid in mm/day matching the obs
  (model 3.101 vs ERAI 2.884 mm/day, a +7.5% wet bias). The entries were
  missing only because PRECT was absent from `diag_var_list`.

  The real defect was the requirement check: it accepts "PRECT **or**
  PRECC+PRECL", which is right against a baseline simulation but wrong under
  `compare_obs`, where neither PRECC nor PRECL has an observational
  counterpart. The check passed and both entries then vanished with no
  indication why. It now warns explicitly and points at PRECT.

  **The obs Taylor diagram is now complete — all 10 variables.**
  TropicalLandPrecip corr 0.851 / ratio 0.882 / bias +2.3%;
  TropicalOceanPrecip corr 0.858 / ratio 1.001 / bias +1.1%;
  EquatorialPacificStress corr 0.712 / ratio 1.347 / bias +26.2%.

## Not exercised

Unstructured (`ncol`) input: the `source_grid` `NameError` fix on that path is
reasoned, not run — both test cases here are regular lat/lon.

Memory note: regridding a 3-D CAM field to ERA5 0.25° needs ~3.2 GB per
variable, and the horizontal regrid runs *before* the vertical interpolation,
so it inflates to 721×1440 while still carrying all 32 hybrid levels. The run
peaked near 60 GB and is killed outright on a login node. Same ordering as the
old driver, so not a regression, but obs comparisons against high-resolution
reanalysis need a batch job with real memory.

brianpm:
Fixed the seasonal plots.
Confirmed multi-case is working (with a couple adjustments). See example.
Implemented correct PMID precedence (over hybrid coeffs).

image

brianpm-ucar:
## Review notes

Reviewed against `AGENTS.md`. Reproduced the CI checks locally: `pytest lib/test/unit_tests` passes (33 now), pylint clears the 9.5 threshold on the three gated files this PR touches (`adf_config` 9.52, `adf_info` 9.71, `adf_web` 9.70), `py_compile` is clean, and `xesmf>=0.8.8` is already in `env/conda_environment.yaml`.

Two things were blocking; both are fixed in 4e142cd. Everything else below is for a follow-up PR.

### Fixed here

**Regridded files could be converted twice, or not at all.** The regridding stage now applies the variable-defaults conversion when it writes, and the loaders stopped applying it at plot time. Two gaps in that hand-off:

- `load_da` / `load_climo_ds` stamped `transformed` only when the variable also has a `new_unit`. `TAUX` and `TAUY` are scaled by `-1` with no `new_unit`, so their regridded files came out converted but unstamped. Confirmed on this branch's own output: `PRECT` and `PS` carry `transformed=1`, `TAUX` does not.
- `apply_scaling` defaulted to `False` unconditionally. Regrid output is named identically on `main` (`{rlbl}_{case}_{var}_regridded.nc`) and the shipped default is `cam_overwrite_regrid: false`, so a run reusing an existing `cam_regrid_loc` picks up `main`-era files — which are raw, because `main` converted at plot time. That plots PRECT in m/s against mm/day contour levels with no error and no blank plot.

`apply_scaling` now defaults to `None`, meaning "decide from the file": convert unless the data carries the stamp. `True`/`False` still force it. The stamp is only consulted for variables that actually have a conversion configured, so there is no extra file open in the common case. Covered by `lib/test/unit_tests/test_regrid_scaling.py`.

⚠️ Regrid directories written by earlier commits **on this branch** are converted but unstamped, so they will now be converted a second time. Delete them or re-run with `cam_overwrite_regrid: true`.

**The new unit tests would have failed CI.** `ADF_unit_tests.yaml` installs only `pyyaml` and `pytest`; the three new test files import numpy/xarray (and through them xesmf, matplotlib, geocat), so collection would have errored and taken the existing 18 tests down with it. This hasn't shown up yet only because the workflow is sitting at `action_required` pending fork approval. Guarded with `pytest.importorskip`; verified in a bare venv — 18 passed, 4 skipped.

### For a follow-up PR

Nothing below blocks the merge.

*Correctness*

- `lib/adf_utils.py`, `lev_to_plev`: in the `convert_to_mb=False` branch, `data_interp_rename.attrs['units'] = "Pa"` is set on the **data variable**, not on the `lev` coordinate — it relabels temperature as Pa. Dormant today because the only caller passes `convert_to_mb=True`, but it is a latent unit bug in a public util.
- `cam_taylor_diagram.py`: `.sel(lat=slice(-30,30))` and `.sel(lat=slice(-5,5))` return an **empty** array on a descending-latitude grid, and the statistics then come back NaN with no message. ERA5-derived reference grids are a real source of descending latitude. Worth an explicit `.sortby('lat')` or an order-aware slice.
- `lib/adf_dataset.py`, `load_da`: `units` is no longer set unconditionally, so a variable whose file carries no `units` attribute now has none at all. `polar_map.py:48` does a bare `mseason.attrs['units']` and would `KeyError`. CAM output always has units, so this is unlikely — a `setdefault('units', 'none')` restores the old guarantee for free.
- `lib/adf_info.py`, `hist_str_to_list`: a nested list whose length ≠ `num_cases` (e.g. `[['h0','h1']]` with two cases) falls through to the "same streams for every case" branch and becomes triple-nested, which later crashes inside `starting_location.glob()` with a confusing message. Reject wrong-length nesting explicitly, and use `self.end_diag_fail()` rather than `raise ValueError` (AGENTS §4.5).
- `regrid_and_vert_interp.py`: `_handle_vertical_interpolation` raises `ValueError` in three places. Two are unreachable from the main loop (the caller already `continue`s), but the missing-`hyam`/`hybm` one is reachable and kills the whole run over one bad variable. AGENTS §4.5 wants warn-and-skip in `scripts/`.

*Robustness / design*

- Weight files are keyed only on grid **sizes** (`weights_structured_192_288_to_structured_192_288_conservative.nc`), and they persist in `cam_regrid_loc/regrid_weights` across runs. Two different grids of the same shape would silently reuse the wrong weights. Add the first/last lat and lon, or a hash of the coordinate arrays.
- `silence_output()` calls `os.dup2` on `sys.stdout.fileno()`, which raises `io.UnsupportedOperation` wherever stdout is not a real file descriptor — including the `jupyter_sample.ipynb` driver this repo ships. Wrap the fd juggling in a `try` and fall back to `redirect_stdout` alone.
- `np.where(landfrac >= .95, prect, np.nan)` bypasses xarray alignment; `prect.where(landfrac >= 0.95)` aligns by coordinate.
- `plev_to_plev` drops all attributes and leaves a `## DEAL WITH METADATA BETTER HERE` marker. The one caller restores the data attrs, so only the `lev` units are lost.
- `p0 = source_ds.get('P0', 100000.0)` then `p0.values[0]` is an `IndexError` on a 0-d `P0`. `main` made the same assumption (`P0_tmp[0]`), so this is pre-existing, not a regression.

*Documentation and style*

- **The default regridding method changed from xarray `interp_like` (bilinear) to xESMF conservative.** That changes every regridded field and every plot the ADF produces. It is a defensible improvement, but it is not mentioned in the PR description and belongs in the release notes.
- 40 added lines carry trailing whitespace and 21 exceed 100 columns; `darker --revision main...HEAD` clears both without reformatting the surrounding files (AGENTS §6.1).
- `cam_taylor_diagram(adfobj)` has a one-line docstring. `main` had none, so this is not a regression, but since the file is a rewrite it is the moment to add the NumPy-style `Parameters`/`Returns` plus the ADF `Notes` block listing the `adfobj` attributes used (AGENTS §6.2).
- `LANDFRAC.obs_file` in the **native-resolution** `adf_variable_defaults.yaml` points at `ERA5_LSM_1deg_conservativeregrid.nc`. That is the only LSM file in `ADF_obs`, so it is the right choice — worth a comment so nobody "fixes" the filename later. All five new obs files exist on GLADE.

### Not verified

Plot correctness (marker placement, bias sizing, legend alignment), the conservative-vs-bilinear science difference, and whether the ERA5 `TAUX` `scale_factor: -1` sign convention matches the model. Those need eyes on the figures and the dataset.

PR #444, opened 2026-03-16
Handle empty dict if no git info is available

TomasTorsvik:
ADF will fail if `get_git_info()` returns `None`, which is the case if no git repository is found.

This PR suggests a way to handle such a case, so that the processing can continue. Perhaps there are better ways, so please feel free to suggest alternative solutions.

### DIFF
--- a/lib/adf_web.py
+++ b/lib/adf_web.py
@@ -250,11 +250,16 @@ class AdfWeb(AdfObs):
 
             # Gather Git info
             git_info = self.get_git_info()
-            f.write("\n")
-            f.write(f"<br><br><strong><a {font_22}>Git Info</a></strong><br>")
-            for key,val in git_info.items():
-                f.write(f"{two_space}<a {font_16}><strong>{key}:</strong> {val}</a></><br>")
-            f.write("</p>")
+            if git_info is None:
+                f.write("\n")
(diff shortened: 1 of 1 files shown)


TomasTorsvik:
@justin-richling , @nusbaume , @YanchunHe 
I suppose we don't have an established workflow for pushing changes upstream from NorESM (https://github.com/NorESMhub/ADF). Would be good to have at least some contact points. Normally, we want to only inherit the NCAR changes, so it's probably only exceptional cases where we want to push something back.

justin-richling (approved):

justin-richling:
@TomasTorsvik I apologize for letting this PR get away from me. Thanks for the suggestion; I did not think about the lack of git! 

I'm happy to be the point of contact for your changes or @nusbaume would also be a good contact as well, but if you do have some potential changes, we'd always be happy to work with your team. I'm not sure how it looks on your end, but you can either just `@` one of us on the PR comments, or request one or both as a reviewer. 

Thanks again for the code and the support of the ADF!

TomasTorsvik:
@justin-richling - thanks for responding! I think we are mostly including a few modifications for our HPC/storage and NorESM-specific changes, but we are happy to contribute upstream to general ADF development as well. I will ping you and @nusbaume if something more comes up from our side.

PR #448, opened 2026-04-03
Fix `None` for `fils` in `load_reference_climo_da`

### LINKED ISSUE #447: AOD lat/lon diags break if `AODVISdn` is in the variable list in config but not in actual files
### ADF run type

Model vs. Model

### What happened?

If the user supplies `AODVISdn` in the `diag_var_list` but the reference climo files don't get created, then the ADF will crash with a `None` type error in `lib/adf_dataset.load_dataset`

This is because in `aod_latlon` when calling `load_reference_climo_da` from `lib/adf_dataset.py`, this calls `get_reference_climo_file` with the problematic line:

```python
fils = sorted(Path(ref_loc).glob(f"{self.ref_case_label}_{var}_climo.nc"))
if fils:
    return fils
return None
```

Then later calling `load_da` and thus `load_dataset`, this line is where it breaks:

```python
if (len(fils) == 0):
```

Because `fils` comes back as None, instead of an empty list.

This would be an easy fix, probably just need to return `fils` after 

`fils = sorted(Path(ref_loc).glob(f"{self.ref_case_label}_{var}_climo.nc"))`

since a missing file with glob will make an empty list by default.

### ADF Hash you are using

7aaa242

CGD machine, CISL machine

NPL (CISL machines only)

### Extra info

_No response_

justin-richling:
Quick fix for broken `len` if `fils` is `None` for AODVISdn for the reference case

Will close #447

### DIFF
--- a/lib/adf_dataset.py
+++ b/lib/adf_dataset.py
@@ -199,7 +199,8 @@ class AdfData:
         a = self.adf.get_cam_info("cam_climo_loc", required=True) # list of paths (could be multiple cases)
         caseindex = (self.case_names).index(case) # the entry for specified case
         model_cl_loc = Path(a[caseindex])
-        return sorted(model_cl_loc.glob(f"{case}_{variablename}_climo.nc"))
+        fils = sorted(model_cl_loc.glob(f"{case}_{variablename}_climo.nc"))
+        return fils
 
 
     # Reference case (baseline/obs)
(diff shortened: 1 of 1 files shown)


justin-richling:
Looking at this now, I think the AOD lat/lon should be pointing to the regridded climo files not the raw climo?

justin-richling:
Merging this now and will address the AOD climo files in a separate issue.

PR #454, opened 2026-05-14
Add plotting enhancements

### LINKED ISSUE #373: Add option for contour ranges for 3d variables at different levels
### New feature type

New plot and/or plot enhancement

### What is this new feature?

It was brought to my attention that there is a desire to have the option to set the contour ranges for different levels in 3d variables for plotting. 

Ie setting contours for 250 and 800mb for winds.

This shouldn't be too difficult, just need to work on the execution. 

### Assistance required?

### Extra info

_No response_

### LINKED ISSUE #425: Clean up Meridional plots to mimic code of Zonal plots
With the recent clean up of the Zonal plots, these enhancements/clean ups should also be extended to Meridional plots.

### LINKED ISSUE #452: Make PR for Plotting Enhancements
### New feature type

New plot and/or plot enhancement

### What is this new feature?

Add flexibility for plotting (contour levels, colormap, etc.) based on pressure level and/or plot type (polar vs global)

Update plots for more professional look, ie Zonal and Meridional are very bare boned, lat/lon were missing units, some variables were not getting proper scaling, fit colorbar better to axes and extend if necessary, etc.

Combine Lat/Lon Vector into Lat/Lon code since it was a lot of reused code.
 - NOTE: the calculation for percent difference is currently off for lat/lon vectors plots, so this map is being ignored and vector plots default back to the two cases on top and difference centered below

Add option in variable defaults config for variable nickname; remove all percent diff args in defaults yaml and hard code in plotting script. The user will still have the option to supply these args in the yaml file if default is not desired.

Add option for old NCL color maps

Add units cleaning method to ensure units are always available, this was not enforced throughout the code base.

Add code to resolve levels for hemisphere and resolve colormap in `prep_contour_plot`

### Assistance required?

### Extra info

_No response_

justin-richling:
### New feature type

New plot and/or plot enhancement

### What is this new feature?

Add flexibility for plotting (contour levels, colormap, etc.) based on pressure level and/or plot type (polar vs global)

Update plots for more professional look, ie Zonal and Meridional are very bare boned, lat/lon were missing units, some variables were not getting proper scaling, fit colorbar better to axes and extend if necessary, etc.

- NOTE: There is an additional enhancement for 2d line plots for Zonal/Meridional that allows a color map for the 2d variables (See SWCF example below - variable defaults argument: `colormap_2d`)

Combine Lat/Lon Vector into Lat/Lon code since it was a lot of reused code.
 - NOTE: the calculation for percent difference is currently off for lat/lon vectors plots, so this map is being ignored and vector plots default back to the two cases on top and difference centered below

Add option in variable defaults config for variable nickname; remove all percent diff args in defaults yaml and hard code in plotting script. The user will still have the option to supply these args in the yaml file if default is not desired.

Add option for old NCL color maps

Add units cleaning method to ensure units are always available, this was not enforced throughout the code base.

Add code to resolve levels for hemisphere and resolve colormap in `prep_contour_plot`

---

#### Core new features:

Different contours based on vertical levels

```yaml
RESTOM:
  colormap: "RdBu_r"
  contour_levels_range: [-100, 101, 5]
  diff_contour_range: [-10, 10, 0.5]
  polar_map:
    colormap:
      nh: "RdBu_r"
      sh: "terrain"
    contour_levels_range:
      #nh: [-100, 101, 5]
      sh: [-100, 0, 2.5]
   scale_factor: 1
  add_offset: 0
  new_unit: "W m$^{-2}$"
  category: "TOA energy flux"
  derivable_from: ['FLNT','FSNT']
```

<table>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

<table>
<tr>
    <td></td>
  </tr>
</table>

<table>
<tr>
    <td></td>
  </tr>
</table>

---

```yaml
SWCF:
  colormap: "Reds"
  polar_map:
    colormap:
      nh: "fakemap"
      sh: "Blues"
  contour_levels_range: [-150, 51, 10]
  diff_colormap: "BrBG"
  diff_contour_range: [-20, 21, 2]
  scale_factor: 1
  add_offset: 0
  new_unit: "Wm$^{-2}$"
  obs_file: "CERES_EBAF_Ed4.1_2001-2020.nc"
  obs_name: "CERES_EBAF_Ed4.1"
  obs_var_name: "toa_cre_sw_mon"
  obs_scale_factor: 1
  obs_add_offset: 0
  category: "TOA energy flux"
  colormap_2d: True
```

<table>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

<table>
<tr>
    <td></td>
  </tr>
</table>

---

```yaml
U:
  global_latlon_map:
    colormap: # if not specified, default to viridis for any missing values nested (levs, ie 850) or not
      200: "ncl_default"
      #850: "Spectral"
    contour_levels_range:
      200: [-10, 51, 2.5]
      850: [-10, 31, 2.5]
    diff_contour_range:
      200: [-15, 16, 2]
      850: [-5, 5.1, 0.5]
  polar_map:
    colormap:
      nh: "terrain"
      sh: "fake_cmap"
    contour_levels_range:
      nh:
        200: [-10, 31, 2.5]
        850: [-10, 13, 1]
      sh:
        200: [-10, 51, 5]
        850: [-10, 31, 2.5]
    diff_contour_range:
      200: [-10, 10, 1]
      850: [-5, 5.1, 0.5]
      #nh:
      #  200: [-15, 16, 2]
      #  850: [-5, 5.1, 0.5]
      #sh:
      #  200: [-15, 16, 2]
      #  850: [-5, 5.1, 0.5]
  zonal_mean:
    colormap: "ncl_default"
    contour_levels_linspace: [-10, 30, 15] # These will be applied to all zonal 3d mean plots
    #diff_contour_range: # default to figuring our the contour levels from difference from max/min
  scale_factor: 1
  add_offset: 0
  new_unit: "ms$^{-1}$"
  obs_file: "U_ERA5_monthly_climo_197901-202112.nc"
  obs_name: "ERA5"
  obs_var_name: "U"
  vector_pair: "V"
  vector_name: "Wind"
  category: "State"
```

<table>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

<table>
  <tr>
    <td></td>
    <td></td>
  </tr>
</table>

<table>
<tr>
    <td></td>
  </tr>
<tr>
    <td></td>
  </tr>
</table>

<table>
<tr>
    <td></td>
  </tr>
<tr>
    <td></td>
  </tr>
</table>

---

```yaml

Surface_Wind_Stress:
  nickname: "Surface Wind Stress"
  new_unit: "N m$^{-2}$"
  category: "Surface variables"

TAUX:
  new_unit: "N m$^{-2}$"
  vector_pair: "TAUY"
  vector_name: "Surface_Wind_Stress"
  category: "Surface variables"
  scale_factor: -1
  add_offset: 0

TAUY:
  new_unit: "N m$^{-2}$"
  vector_pair: "TAUX"
  vector_name: "Surface_Wind_Stress"
  category: "Surface variables"
  scale_factor: -1
  add_offset: 0

```

<table>
<tr>
    <td></td>
  </tr>
</table>

---

closes #452
closes #425
closes #373


env/npl-2024a_environment.yaml
```diff
@@ -0,0 +1,649 @@
+name: adf_npl-2024a
```
nusbaume, line 1: I personally don't think we should have an NPL environment file in the ADF (or LDF).  In general it's best  to only have the minimum number of packages you need to run all the ADF scripts, which I am guessing is significantly less than what is contained here.  

If you need help creating a more up-to-date environment file for the ADF then please let me know, as I would be happy to help work on it.  Thanks!

.gitignore
```diff
@@ -111,5 +111,8 @@ GitHub.sublime-settings
 !.vscode/extensions.json 
 .history
 
+#NCL RGB files
```
nusbaume, line 114: Total nit-pick, but I might add a space in the comment here to match the format of all the other comments in this file:


```suggestion
# NCL RGB files
```

scripts/plotting/polar_map.py
```diff
@@ -208,6 +211,8 @@ def polar_map(adfobj):
                 
             case_name = plot['case']
             case_idx = plot['case_idx']
+            vres["season"] = plot['season']
+            vres["hemi"] = plot['hemi']
```
nusbaume, line 215: Another nit-pick, but I might use the same quotation marks on both sides of the equation here (in case someone who doesn't know python well looks at the code and assumes the behavior is different):


```suggestion
            vres['season'] = plot['season']
            vres['hemi'] = plot['hemi']
```

scripts/plotting/global_latlon_vect_map.py
```diff
@@ -415,6 +422,23 @@ def global_latlon_vect_map(adfobj):
...
+                                vpseasons[s] = vpseasons[s].where(np.isfinite(vpseasons[s]), np.nan)
+
+                                vres["umdlfld_nowrap"] = umseasons[s]
+                                vres["vmdlfld_nowrap"] = vmseasons[s]
+                                vres["uobsfld_nowrap"] = uoseasons[s]
+                                vres["vobsfld_nowrap"] = voseasons[s]
+                                vres["udiffld_nowrap"] = udseasons[s]
+                                vres["vdiffld_nowrap"] = vdseasons[s]
+                                vres["upctdiffld_nowrap"] = upseasons[s]
+                                vres["vpctdiffld_nowrap"] = vpseasons[s]
```
nusbaume, line 437: I realize this is purely an opinion, but in general I feel that "kwarg"-like dictionary inputs into python functions should only contain metadata (e.g. `plot_type`), and not required data like the fields being plotted.  Given this, I think I would bring in the extra vector data fields as optional function arguments instead of keyword arguments.  Of course feel free to push back here if you disagree!

scripts/plotting/global_latlon_vect_map.py
```diff
@@ -444,12 +467,12 @@ def global_latlon_vect_map(adfobj):
...
-                                                        [syear_baseline,eyear_baseline],lv,
-                                                        umseasons[s], vmseasons[s],
-                                                        uoseasons[s], voseasons[s],
-                                                        udseasons[s], vdseasons[s], obs, **vres)
+                                pf.plot_map_and_save(adfobj, plot_name, case_nickname, base_nickname,
+                                                    [syear_cases[case_idx],eyear_cases[case_idx]],
+                                                    [syear_baseline,eyear_baseline],
+                                                    umseasons[s], uoseasons[s],
+                                                    udseasons[s], upseasons[s],
+                                                    obs=obs, **vres)
```
nusbaume, line 475: If you do accept my suggestion on passing the main data fields as optional arguments, then this function call would look something like this instead:


```suggestion
                                pf.plot_map_and_save(adfobj, plot_name, case_nickname, base_nickname,
                                                    [syear_cases[case_idx],eyear_cases[case_idx]],
                                                    [syear_baseline,eyear_baseline],
                                                    umseasons[s], uoseasons[s],
                                                    udseasons[s], upseasons[s], 
                                                    vm=vmseasons[s], vo=voseasons[s],
                                                    vd=vdseasons[s], vp=vpseasons[s],
                                                    obs=obs, **vres)
```

Then in the `plot_map_and_save` function itself all of the `vx` variables would be `None` in the function argument list.

scripts/plotting/meridional_mean.py
```diff
@@ -98,161 +85,224 @@ def meridional_mean(adfobj):
...
+                #Check meridional log-p:
+                plot_name_log = plot_loc / f"{var}_{s}_Meridional_logp_Mean.{plot_type}"
+
+                # Check redo_plot. If set to True: remove old plot, if it already exists:
+                if (not redo_plot) and plot_name_log.is_file():
+                    logp_meridional_skip.append(plot_name_log)
+                    #Continue to next iteration:
+                    adfobj.add_website_data(plot_name_log, f"{var}_logp", case_name, season=s,
+                                            plot_type="Meridional", category=web_category)
+                    pass
```
nusbaume, line 117: I would get rid of the `pass` here, as it is literally doing nothing.

scripts/plotting/meridional_mean.py
```diff
@@ -98,161 +85,224 @@ def meridional_mean(adfobj):
...
+                #Check regular meridional
+                plot_name = plot_loc / f"{var}_{s}_Meridional_Mean.{plot_type}"
+                # Check redo_plot. If set to True: remove old plot, if it already exists:
+                if (not redo_plot) and plot_name.is_file():
+                    meridional_skip.append(plot_name)
+                    #Add already-existing plot to website (if enabled):
+                    adfobj.add_website_data(plot_name, var, case_name, season=s,
+                                                        plot_type="Meridional")
+
+                    continue
```
nusbaume, line 132: I would actually get rid of the `continue` here as well, as we are at the end of the loop at this point anyways (so the `continue` isn't really buying us anything).

scripts/plotting/meridional_mean.py
```diff
@@ -98,161 +85,224 @@ def meridional_mean(adfobj):
...
+            base_name = adfobj.data.ref_labels[var]
+
+        # Gather reference variable data
+        odata = adfobj.data.load_reference_regrid_da(base_name, var)
+
+        #Check if regridded file exists, if not skip zonal plot for this var
+        if odata is None:
+            dmsg = f"\t    WARNING: No regridded baseline file for {base_name} for variable `{var}`, zonal mean plotting skipped."
+            adfobj.debug_log(dmsg)
+            continue
```
nusbaume, line 185: I assume it is supposed to say `meridional` here?


```suggestion
        #Check if regridded file exists, if not skip meridional plot for this var
        if odata is None:
            dmsg = f"\t    WARNING: No regridded baseline file for {base_name} for variable `{var}`, meridional mean plotting skipped."
            adfobj.debug_log(dmsg)
            continue
```

scripts/plotting/meridional_mean.py
```diff
@@ -98,161 +85,224 @@ def meridional_mean(adfobj):
...
-                mclim_ds = utils.load_dataset(mclim_fils)
+            # load re-gridded model files:
+            mdata = adfobj.data.load_regrid_da(case_name, var)
 
-                # stop if data is invalid:
-                if (oclim_ds is None) or (mclim_ds is None):
-                    warnings.warn(f"invalid data, skipping meridional mean plot of {var}")
-                    continue
+            if mdata is None:
+                dmsg = f"\t    WARNING: No regridded test file for {case_name} for variable `{var}`, zonal mean plotting skipped."
```
nusbaume, line 210: Meridional?


```suggestion
                dmsg = f"\t    WARNING: No regridded test file for {case_name} for variable `{var}`, meridional mean plotting skipped."
```

scripts/plotting/meridional_mean.py
```diff
@@ -98,161 +85,224 @@ def meridional_mean(adfobj):
...
-                if validate_lat_lev['has_lev']:
-                    print(f"\t    INFO: {var} has lev dimension.")
-                    has_lev = True
-                else:
-                    has_lev = False
+            # determine whether it's 2D or 3D
+            # 3D triggers search for surface pressure
+            # check data dimensions:
+            #has_lat, has_lev = utils.zm_validate_dims(mdata)
+            lat_lev = utils.validate_dims(mdata, ['lat', 'lev'])
```
nusbaume, line 218: Remove commented-out line?


```suggestion
            lat_lev = utils.validate_dims(mdata, ['lat', 'lev'])
```

scripts/plotting/meridional_mean.py
```diff
@@ -98,161 +85,224 @@ def meridional_mean(adfobj):
...
+            else:
+                has_lev = False
 
-                #
-                # Seasonal Averages
-                # Note: xarray can do seasonal averaging, but depends on having time accessor,
-                # which these prototype climo files don't.
+            #Check to make sure each case has vertical levels if one of the cases does
+            if (has_lev) and (not has_lev_ref):
+                print(f"\t    WARNING: expecting lev boolean for both case: {has_lev} and ref: {has_lev_ref}")
```
nusbaume, line 230: I am not sure this warning would mean much to an inexperienced user.  Would it be better to say something like `WARNING:  test case has vertical levels dimension for '{var}', but reference case does not.`?

I have the same request for the other warning right below here as well, although in that case it would be the test case that is missing vertical levels.

scripts/plotting/meridional_mean.py
```diff
@@ -98,161 +85,224 @@ def meridional_mean(adfobj):
...
+            #Create new dictionaries:
+            mseasons = {}
+            oseasons = {}
+
+            #Loop over season dictionary:
+            for s in seasons:
+                vres["season"] = s
+
+                # time to make plot; here we'd probably loop over whatever plots we want for this variable
+                # I'll just call this one "Zonal_Mean"  ... would this work as a pattern [operation]_[AxesDescription] ?
```
nusbaume, line 249: I might just get rid of these two comments, as I feel at this point the ADF is fairly settled on plotting names and what plots we generally want, so I am worried these might just cause more confusion than clarity for the average user.

scripts/plotting/meridional_mean.py
```diff
@@ -98,161 +85,224 @@ def meridional_mean(adfobj):
...
+                #       so we could have made one script instead of two.
+                #       Merging would make overall timing better because looping twice will double I/O steps.
                 #
 
-                #Create new dictionaries:
-                mseasons = {}
-                oseasons = {}
+                # difference: each entry should be (lat, lon) or (plev, lat, lon)
+                # dseasons[s] = mseasons[s]    oseasons[s]
+                # difference will be calculated in plot_zonal_mean_and_save;
```
nusbaume, line 257: meridional?


```suggestion
                # difference will be calculated in plot_meridional_mean_and_save;
```

scripts/plotting/meridional_mean.py
```diff
@@ -98,161 +85,224 @@ def meridional_mean(adfobj):
...
+                                                    mseasons[s], oseasons[s], has_lev, log_p=False, obs=adfobj.compare_obs, **vres)
 
                     #Add plot to website (if enabled):
-                    adfobj.add_website_data(plot_name, var, case_name, season=s,
-                                            plot_type="Meridional")
+                    adfobj.add_website_data(plot_name, var, case_name, season=s, plot_type="Meridional")
+                #End if
+
+                #Create log-pressure plots as well (if applicable)
+                if (plot_name_log) and (plot_name_log not in logp_meridional_skip):
```
nusbaume, line 288: Total nit-pick, but I might get rid of the parentheses in the first part of the if-statement here:


```suggestion
                if plot_name_log and (plot_name_log not in logp_meridional_skip):
```

scripts/plotting/meridional_mean.py
```diff
@@ -98,161 +85,224 @@ def meridional_mean(adfobj):
...
-                                                mseasons[s], oseasons[s], has_lev, latbounds=slice(-5,5), obs=obs, **vres)
+                    pf.plot_meridional_mean_and_save(adfobj, plot_name, case_nickname, adfobj.data.ref_nickname,
+                                                    [syear_cases[case_idx],eyear_cases[case_idx]],
+                                                    [syear_baseline,eyear_baseline],
+                                                    mseasons[s], oseasons[s], has_lev, log_p=False, obs=adfobj.compare_obs, **vres)
 
                     #Add plot to website (if enabled):
-                    adfobj.add_website_data(plot_name, var, case_name, season=s,
-                                            plot_type="Meridional")
+                    adfobj.add_website_data(plot_name, var, case_name, season=s, plot_type="Meridional")
```
nusbaume, line 284: Just double-checking if this should have a `category=web_category` argument as well?

lib/plotting_utils.py
```diff
@@ -228,7 +246,7 @@ def get_central_longitude(*args):
 #  -- zonal & meridional mean code --
 #
 
-def _plot_line(axobject, xdata, ydata, color, **kwargs):
+'''def _plot_line(axobject, xdata, ydata, color, **kwargs):
```
nusbaume, line 249: Should we just remove this function instead of commenting it out?  It looks like a different `_plot_line` exists just below it anyways.

lib/plotting_utils.py
```diff
@@ -243,6 +261,65 @@ def _plot_line(axobject, xdata, ydata, color, **kwargs):
...
+        if color is not None:
+            axobject.plot(xdata, ydata, c=color, **plot_kwargs)
+        else:
+            axobject.plot(xdata, ydata, **plot_kwargs)
+
+    # Y label logic
+    if hasattr(ydata, "units"):
+        axobject.set_ylabel("[{units}]".format(units=getattr(ydata, "units")))
+    elif "units" in kwargs:
+        axobject.set_ylabel("[{units}]".format(kwargs["units"]))
```
nusbaume, line 321: Doesn't this `format` statement need a keyword argument?


```suggestion
        axobject.set_ylabel("[{units}]".format(units=kwargs["units"]))
```

lib/plotting_utils.py
```diff
@@ -253,9 +330,8 @@ def meridional_plot_line(ax, lon, data, color, **kwargs):
     #
     # annotate
     #
-    ax.set_xlabel("LONGITUDE")
     if hasattr(data, "units"):
-        ax.set_ylabel("{units}".format(units=getattr(data,"units")))
+        ax.set_ylabel("{units}".format(units=getattr(data,"units")),fontsize=6)
     elif "units" in kwargs:
         ax.set_ylabel("{units}".format(kwargs["units"]))
```
nusbaume, line 336: I believe this `format` statement needs a keyword argument as well:


```suggestion
        ax.set_ylabel("{units}".format(units=kwargs["units"]))
```

lib/plotting_utils.py
```diff
@@ -267,9 +343,8 @@ def zonal_plot_line(ax, lat, data, color, **kwargs):
     #
     # annotate
     #
-    ax.set_xlabel("LATITUDE")
     if hasattr(data, "units"):
-        ax.set_ylabel("{units}".format(units=getattr(data,"units")))
+        ax.set_ylabel("{units}".format(units=getattr(data,"units")),fontsize=6)
     elif "units" in kwargs:
         ax.set_ylabel("{units}".format(kwargs["units"]))
```
nusbaume, line 349: keyword argument needed for `format`?




```suggestion
        ax.set_ylabel("{units}".format(units=kwargs["units"]))
```

lib/plotting_utils.py
```diff
@@ -307,6 +382,311 @@ def meridional_plot_preslon(ax, lon, lev, data, **kwargs):
...
+#--------------------
+
+ncl_defaults = ["ncl_default"]
+
+def guess_ncl_url(cmap):
+    return f"https://www.ncl.ucar.edu/Document/Graphics/ColorTables/Files/{cmap}.rgb"
+
+
+def download_ncl_colormap(url, dest):
+    urlretrieve(url, dest)
```
nusbaume, line 396: I don't think this function header is buying us anything when we can just use `urlretrieve` directly.

lib/plotting_utils.py
```diff
@@ -307,6 +382,311 @@ def meridional_plot_preslon(ax, lon, lev, data, **kwargs):
...
+            else:
+                msg += f"\n\tFile will be downloaded and saved as {filename}"
+                download_ncl_colormap(fil, str(filename))
+        else:
+            is_url = False
+            filename = Path(fil)
+    elif isinstance(fil, Path):
+        filename = fil
+    else:
+        raise ValueError(f"\tERROR: what to do with type {type(fil)}")
```
nusbaume, line 419: I don't think we want to raise an error here, in keeping with the philosophy of the ADF only dying when it literally can't go on.  Instead I might just print a warning message and then just `return None` out of the function.

lib/plotting_utils.py
```diff
@@ -307,6 +382,311 @@ def meridional_plot_preslon(ax, lon, lev, data, **kwargs):
...
+        
+    # NCL's colormaps are not regularized enough to just use read_csv. 
+    # We have to determine how many lines to skip because it varies.
+    # NCL has some files that have "comments" at the end
+    # which will look like additional columnns
```
nusbaume, line 424: Typo here:


```suggestion
    # which will look like additional columns
```

lib/plotting_utils.py
```diff
@@ -307,6 +382,311 @@ def meridional_plot_preslon(ax, lon, lev, data, **kwargs):
...
+    # AND THE VALUES ARE INTEGERS 0-255.
+    with open(filename) as f:
+        table_exists = False
+        for count, line in enumerate(f):
+            line_str = line.strip() # remove leading/trailing whitespace
+            if (len(line_str) == 0) or (not line_str[0].isdigit()):
+                continue # empty line or non-data line 
+                # NOTE: also skips if the first value is negative (hlu_default) 
+            else:
+                if re.search(r'[^\s0-9-\.]', line_str): # any non number characters ASSUMED AT END
```
nusbaume, line 437: It would probably be slightly more performant to do something like:

```
digit_regex = re.compile(r'[^\s0-9-\.]')
```

Before the `enumerate` loop, and then replace `re.search` with `digit_regex.search(line_str)` here and below.

lib/plotting_utils.py
```diff
@@ -307,6 +382,311 @@ def meridional_plot_preslon(ax, lon, lev, data, **kwargs):
...
+                    # take the string up to the non-matching character
+                    line_vals = line_str[:re.search(r'[^\s0-9-\.]', line_str).start()-1].strip().split()
+                else:
+                    line_vals = line_str.split()
+                try: 
+                    row = [float(r) for r in line_vals]
+                except:
+                    msg += f"\n\tERROR reading RGB file {line_vals}"
```
nusbaume, line 445: I assume this is supposed to `except` when `float()` fails?  If so then I might specify the exception.  Also I would replace "ERROR" with "WARNING" in the actual message string since this failure won't kill the ADF:


```suggestion
                except ValueError:
                    msg += f"\n\tWARNING reading RGB file {line_vals}"
```

lib/plotting_utils.py
```diff
@@ -307,6 +382,311 @@ def meridional_plot_preslon(ax, lon, lev, data, **kwargs):
...
+    return f"https://www.ncl.ucar.edu/Document/Graphics/ColorTables/Files/{cmap}.rgb"
+
+
+def download_ncl_colormap(url, dest):
+    urlretrieve(url, dest)
+
+
+def read_ncl_colormap(adfobj, fil):
+    # determine if fil is a URL:
+    # if so, we have to download it
```
nusbaume, line 401: Convert to docstring?  If possible it would probably be good to add docstrings to the other functions below as well with a one sentence description of what they do.

lib/plotting_utils.py
```diff
@@ -307,6 +382,311 @@ def meridional_plot_preslon(ax, lon, lev, data, **kwargs):
...
+    return table
+
+
+def ncl_to_mpl(adfobj, nclmap, name):
+    msg = f"{script_name}: ncl_to_mpl()"
+    if nclmap.max() > 1:
+        try:
+            vals = nclmap / 255
+        except:
+            msg += f"\n\tERROR: could not divide by 255. {type(nclmap) = }"
```
nusbaume, line 461: Replace `ERROR` with `WARNING`?


```suggestion
            msg += f"\n\tWARNING: could not divide by 255. {type(nclmap) = }"
```

lib/plotting_utils.py
```diff
@@ -307,6 +382,311 @@ def meridional_plot_preslon(ax, lon, lev, data, **kwargs):
...
+            msg += f" {nclmap}"
+            adfobj.debug_log(msg)
+            return None
+    else:
+        msg += f"\n\t{name} seems to be 0-1"
+        vals = nclmap
+    assert vals.shape[1] == 3, 'vals.shape should be (N,3)'
```
nusbaume, line 468: I believe an `assert` call will kill the run if it fails.  We should instead make this an if-statement that prints a log message and then returns `None` if it fails instead.

lib/plotting_utils.py
```diff
@@ -307,6 +382,311 @@ def meridional_plot_preslon(ax, lon, lev, data, **kwargs):
...
+    if ncolors > 100:
+        my_cmap = mpl.colors.LinearSegmentedColormap.from_list(name, vals)
+        my_cmap_r = my_cmap.reversed()
+    else:
+        my_cmap = mpl.colors.ListedColormap(vals, name)
+        my_cmap_r = my_cmap.reversed()
+    # my_cmap, my_cmap_r from reversing a colormap
+    # ALLOW MPL TO KNOW ABOUT THE COLORMAP:
+    # mpl.colormaps.register(cmap=my_cmap)
+    # mpl.colormaps.register(cmap=my_cmap_r)
```
nusbaume, line 479: Remove commented-out code?

lib/plotting_utils.py
```diff
@@ -307,6 +382,311 @@ def meridional_plot_preslon(ax, lon, lev, data, **kwargs):
...
+def load_colormap(adfobj, cmap_name):
+    msg = f"{script_name}: load_colormap()"
+    if cmap_name in plt.colormaps():
+        adfobj.debug_log(msg)
+        return cmap_name
+    else:
+        msg += f"\n\t{cmap_name} not a standard Matplotlib colormap. Trying NCL..."
+        url = guess_ncl_url(cmap_name)
+        locfil = Path(".") / f"{cmap_name}.rgb"
+        data = read_ncl_colormap(adfobj, locfil,msg) if locfil.is_file() else read_ncl_colormap(adfobj, url)
```
nusbaume, line 503: I don't think `msg` can be in the `read_ncl_colormap` call here:


```suggestion
        data = read_ncl_colormap(adfobj, locfil) if locfil.is_file() else read_ncl_colormap(adfobj, url)
```

lib/plotting_utils.py
```diff
@@ -307,6 +382,311 @@ def meridional_plot_preslon(ax, lon, lev, data, **kwargs):
...
+        cm, cmr = ncl_to_mpl(adfobj, data, cmap_name)
+        if not cm:
+            msg += f"\n\tFailed to load {cmap_name}. Defaulting to 'coolwarm'."
+            adfobj.debug_log(msg)
+            return 'viridis'
```
nusbaume, line 508: Is `coolwarm` the same thing as `viridis`?  If not then we should probably change one or the other to match.  This holds for all the other references to `coolwarm` and `viridis` below as well.

lib/plotting_utils.py
```diff
@@ -307,6 +382,311 @@ def meridional_plot_preslon(ax, lon, lev, data, **kwargs):
...
+        if (cmap_case not in plt.colormaps()) and (cmap_case not in ncl_defaults):
+            msg += f"\n\tInvalid cmap '{cmap_case}' for {plotty}, defaulting to {default_cmap}"
+            cmap_case = default_cmap
+    
+    adfobj.debug_log(msg)
+
+    return cmap_case
+
+
+# Conour Plot Prep Functions
```
nusbaume, line 593: Typo:


```suggestion
# Contour Plot Prep Functions
```

lib/plotting_utils.py
```diff
@@ -307,6 +382,311 @@ def meridional_plot_preslon(ax, lon, lev, data, **kwargs):
...
+        msg = f"{script_name}: resolve_levels()"
+
+        # Map keys based on plot type
+        key_map = {
+            "diff": ("diff_contour_levels", "diff_contour_range", "diff_contour_linspace"),
+            "case": ("contour_levels", "contour_levels_range", "contour_levels_linspace"),
+        }
+        contour_levels, contour_range, contour_linspace = key_map.get(plotty, (None, None, None))
+
+        def process_entry(entry, kind, msg):
```
nusbaume, line 632: I might give this function a less generic name, like `create_contour_levels`.

lib/plotting_utils.py
```diff
@@ -307,6 +382,311 @@ def meridional_plot_preslon(ax, lon, lev, data, **kwargs):
...
+                        msg += f"\n\tLevels specified for {plotty}: numpy.arange."
+                        adfobj.debug_log(msg)
+                        return np.arange(*entry)
+                    elif kind == "linspace":
+                        msg += f"\n\tLevels specified for {plotty}: numpy.linspace."
+                        adfobj.debug_log(msg)
+                        return np.linspace(*entry)
+                    else:
+                        msg += f"\n\tLevels specified for {plotty} as list of 3 values, please add more values."
+                        msg += " Will get contrours from data range."
```
nusbaume, line 646: Typo:


```suggestion
                        msg += " Will get contours from data range."
```

lib/plotting_utils.py
```diff
@@ -307,6 +382,311 @@ def meridional_plot_preslon(ax, lon, lev, data, **kwargs):
...
+                        adfobj.debug_log(msg)
+                        return None
+                elif len(entry) < 3:
+                    msg += f"\n\tNot enough {kind} entries for {plotty} (<3) — ambiguous"
+                    adfobj.debug_log(msg)
```
nusbaume, line 651: I think (?) we need to `return None` here as well:


```suggestion
                    msg += f"\n\tNot enough {kind} entries for {plotty} (<3) — ambiguous."
                    msg += " Will get contours from data range."
                    adfobj.debug_log(msg)
                    return None
```

lib/plotting_utils.py
```diff
@@ -334,115 +714,205 @@ def prep_contour_plot(adata, bdata, diffdata, pctdata, **kwargs):
...
+    msg += f"\n\tFinal difference colormap: {cmap_diff}\n\n"
+
+    # CONTOUR LEVELS
+    #---------------
+    levels_diff = resolve_levels(adfobj, "diff", plot_type_dict, kwargs, polar_names)
+
+    msg += f"\n\tPre check difference LEVELS: {type(levels_diff)}\n\t\t{levels_diff}\n"
+    if levels_diff is None:
+        msg += f"\n\tSetting the difference levels from max/min"
+        absmaxdif = np.max(np.abs(diffdata))
```
nusbaume, line 834: Should this be `nanmax` here?


```suggestion
        absmaxdif = np.nanmax(np.abs(diffdata))
```

lib/plotting_utils.py
```diff
@@ -334,115 +714,205 @@ def prep_contour_plot(adata, bdata, diffdata, pctdata, **kwargs):
...
+        norm_diff = mpl.colors.Normalize(vmin=np.min(levels_diff), vmax=np.max(levels_diff))
+
     # Percent Difference options -- Check in kwargs for colormap and levels
-    if "pct_diff_colormap" in kwargs:
-        cmappct = kwargs["pct_diff_colormap"]
+    # COLOR MAP
+    #----------
+    # determine levels & color normalization:
+    #minval = np.min(diffdata)
+    #maxval = np.max(diffdata)
```
nusbaume, line 859: Remove commented-out code?

lib/plotting_utils.py
```diff
@@ -334,115 +714,205 @@ def prep_contour_plot(adata, bdata, diffdata, pctdata, **kwargs):
...
+
+        url = guess_ncl_url(cmap_pctdiff)
+        locfil = "." / f"{cmap_pctdiff}.rgb"
+        if locfil.is_file():
+            data = read_ncl_colormap(adfobj, locfil)
+        else:
+            data = read_ncl_colormap(adfobj, url)
+        cm, cmr = ncl_to_mpl(adfobj, data, cmap_pctdiff)
+        #ncl_colors[cm.name] = cm
+        #ncl_colors[cmr.name] = cmr
```
nusbaume, line 878: Could this code block possibly be replaced with `try_load_ncl_cmap`?

lib/plotting_utils.py
```diff
@@ -334,115 +714,205 @@ def prep_contour_plot(adata, bdata, diffdata, pctdata, **kwargs):
...
+            adfobj.debug_log(msg)
+        else:
+            cmap_pctdiff = cm
+
+    # CONTOUR LEVELS
+    #---------------
+    if "pct_diff_contour_levels" in plot_type_dict:
+        levels_pctdiff = plot_type_dict["pct_diff_contour_levels"]  # a list of explicit contour levels
+    elif "pct_diff_contour_range" in plot_type_dict:
+            assert len(plot_type_dict['pct_diff_contour_range']) == 3, "pct_diff_contour_range must have exactly three entries: min, max, step"
```
nusbaume, line 893: Instead of calling `assert` here (which will kill the ADF run), should we just print this message to the log (or screen) and then set a default range?

Also this code block looks a little over-indented?

lib/plotting_utils.py
```diff
@@ -457,23 +927,50 @@ def prep_contour_plot(adata, bdata, diffdata, pctdata, **kwargs):
...
+            'levels_sim': levels_sim,
+            'plot_log_p': plot_log_p,
+            'extend_sim': extend,
+            'extend_diff': extend_diff,
+            'extend_pctdiff': extend_pctdiff,
+            'units': units
             }
 
 
+def add_var_to_vres(adfobj, var, vres):
```
nusbaume, line 958: I might change this function slightly to make it clear what it is doing (as I was somewhat confused when seeing it in the plotting scripts):


```suggestion
def add_var_name_key_to_vres(adfobj, var, vres):
```

lib/plotting_functions.py
```diff
@@ -269,363 +194,154 @@ def make_polar_plot(wks, case_nickname,
...
+    maxval    = np.max([np.max(d1), np.max(d2)])
+    absmaxdif = np.max(np.abs(dif))
+    absmaxpct = np.max(np.abs(pct))
 
-    levs = np.unique(np.array(levels1))
-    levs_diff = np.unique(np.array(levelsdiff))
-    levs_pctdiff = np.unique(np.array(levelspctdiff))
+    means.extend([d1_region_mean,d2_region_mean, pct_region_mean, dif_region_mean])
+    mins.extend([d1_region_min,d2_region_min, pct_region_min, dif_region_min])
+    maxs.extend([d1_region_max,d2_region_max, pct_region_max, dif_region_max])
```
nusbaume, line 211: I might add a few spaces here to keep the formatting consistent:


```suggestion
    means.extend([d1_region_mean, d2_region_mean, pct_region_mean, dif_region_mean])
    mins.extend([d1_region_min, d2_region_min, pct_region_min, dif_region_min])
    maxs.extend([d1_region_max, d2_region_max, pct_region_max, dif_region_max])
```

lib/plotting_functions.py
```diff
@@ -269,363 +194,154 @@ def make_polar_plot(wks, case_nickname,
...
+                                            **cp_info['contourf_opt']))
+
+        # Set stats for title
+        stat_mean = f"Mean: {means[i]:5.2f}"
+        stat_max =  f"Max: {maxs[i]:5.2f}"
+        stat_min = f"Min: {mins[i]:5.2f}"
+        stats = f"{stat_mean}\n{stat_max}\n{stat_min}"
+        #axs[i].set_title(stats, loc='right', fontsize=8)
+        axs[i].text(-0.05, 0.0, stats, transform=axs[i].transAxes,fontsize=cbar_size)
+        #axs[i].text(0.8, 0.925, stats, transform=axs[i].transAxes,fontsize=cbar_size)
```
nusbaume, line 251: Remove commented-out code?


```suggestion
        axs[i].text(-0.05, 0.0, stats, transform=axs[i].transAxes,fontsize=cbar_size)
```

lib/plotting_functions.py
```diff
@@ -269,363 +194,154 @@ def make_polar_plot(wks, case_nickname,
...
-        img4 = ax4.contourf(lons, lats, dif_cyclic, colors="w", norm=dnorm)
-        ax4.text(0.4, 0.4, empty_message, transform=ax4.transAxes, bbox=props)
-    else:
-        img4 = ax4.contourf(lons, lats, dif_cyclic, cmap=cmapdiff, norm=dnorm, levels=levelsdiff)
-        
     #Set Main title for subplots:
-    st = fig.suptitle(wks.stem[:-5].replace("_"," - "), fontsize=18)
+    var_name = kwargs['var_name']
+    if "lev" in kwargs:
+        lev = kwargs["lev"]
```
nusbaume, line 258: I think this variable name might be confusing given the existence of `levs` and `levels` above.  Maybe rename to something like `vert_lev` instead?  Alternatively we could rename the variables above instead.

lib/plotting_functions.py
```diff
@@ -269,363 +194,154 @@ def make_polar_plot(wks, case_nickname,
...
+        var_name = f"{var_name} at {lev} hPa"
+    st = fig.suptitle(f"{var_name}: {kwargs['season']}", fontsize=14,
+                      fontfamily=["DejaVu Sans", "Liberation Sans", "sans-serif"])
     st.set_y(0.95)
 
     #Set plot titles
     case_title = "$\mathbf{Test}:$"+f"{case_nickname}\nyears: {case_climo_yrs[0]}-{case_climo_yrs[-1]}"
-    ax1.set_title(case_title, loc='left', fontsize=6) #fontsize=tiFontSize
+    axs[0].set_title(case_title, loc='left', fontsize=8) #fontsize=tiFontSize
```
nusbaume, line 266: Are we eventually going to use `tiFontSize` again?  If not then I might remove the commented-out code here and below.

lib/plotting_functions.py
```diff
@@ -681,37 +397,86 @@ def plot_map_and_save(wks, case_nickname, base_nickname,
...
+        mdl_mag_ma  = np.sqrt(umdlfld**2 + vmdlfld**2)
+        obs_mag_ma  = np.sqrt(uobsfld**2 + vobsfld**2)
+
+        #Convert vector magnitudes to xarray DataArrays:
+        mdl_mag  = xr.DataArray(mdl_mag_ma)
+        obs_mag  = xr.DataArray(obs_mag_ma)
+        sim_mags = [mdl_mag, obs_mag]
+        diff_mag = mdl_mag - obs_mag
+        pctdiff_mag = diff_mag / np.abs(obs_mag) * 100.0
+        #pctdiff_mag = diff_mag / np.abs((mdl_mag + obs_mag)/2) * 100.0
```
nusbaume, line 448: Remove commented-out code?

lib/plotting_functions.py
```diff
@@ -681,37 +397,86 @@ def plot_map_and_save(wks, case_nickname, base_nickname,
...
+        pctdiff_mag = pctdiff_mag.where(np.isfinite(pctdiff_mag), np.nan)
+        pctdiff_mag = pctdiff_mag.fillna(0.0)
+
+        wrap_fields = (umdlfld, uobsfld, udiffld, upctdiffld)
+
+        # get statistics (from non-wrapped)
+        fields = (mdl_mag, obs_mag, pctdiff_mag, diff_mag)
+        wgt = np.cos(np.radians(lat))
+        #area_avg = [utils.spatial_average(x, weights=wgt, spatial_dims=None) for x in fields]
+        #area_avg = [400 for x in fields]
```
nusbaume, line 458: Is there a reason we are not calculating the weights using the `spatial_average()` utility?  Otherwise it could be possible that vector plots and contour plots will have different spatial weightings.

lib/plotting_functions.py
```diff
@@ -728,13 +494,24 @@ def plot_map_and_save(wks, case_nickname, base_nickname,
     fig = plt.figure(figsize=(14,10))
 
     # LAYOUT WITH GRIDSPEC
-    gs = mpl.gridspec.GridSpec(3, 6, wspace=2.0,hspace=0.0) # 2 rows, 4 columns, but each map will take up 2 columns
+    wspace = 1.66
+    hspace = -0.25
+    if vector:
+        hspace = 0
+    else:
+        hspace = -0.25
```
nusbaume, line 502: I think this can be simplified:


```suggestion
    if vector:
        hspace = 0
    else:
        hspace = -0.25
```

lib/plotting_functions.py
```diff
@@ -747,29 +524,77 @@ def plot_map_and_save(wks, case_nickname, base_nickname,
...
-            levels = cp_info['levels1']
-            cmap = cp_info['cmap1']
-            norm = cp_info['norm1']
+    for i, a in enumerate(wrap_fields):
+        if i == diff_idx:
+            levels = cp_info['levels_diff']
+            cmap = cp_info['cmap_diff']
+            extend = cp_info["extend_diff"]
+            if vector:
+                # Get difference limits, in order to plot the correct range:
```
nusbaume, line 539: Is there a reason we can't use `norm = cp_info['norm_diff']` like we do for the regular contour plot?

lib/plotting_functions.py
```diff
@@ -747,29 +524,77 @@ def plot_map_and_save(wks, case_nickname, base_nickname,
...
+                            transform=ccrs.PlateCarree(), cmap='Reds')
+                    ax[i].quiver(lons[skip], lats[skip], uobsfld[skip], vobsfld[skip], obs_mag.values[skip],
+                            transform=ccrs.PlateCarree(), cmap='Reds')
+            else:
+                img.append(ax[i].contourf(lons, lats, a, levels=levels, cmap=cmap, norm=norm,
+                                      transform=ccrs.PlateCarree(), transform_first=True, extend=extend,
+                                      **cp_info['contourf_opt']))
         #End if
-        ax[i].set_title("AVG: {0:.3f}".format(area_avg[i]), loc='right', fontsize=11)
+        #ax[i].set_title("AVG: {0:.3f}".format(area_avg[i]), loc='right', fontsize=11)
```
nusbaume, line 597: Remove commented-out code?

lib/plotting_functions.py
```diff
@@ -980,6 +817,34 @@ def plot_zonal_mean_and_save(wks, case_nickname, base_nickname,
...
 
+    azm = utils.zonal_mean_xr(adata)
+    bzm = utils.zonal_mean_xr(bdata)
+    diff = azm - bzm
+        
+    # calculate the percent change
+    pct = (azm - bzm) / np.abs(bzm) * 100.0
+    #check if pct has NaN's or Inf values and if so set them to 0 to prevent plotting errors
+    pct = pct.where(np.isfinite(pct), np.nan)
+    pct = pct.fillna(0.0)
```
nusbaume, line 828: Do you actually need the `fillna` line here?


```suggestion
    pct = pct.where(np.isfinite(pct), 0.0)
```

lib/plotting_functions.py
```diff
@@ -990,115 +855,188 @@ def plot_zonal_mean_and_save(wks, case_nickname, base_nickname,
...
-        
-        # calculate the percent change
-        pct = (azm - bzm) / np.abs(bzm) * 100.0
-        #check if pct has NaN's or Inf values and if so set them to 0 to prevent plotting errors
-        pct = pct.where(np.isfinite(pct), np.nan)
-        pct = pct.fillna(0.0)
-        
-        fig, ax = plt.subplots(nrows=3)
-        ax = [ax[0],ax[1],ax[2]]
+        fig, ax = plt.subplots(nrows=3)#figsize=(6,8),
```
nusbaume, line 975: Remove commented-out code?

lib/plotting_functions.py
```diff
@@ -990,115 +855,188 @@ def plot_zonal_mean_and_save(wks, case_nickname, base_nickname,
...
+                   borderaxespad=0.0,fontsize=6,frameon=False,labelspacing=0.3)
         
-        zonal_plot(adata['lat'], pct, ax=ax[2], color="k")
-        ax[2].set_title("Test % Diff Baseline", loc='left', fontsize=10,fontweight="bold")
+        use_cmap = kwargs.get("colormap_2d", False)
+        plot_kwargs = kwargs.copy()
+        plot_kwargs["colormap_2d"] = use_cmap
+        diff_kwargs = plot_kwargs.copy()
+        diff_kwargs["type"] = "diff"
+        diff_kwargs.pop("norm", None)
```
nusbaume, line 995: Instead of making multiple copies, why not just make one copy that is then modified for the `diff` and `pctdiff` plots?  The only reason you would need multiple copies if you needed to save `diff_kwargs` after creating the zonal plot for the difference field.

lib/plotting_functions.py
```diff
@@ -1239,92 +1199,190 @@ def plot_meridional_mean_and_save(wks, case_nickname, base_nickname,
...
     if has_lev:
-        # generate dictionary of contour plot settings:
-        cp_info = plot_utils.prep_contour_plot(adata, bdata, diff, pct, **kwargs)
 
-        # generate plot objects:
-        fig, ax = plt.subplots(figsize=(10,10),nrows=4, constrained_layout=True, sharex=True, sharey=True,**cp_info['subplots_opt'])
-        levs = np.unique(np.array(cp_info['levels1']))
-        levs_diff = np.unique(np.array(cp_info['levelsdiff']))
-        levs_pctdiff = np.unique(np.array(cp_info['levelspctdiff']))
+        # Generate zonal plot:
```
nusbaume, line 1204: Meridional plot?


```suggestion
        # Generate meridional plot:
```

lib/plotting_functions.py
```diff
@@ -1239,92 +1199,190 @@ def plot_meridional_mean_and_save(wks, case_nickname, base_nickname,
...
+        fig, ax = plt.subplots(figsize=(8,10),nrows=4, constrained_layout=True,
+                               sharey=True, **cp_info['subplots_opt'])
 
+        levs = np.unique(np.array(levels_sim))
```
nusbaume, line 1208: Just noting again that naming this variable to something like `contour_levs` might help make it obvious that this is not related to vertical levels.

lib/plotting_functions.py
```diff
@@ -1239,92 +1199,190 @@ def plot_meridional_mean_and_save(wks, case_nickname, base_nickname,
...
-                borderaxespad=0.0,fontsize=6,frameon=False)
+        fig.legend(handles=[line,line2],bbox_to_anchor=(0.125, 0.84, 1.05, .102),loc="upper left",
+                   borderaxespad=0.0,fontsize=6,frameon=False,labelspacing=0.3)
+
+        use_cmap = kwargs.get("colormap_2d", False)
+        plot_kwargs = kwargs.copy()
+        plot_kwargs["colormap_2d"] = use_cmap
+        diff_kwargs = plot_kwargs.copy()
+        diff_kwargs["type"] = "diff"
+        diff_kwargs.pop("norm", None)
```
nusbaume, line 1338: Can we get away with only one copy of `kwargs` here as well?

nusbaume (changes requested):
Thanks for all of the code cleanup and plot enhancements @justin-richling!  I realize it looks like I have a bazillion comments, but the vast majority should hopefully be easy or trivial to implement.  Of course if you do have any questions or concerns with anything I wrote just let me know!

lib/plotting_functions.py
```diff
@@ -269,363 +194,154 @@ def make_polar_plot(wks, case_nickname,
...
 
     # __COLORBARS__
-    cb_mean_ax = inset_axes(ax2,
-                    width="5%",  # width = 5% of parent_bbox width
+    cb_mean_ax = inset_axes(axs[1],
+                    width="5%",
                     height="90%",  # height : 90%
                     loc='lower left',
                     bbox_to_anchor=(1.05, 0.05, 1, 1),
                     bbox_transform=ax2.transAxes,
```
brianpm, line 302: I think this `ax2` needs to be pointed at an `axs` element

lib/plotting_functions.py
```diff
@@ -269,363 +194,154 @@ def make_polar_plot(wks, case_nickname,
...
+    cbar.ax.tick_params(labelsize=cbar_size)
     
-    cb_pct_ax = inset_axes(ax3,
-                    width="5%",  # width = 5% of parent_bbox width
+    cb_pct_ax = inset_axes(axs[2],
+                    width="5%",
                     height="90%",  # height : 90%
                     loc='lower left',
                     bbox_to_anchor=(1.05, 0.05, 1, 1),
                     bbox_transform=ax3.transAxes,
```
brianpm, line 314: `ax3` should be in `axs`

lib/plotting_functions.py
```diff
@@ -269,363 +194,154 @@ def make_polar_plot(wks, case_nickname,
...
+                    )
 
-    cb_diff_ax = inset_axes(ax4,
-                    width="5%",  # width = 5% of parent_bbox width
+    cb_diff_ax = inset_axes(axs[3],
+                    width="5%",
                     height="90%",  # height : 90%
                     loc='lower left',
                     bbox_to_anchor=(1.05, 0.05, 1, 1),
                     bbox_transform=ax4.transAxes,
```
brianpm, line 323: `ax4` should be in `axs`

lib/plotting_functions.py
```diff
@@ -269,363 +194,154 @@ def make_polar_plot(wks, case_nickname,
...
+        # for the map. We can pan/zoom as much as we like - the boundary will be
+        # permanently circular.
+        theta = np.linspace(0, 2*np.pi, 100)
+        center, radius = [0.5, 0.5], 0.5
+        verts = np.vstack([np.sin(theta), np.cos(theta)]).T
+        circle = mpl.path.Path(verts * radius + center)
+        a.set_boundary(circle, transform=a.transAxes)
+        a.gridlines(draw_labels=False, crs=ccrs.PlateCarree(), 
+                    lw=1, color="gray",y_inline=True, 
+                    xlocs=range(-180,180,90), ylocs=range(0,90,10))
```
brianpm, line 294: Is `range(0,90,10)`going to work for southern hemisphere?

lib/plotting_functions.py
```diff
@@ -820,37 +651,42 @@ def plot_map_and_save(wks, case_nickname, base_nickname,
...
-    cb_mean_ax = inset_axes(ax2,
-                    width="5%",  # width = 5% of parent_bbox width
-                    height="100%",  # height : 100%
+    cb_mean_ax = inset_axes(ax[1],
+                    width="5%",
+                    height="100%",
                     loc='lower left',
-                    bbox_to_anchor=(1.05, 0, 1, 1),
+                    bbox_to_anchor=(1.02, 0, 1, 1),
                     bbox_transform=ax2.transAxes,
```
brianpm, line 659: Should `ax2` be `axs[...]`?

lib/plotting_functions.py
```diff
@@ -820,37 +651,42 @@ def plot_map_and_save(wks, case_nickname, base_nickname,
...
+    cbar.ax.set_title(units, fontsize=cbar_size, pad=cbar_labelpad, loc='left')
+    cbar.ax.tick_params(labelsize=cbar_size)
+
+    if not vector:
+        cb_pct_ax = inset_axes(ax[pctdiff_idx],
+                        width="5%",
+                        height="100%",
+                        loc='lower left',
+                        bbox_to_anchor=(1.02, 0, 1, 1),
+                        bbox_transform=ax3.transAxes,
```
brianpm, line 672: `ax3` maybe `axs[...]`

lib/plotting_utils.py
```diff
@@ -334,115 +714,205 @@ def prep_contour_plot(adata, bdata, diffdata, pctdata, **kwargs):
...
-        levelspctdiff = kwargs["pct_diff_contour_levels"]  # a list of explicit contour levels
-    elif "pct_diff_contour_range" in kwargs:
-            assert len(kwargs['pct_diff_contour_range']) == 3, "pct_diff_contour_range must have exactly three entries: min, max, step"
-            levelspctdiff = np.arange(*kwargs['pct_diff_contour_range'])
+    if cmap_pctdiff not in plt.colormaps():
+        msg += f"\n\tPercent Difference: {cmap_pctdiff} is not a matplotlib standard color map."
+        msg += f" Trying if this an NCL color map"
+
+        url = guess_ncl_url(cmap_pctdiff)
+        locfil = "." / f"{cmap_pctdiff}.rgb"
```
brianpm, line 871: I think you have to use some pathlib thing here, probably: `locfil = Path.cwd() / f"{cmap_pctdiff}.rgb"`

lib/plotting_functions.py
```diff
@@ -681,37 +397,86 @@ def plot_map_and_save(wks, case_nickname, base_nickname,
...
+
+        #Convert vector magnitudes to xarray DataArrays:
+        mdl_mag  = xr.DataArray(mdl_mag_ma)
+        obs_mag  = xr.DataArray(obs_mag_ma)
+        sim_mags = [mdl_mag, obs_mag]
+        diff_mag = mdl_mag - obs_mag
+        pctdiff_mag = diff_mag / np.abs(obs_mag) * 100.0
+        #pctdiff_mag = diff_mag / np.abs((mdl_mag + obs_mag)/2) * 100.0
+        pctdiff_mag = pctdiff_mag.where(np.isfinite(pctdiff_mag), np.nan)
+        pctdiff_mag = pctdiff_mag.fillna(0.0)
```
brianpm, line 450: Is this what we actually want? Divide by zero is replaced by `np.nan`, but then get set to `0.0`. Won't that mean that any `nan` value ends up looking like a valid result in the plot?

lib/plotting_functions.py
```diff
@@ -747,29 +524,77 @@ def plot_map_and_save(wks, case_nickname, base_nickname,
...
+                    #             transform=ccrs.PlateCarree())"""
+                if i in [0,1]:
+                    img.append(ax[i].contourf(lons, lats, sim_mags[i], cmap='Greys',
+                                              transform=ccrs.PlateCarree(),
+                                              transform_first=True, extend=extend
+                        ))
+                    ax[i].quiver(lons[skip], lats[skip], umdlfld[skip], vmdlfld[skip], mdl_mag.values[skip],
+                            transform=ccrs.PlateCarree(), cmap='Reds')
+                    ax[i].quiver(lons[skip], lats[skip], uobsfld[skip], vobsfld[skip], obs_mag.values[skip],
+                            transform=ccrs.PlateCarree(), cmap='Reds')
```
brianpm, line 591: Both quiver calls are unconditional inside the if i in [0,1] block — so for both i=0 (model panel) and i=1 (baseline panel), it draws model vectors first and then baseline vectors on top. Both panels end up with overlaid arrows.

It seems like the intent is  i=0 → model only, i=1 → baseline only. The fix should be:
```
  if i == 0:
      ax[i].quiver(..., umdlfld[skip], vmdlfld[skip], mdl_mag.values[skip], ...)
  elif i == 1:
      ax[i].quiver(..., uobsfld[skip], vobsfld[skip], obs_mag.values[skip], ...)
```

lib/adf_variable_defaults.yaml
```diff
@@ -153,8 +153,6 @@ AODDUST:
   scale_factor: 1
   add_offset: 0
   new_unit: ""
-  pct_diff_contour_levels: [-100,-75,-50,-40,-30,-20,-10,-8,-6,-4,-2,0,2,4,6,8,10,20,30,40,50,75,100]
```
brianpm, line 156: Before this PR, `pct_diff_colormap` and `pct_diff_contour_levels` were at the top level of a variable's YAML entry and were found directly in kwargs. 

Now:
  - If the user has a `global_latlon_map` (or similar) sub-block, they must put `pct_diff_colormap` inside that block
 which means that the previous behavior of top-level placement is silently ignored.
  - If `plot_type` is not passed into the function at all, `plot_type_dict = {}` and there is no way for the user to override hard-coded values that are given in `lib/plotting_utils.py`:
  - Line 863: `cmap_pctdiff = "PuOr_r"`
  - Line 882: `cmap_pctdiff = 'PuOr_r'`
  - Line 896: `levels_pctdiff = [-100,-75,-50,-40,-30,-20,-10,-8,-6,-4,-2,0,2,4,6,8,10,20,30,40,50,75,100]` 

So the PR description's claim that users can still supply these args is only partially true. The configurability is preserved when nested under a plot-type block but silently lost in the flat/top-level case that was standard before. 

This probably doesn't matter much, but it complicates customizing plots from the YAML file. Needs documentation at minimum.

lib/plotting_utils.py
```diff
@@ -243,6 +261,65 @@ def _plot_line(axobject, xdata, ydata, color, **kwargs):
...
+    """Create a generic line plot and check for some ways to annotate."""
+
+    use_cmap = kwargs.get("colormap_2d", False)
+    cmap = kwargs.get("cmap", "viridis")
+    norm = kwargs.get("norm", None)
+    vmin = kwargs.get("vmin", None)
+    vmax = kwargs.get("vmax", None)
+    if use_cmap:
+        if kwargs["type"] == "pctdiff":
+            cmap = kwargs.get("pct_diff_colormap", "PuOr")
```
brianpm, line 278: The hard-coded yaml default was `PuOr_r` right? Should that be used here?

brianpm (commented):
I left a few comments where I saw some inconsistent changes. Looks good overall. My first thought is that we could probably reduce total code, but no need for now.

lib/plotting_functions.py
justin-richling: Yeah, this is very confusing as now we are misrepresenting the true values, which is not what we want.

Would something like this be more accurate: 

```python
pctdiff_mag = xr.where(obs_mag != 0,
                                  (mdl_mag - obs_mag) / np.abs(obs_mag) * 100,
                                   np.nan,
                                   )
```

That way we also catch anywhere obs/baseline is zero, or is this already handled with the `np.isfinite`?

I'm hoping catching `obs_mag` when zero will fix the times where the pct values are excessively large that I see occassionally.

lib/adf_variable_defaults.yaml
justin-richling: This is a great observation @brianpm, thanks for pointing it out!

lib/plotting_utils.py
justin-richling: Ah, good catch, thanks.

lib/plotting_functions.py
brianpm: I think isfinite does it. I think it works as:

`pctdiff_mag = pctdiff_mag.where(np.isfinite(pctdiff_mag))`

just get rid of the fillna line.

nusbaume (commented):
Review generated by Claude, following the repository's `AGENTS.md` (on `main`).

## Summary

This PR reworks contour/colormap resolution so levels and colormaps can vary by plot type, hemisphere and pressure level, merges the vector map into `plot_map_and_save`, adds NCL colormap support, and tidies units handling. The direction is good and the unit fixes (`adf_dataset.py` `raw_units`, `keep_attrs=True` in `zonal_mean_xr`) are clean. However, there are a few correctness problems that will show up in a default ADF run, so I don't think it's ready to merge as-is.

Per `AGENTS.md` §3, note that **CI covers none of this**: no file in this PR is in the pylint `testable_files` set, and the pytest suite only touches `adf_base`/`adf_config`. A green checkmark here means very little — everything below was found by reading and by executing the new helpers against the shipped `adf_variable_defaults.yaml` in the pinned `adf_v1.0.0` environment.

Also flagging up front: the PR currently reports `mergeable: false` against `main`.

---

## Blocking

**1. Vector-component variables get the vector's name as the plot title on every scalar plot — `lib/plotting_utils.py:958-973` (`add_var_to_vres`)**

`add_var_to_vres` checks `vector_name` first and unconditionally overrides `var_name` with it, even when the caller is making an ordinary scalar plot. Running it against the shipped defaults:

```
U                    -> figure title uses var_name = 'Wind'
V                    -> figure title uses var_name = 'Wind'
TAUX                 -> figure title uses var_name = 'Surface Wind Stress'
TAUY                 -> figure title uses var_name = 'Surface Wind Stress'
PRECT                -> figure title uses var_name = 'PRECT'
```

`var_name` feeds `fig.suptitle(f"{var_name}: {season}")` (`plotting_functions.py:256-260`, `606-610`, `952`, `979`, `1295`, `1322`). `U` and `V` are in the default `diag_var_list`, and `global_latlon_map`, `zonal_mean`, `meridional_mean` and `polar_map` all call `add_var_to_vres` — so in a stock run every scalar U and V plot is titled "Wind". The vector-name branch should only be taken when the caller is actually plotting the vector (e.g. gated on the `vector` flag that `global_latlon_vect_map.py` already sets).

**2. Explicit 3-entry `contour_levels` lists are silently discarded — `lib/plotting_utils.py:632-648` (`resolve_levels` / `process_entry`)**

For `kind == "levels"`, a list of exactly three entries falls into the `else` branch, logs "please add more values", and returns `None` — so the plot silently falls back to data-range levels. On `main`, `contour_levels` was used verbatim regardless of length (`plotting_utils.py:357-361` at the merge base). Any user whose variable-defaults file has e.g. `contour_levels: [-1, 0, 1]` gets different plots with no error and nothing on stdout — the `AGENTS.md` §4.3 backwards-compatibility rule. No variable in the two shipped defaults files has a 3-entry explicit list, so this only bites user-supplied files, but that's exactly where it's hardest to notice.

**3. NCL colormap parser silently corrupts values and crashes on plausible inputs — `lib/plotting_utils.py:399-452` (`read_ncl_colormap`)**

The trailing-comment strip at line 441-443 uses `line_str[:match.start()-1]`. The `-1` drops a real character whenever there's no space before the comment marker. Executed against the PR code:

```
"255 0 0 ; red"   -> [255.0, 0.0, 0.0]    correct
"255 0 10# red"   -> [255.0, 0.0, 1.0]    blue channel 10 silently becomes 1
```

Same function, other inputs:
- comment-only / header-only file → `UnboundLocalError: cannot access local variable 'table'`
- single data row → `table` is 1-D, then `ncl_to_mpl` raises `IndexError: tuple index out of range`
- `except:` at line 444 is bare (`AGENTS.md` §6.3) and leaves `row` bound to the *previous* row when a line fails to parse

The silent-truncation case is the one that matters most (`AGENTS.md` §5.2): wrong colors, no warning.

**4. Runtime colormap download — `lib/plotting_utils.py:391-412`**

`read_ncl_colormap` does `filename = Path.cwd() / fil.split("/")[-1]` and `urlretrieve` into it. Verified on a login node: calling `try_load_ncl_cmap(adf, "ncl_default")` writes `ncl_default.rgb` into the **current working directory** — i.e. wherever the user launched `./run_adf_diag`, usually the ADF clone itself. That's what the new `*.rgb` line in `.gitignore` is papering over. Three problems:

- ADF runs in batch on Derecho/Casper compute nodes, which generally have no outbound internet. The download will fail there, and since nothing is cached on failure it retries for every variable/season that asks for the colormap.
- `urlretrieve` is called with no timeout, so each of those retries can block rather than fail fast. Only `urllib.error.HTTPError` is caught explicitly.
- Writing generated files into the source tree (or any CWD) is surprising and breaks read-only/shared installs.

Since `ncl_defaults = ["ncl_default"]` is a single-entry list, this is one small file — vendoring it under `lib/` (or a configurable cache dir, defaulting somewhere writable) would remove the network dependency entirely.

---

## Non-blocking

- **Shared mutation of `variable_defaults`.** `AdfObs.variable_defaults` returns `copy.copy(...)` — a *shallow* copy, so the per-variable sub-dicts are shared. `vres = adfobj.variable_defaults.get(var, {})` aliases the master entry, and `add_var_to_vres` plus `vres["plot_type"]`/`vres["vector"]`/`vres["umdlfld_nowrap"] = ...` write straight into it. After `global_latlon_vect_map` runs, the master entry for `U` permanently carries `vector: True` and the season's `umdlfld_nowrap`/`vobsfld_nowrap` DataArrays, which then flow as `**vres` into later scripts. Each script does overwrite `plot_type` and `var_name`, so I did not find a visible failure from this today — but it's a live footgun and holds references to full fields. A `copy.deepcopy` of `vres` (or having `add_var_to_vres` return a new dict) closes it.
- **Dead code, one piece of it broken.** `load_colormap` (494) and `choose_colormap_type` (485) are never called. `load_colormap:503` calls `read_ncl_colormap(adfobj, locfil, msg)` — three arguments to a two-parameter function; confirmed `TypeError: read_ncl_colormap() takes 2 positional arguments but 3 were given` as soon as a local `.rgb` exists. It also says "Defaulting to 'coolwarm'" while returning `'viridis'`, and unpacks `cm, cmr = ncl_to_mpl(...)` when `ncl_to_mpl` can return a bare `None`. Either delete these two or fix and wire them up.
- **`try_load_ncl_cmap` (513-538).** Line 536 is `msg + "Something went wrong..."` — an expression statement, so that message never reaches the debug log (`msg +=` intended). If the inner `urllib.error.HTTPError` branch fires, `data` is never assigned and the following `isinstance(data, ...)` raises `UnboundLocalError`, which the outer bare `except Exception` then swallows — it lands on 'viridis' by accident rather than by design.
- **NCL support is narrower than the PR description suggests.** `ncl_defaults = ["ncl_default"]`, and `get_cmap` only attempts an NCL lookup when the name is exactly `"ncl_default"`. Any other NCL table name (`BlueRed`, `amwg`, …) falls through the final validity check and is replaced by the default colormap. Worth either generalizing or saying so in the docs.
- **`meridional_mean.py:181,183,210`** — new warning text says "zonal mean plotting skipped" in the meridional script. Copy-paste from `zonal_mean.py`; user-visible.
- **Stale module docstring, `lib/plotting_functions.py:1-30`.** It still documents `plot_map_vect_and_save`, which this PR deletes, and still shows the old signatures for `make_polar_plot` / `plot_map_and_save` / `plot_zonal_mean_and_save` / `plot_meridional_mean_and_save` without the new leading `adfobj` argument. `AGENTS.md` §7.6.
- **Docstrings on new helpers (`AGENTS.md` §6.2).** 7 of the 11 new functions in `plotting_utils.py` have no docstring at all (`guess_ncl_url`, `download_ncl_colormap`, `read_ncl_colormap`, `ncl_to_mpl`, `choose_colormap_type`, `load_colormap`, `add_var_to_vres`); the other 4 are one-liners with no `Parameters`/`Returns` sections. None were added to the module docstring's `Functions` list. `add_var_to_vres` and `resolve_levels` in particular have non-obvious contracts that need documenting.
- **New variable-defaults keys are undocumented.** The header comment block in `adf_variable_defaults.yaml` isn't updated for `nickname`, `colormap_2d`, `contour_levels_linspace`, or the new nested per-plot-type blocks (`polar_map:`, `global_latlon_map:`, `zonal_mean:`) with their `nh`/`sh`/level sub-keys — the headline feature of this PR. `AGENTS.md` §4.3 asks for new fields to be documented there. Right now those forms exist only in the PR description; the shipped file uses `nickname` once and none of the rest, so nothing exercises them.
- **`black` (`AGENTS.md` §6.1).** Intersecting `black --check --diff` hunks with the lines this PR adds or changes gives roughly 1,333 new/changed lines that black would reformat, concentrated in `plotting_functions.py` (~640), `plotting_utils.py` (~408) and `meridional_mean.py` (~189). `resolve_levels` (620) is also indented as if it were a method — its whole body sits at 8 spaces.
- **`env/npl-2024a_environment.yaml`** — a new 649-line fully-pinned environment, unrelated to plotting enhancements, with no README mention. Intentional, or a leftover? It's a real maintenance surface (`AGENTS.md` §4.5).

## Optional / follow-up

- Removing `pf.plot_map_vect_and_save` is right for the codebase, but it's a public helper that user-maintained scripts may call. Worth a line in the release notes.
- The PR adds 8 new `"$\mathbf{...}$"` string literals; these are not raw strings and produce `SyntaxWarning: invalid escape sequence '\m'` on import under Python 3.12+. Pre-existing pattern (24 occurrences on `main`), but new code could use `r"..."`.
- `cm.get_cmap` at `plotting_utils.py:795` is deprecated. I checked: it still works in the pinned matplotlib 3.9.4 (and 3.10), so this is not urgent — but since the block was rewritten anyway, `plt.get_cmap`/`mpl.colormaps[...]` would be cheap to adopt.
- The vector percent-difference is computed for every season and level (`global_latlon_vect_map.py`) and then discarded at `plotting_functions.py:527` (`if vector: #ignore percent diff for now`). Skipping the computation would save the work and make the intent clearer.
- `read_ncl_colormap:414` sets `is_url = False`, which is never used.

## Not verified

- **The plots themselves.** I did not run the ADF against real CAM output, so the visual claims in the PR description (colorbar fit, extend behaviour, zonal/meridional styling, `colormap_2d`) are unverified — as are the new contour-level and colormap choices as *science*, which need a CAM/AMWG reviewer's eye rather than mine.
- **Whether Derecho/Casper compute nodes can reach `ncl.ucar.edu`.** The download succeeded from a login node; the compute-node behaviour behind blocking finding 4 is my expectation, not something I tested.
- **`pytest lib/test/unit_tests`** — pytest isn't installed in the local `adf_v1.0.0` env, so I couldn't run it. No file in this PR is in the suite's coverage, so I'd expect it to pass unchanged. YAML parses cleanly for all files (`check-yaml` equivalent), and every changed `.py` file compiles.

PR #457, opened 2026-05-30
Add emission and deposition tables to the budget

Behrooz-Roozitalab:
Updates to the budget code including:
-  including 'tropospheric' or global as an option in the yaml file.
- including an option to create and use annual mean files instead of using the raw data.

Also, two tables are added to this version (could be called in YAML file) to calculate the emissions and depositions for all the variables in the files.

### DIFF
--- a/scripts/analysis/aerosol_gas_tables.py
+++ b/scripts/analysis/aerosol_gas_tables.py
@@ -1,7 +1,7 @@
 import numpy as np
 import xarray as xr
 from pathlib import Path
-
+import subprocess
 from datetime import datetime
 import numpy as np
 import itertools
@@ -10,7 +10,8 @@ import pandas as pd
 # Import necessary ADF modules:
 from adf_base import AdfError
 
-def aerosol_gas_tables(adfobj, trop_val=None, **kwargs):
+def aerosol_gas_tables(adfobj, trop_val=None, Tropospheric=None ,Climate=None , **kwargs):
+
     '''
     Calculate aerosol and gaseous budget tables
 
@@ -103,6 +104,15 @@ def aerosol_gas_tables(adfobj, trop_val=None, **kwargs):
             * Added DMS to gases list - reported as DMS not S
             * Automatic addition of gaseous compounds even when not defined in the default list, 
             *   based on Carbon MW (12). It still needs ADF modification to read a list from yaml file. 
+
+        Behrooz Roozitalab, 18 Apr, 2026
+        - fixed:
+            * the option added in YAML file to choose between tropospheric or global budgets. The default is tropospheric.
+    	    * The CESM model's 'AREA' variable is not correct in some versions. It was causing issues for cases when it was available. It is now completely removed.
+            * To increase the speed, we added an option "Climate". When True, we first make the annual mean files using nco, and then calculate the budget.
+            *                        This is not recommended for 1-2 year periods, but it works well with climate simulations that run for 20 years. 
+            *                        The default is False. 
+
     '''
 
     #Notify user that script has started:
@@ -113,6 +123,13 @@ def aerosol_gas_tables(adfobj, trop_val=None, **kwargs):
     if trop_val is None:
         trop_val = 'tropopause'
 
+    # Check which type of tables to be created, default to 'True'
+    if Tropospheric is None:
+        Tropospheric = True
+
+    # Check which type of tables to be created, default to 'False'
+    if Climate is None:
+        Climate = False
     # Inputs
     #-------
     # Variable defaults info
@@ -136,7 +153,7 @@ def aerosol_gas_tables(adfobj, trop_val=None, **kwargs):
     # if True, calculate only Tropospheric values
     # if False, all layers
     # tropopause is defiend as either directly or indirectly. Look for tropopause to see the definition
-    Tropospheric = bres['Tropospheric']
+    #Tropospheric = bres['Tropospheric']
 
     ### NOT WORKING FOR NOW
     # To calculate the budgets only for a region
@@ -152,8 +169,11 @@ def aerosol_gas_tables(adfobj, trop_val=None, **kwargs):
     # automatic generation of MW
     for var in VARIABLES:
         if var not in MW.keys():
-            print(f"using Carbon molecular weight for {var}")
-            MW[var]=12
+            if 'so4' in var:
+                MW[var]=32.066 # using sulfur molecular weight.
+            else:    
+                print(f"using Carbon molecular weight for {var}")
+                MW[var]=12
 
                 
     # Avogadro's Number
@@ -266,11 +286,16 @@ def aerosol_gas_tables(adfobj, trop_val=None, **kwargs):
         end_period = datetime.strptime(end_date, "%Y-%m-%d")
 
         # Calculated duration of time period in seconds?
-        durations[case] = (end_period-start_period).days*86400 #+365*86400
+        if Climate:
+            # Use annual mean file
+            durations[case] = 365*86400
+            num_yrs[case]=1
 
-
-        # Get number of years for calculations
-        num_yrs[case] = (int(end_year)-int(start_year)) #+1
+        else:
+            # Calculate the duration
+            durations[case] = (end_period-start_period).days*86400 #+365*86400
+            # Get number of years for calculations
+            num_yrs[case] = (int(end_year)-int(start_year)) #+1
 
         # Get currenty history file directory
         data_dir = data_dirs[i]
@@ -289,10 +314,35 @@ def aerosol_gas_tables(adfobj, trop_val=None, **kwargs):
         text = f'\n\t Calculating values for {case}'
         print(text)
         print("\t " + "-" * (len(text) - 2))
-
         # Gather dictionary data for current case
-        # NOTE: The calculations can take a long time...
-        Dic_crit, Dic_scn_var_comp[case],Tropospheric,tropospheric_method = make_Dic_scn_var_comp(adfobj, VARIABLES, data_dir, dic_SE, Files, ext1_SE, AEROSOLS,Tropospheric,trop_val)
+        if Climate:
+            # NOTE: The calculations based on annual mean files
+            Files_str= [str(f) for f in Files]
+
+            #Special ADF variable which contains the output paths for
+            #all generated plots and tables for each case:
+            output_locs = adfobj.plot_location
+            #Convert output location string to a Path object:
+            output_location = Path(output_locs[0])
+
+            cmd = ["ncra", "-O", *Files_str, f"{output_location}/{case}_ANN.nc"]
+            subprocess.run(cmd, check=True)
+            try:
+                subprocess.run(cmd, check=True, capture_output=True, text=True)
+            except FileNotFoundError:
+                print("Error: 'ncra' command not found. Please install NCO utilities.")
+            except subprocess.CalledProcessError as e:
+                print(f"NCO Error (Exit Code {e.returncode}): {e.stderr}")
+
+            #os.sys(f"ncra {Files} {output_location}/{case}_ANN.nc")
+            File_mean=[f"{case}_ANN.nc"]
+
+            Dic_crit, Dic_scn_var_comp[case],Tropospheric,tropospheric_method = make_Dic_scn_var_comp(adfobj, VARIABLES, output_location, dic_SE, File_mean, ext1_SE, AEROSOLS,Tropospheric,trop_val)
+        else:
+            print(f'Using raw files instead of annual mean files.')
+
+            # NOTE: The calculations can take a long time...
+            Dic_crit, Dic_scn_var_comp[case],Tropospheric,tropospheric_method = make_Dic_scn_var_comp(adfobj, VARIABLES, data_dir, dic_SE, Files, ext1_SE, AEROSOLS,Tropospheric,trop_val)
         # Regional refinement
         # NOTE: This function 'Inside_SE' is unavailable at the moment! - JR 10/2024
         if regional:
@@ -310,7 +360,7 @@ def aerosol_gas_tables(adfobj, trop_val=None, **kwargs):
             if tropospheric_method=='pressure':
                 # using pressure > 500hPa
                 trop = np.where(current_crit<500,np.nan,current_crit)
-            if tropospheric_method=='ozone':
+            elif tropospheric_method=='ozone':
                 # using ozone <150 ppb
                 trop = np.where(current_crit>150,np.nan,current_crit)
             elif tropospheric_method=='tropopause':
@@ -401,23 +451,17 @@ def Get_files(adfobj, data_dir, start_year, end_year, h_case, **kwargs):
 
             areas = tmp_area*Earth_area/np.nansum(tmp_area)
         except KeyError:
-            try:
-                tmp_area = tmp_file['AREA'+ext1_SE].isel(time=0).data
-                areas=tmp_area
-                #Earth_area = 4 * np.pi * Earth_rad**(2)
-                #areas = tmp_area*Earth_area/np.nansum(tmp_area)
-            except:
-                dlon = np.abs(lon[1]-lon[0])
-                dlat = np.abs(lat[1]-lat[0])
+            dlon = np.abs(lon[1]-lon[0])
+            dlat = np.abs(lat[1]-lat[0])
 
-                lon2d,lat2d = np.meshgrid(lon,lat)
-                #area=np.zeros_like(lat2d)
+            lon2d,lat2d = np.meshgrid(lon,lat)
+            #area=np.zeros_like(lat2d)
 
-                dy = Earth_rad*dlat*np.pi/180
-                dx = Earth_rad*np.cos(lat2d*np.pi/180)*dlon*np.pi/180
+            dy = Earth_rad*dlat*np.pi/180
+            dx = Earth_rad*np.cos(lat2d*np.pi/180)*dlon*np.pi/180
 
-                tmp_area = dx*dy
-                areas = tmp_area
+            tmp_area = dx*dy
+            areas = tmp_area
     # End if
 
     # Variables to return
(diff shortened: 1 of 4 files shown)


scripts/analysis/aerosol_gas_tables.py
```diff
@@ -136,7 +153,7 @@ def aerosol_gas_tables(adfobj, trop_val=None, **kwargs):
     # if True, calculate only Tropospheric values
     # if False, all layers
     # tropopause is defiend as either directly or indirectly. Look for tropopause to see the definition
-    Tropospheric = bres['Tropospheric']
+    #Tropospheric = bres['Tropospheric']
```
justin-richling, line 156: Sorry for the late review on this @Behrooz-Roozitalab! The only question I have is if this is commented out, could we just remove it, or is this something that might get flagged later?

scripts/analysis/depositions_table.py
```diff
@@ -0,0 +1,975 @@
...
+    # End if
+    
+    # Check to ensure number of case names matches number history file locations.
+    # If not, exit script
+    if len(hist_locs) != len(case_names):
+        errmsg = "Error: number of cases does not match number of history file locations. Script is exiting."
+        raise AdfError(errmsg)
+
+    # Initialize nicknames dictionary
+    #nicknames = {}
```
justin-richling, line 122: Can we remove this if it is not being used?

scripts/analysis/depositions_table.py
```diff
@@ -0,0 +1,975 @@
...
+            # Get number of years for calculations
+            num_yrs[case] = (int(end_year)-int(start_year)) #+1
+
+
+        # Calculated duration of time period in seconds?
+        #durations[case] = (end_period-start_period).days*86400 #+365*86400
+
+
+        # Get number of years for calculations
+        #num_yrs[case] = (int(end_year)-int(start_year)) #+1
```
justin-richling, line 196: Can we also remove these commented out lines?

scripts/analysis/depositions_table.py
```diff
@@ -0,0 +1,975 @@
...
+            cmd = ["ncra", "-O", *Files_str, f"{output_location}/{case}_ANN.nc"]
+            subprocess.run(cmd, check=True)
+            try:
+                subprocess.run(cmd, check=True, capture_output=True, text=True)
+            except FileNotFoundError:
+                print("Error: 'ncra' command not found. Please install NCO utilities.")
+            except subprocess.CalledProcessError as e:
+                print(f"NCO Error (Exit Code {e.returncode}): {e.stderr}")
+
+            #os.sys(f"ncra {Files} {output_location}/{case}_ANN.nc")
```
justin-richling, line 256: Is this also safe to remove?

scripts/analysis/depositions_table.py
```diff
@@ -0,0 +1,975 @@
...
+        else:
+            print(f'Using raw files instead of annual mean files.')
+
+            # NOTE: The calculations can take a long time...
+            Dic_scn_var_comp[case] = make_Dic_scn_var_comp_2D(adfobj, VARIABLES, data_dir, dic_SE, Files, ext1_SE, AEROSOLS)
+
+ 
+        # Gather dictionary data for current case
+        # NOTE: The calculations can take a long time...
+        #Dic_scn_var_comp[case] = make_Dic_scn_var_comp_2D(adfobj, VARIABLES, data_dir, dic_SE, Files, ext1_SE, AEROSOLS)
```
justin-richling, line 269: Can we also remove this and line 273, `#inside = Inside_SE_region(current_lat,current_lon,dir_shapefile)`?

scripts/analysis/depositions_table.py
```diff
@@ -0,0 +1,975 @@
...
+            Earth_area = 4 * np.pi * Earth_rad**(2)
+
+            areas = tmp_area*Earth_area/np.nansum(tmp_area)
+        except KeyError:
+
+            dlon = np.abs(lon[1]-lon[0])
+            dlat = np.abs(lat[1]-lat[0])
+
+            lon2d,lat2d = np.meshgrid(lon,lat)
+            #area=np.zeros_like(lat2d)
```
justin-richling, line 362: Should we keep this or remove it?

scripts/analysis/depositions_table.py
```diff
@@ -0,0 +1,975 @@
...
+                        dic_SE[var+'_DDFC'][cloud_key+ext1_SE+'DDF']=1                            
+                else:
+                    dic_SE[var+'_DDFC']['PS'+ext1_SE]=0.  
+    
+                    
+
+                # for SFWET in cloud water:
+                # original unit: [kg/m2/s]
+                if cloud_key+ext1_SE+'SFWET' in ListVars:   
+                    #if var=='SO4':
```
justin-richling, line 546: Remove this line?

scripts/analysis/depositions_table.py
```diff
@@ -0,0 +1,975 @@
...
+
+            # Gather info for debugging
+            for var_m in missing_vars:
+                if var_m not in missing_vars_tot:
+                    missing_vars_tot.append(var_m)
+            for var_n in needed_vars:
+                if var_n not in needed_vars_tot:
+                    needed_vars_tot.append(var_n)
+        # End for
+    # End for
```
justin-richling, line 657: Looks like this end for comment could go a couple of lines down after `Dic_var_comp[current_var] = Dic_comp`

scripts/analysis/emissions_table.py
```diff
@@ -0,0 +1,897 @@
...
+            # Calculate the duration
+            durations[case] = (end_period-start_period).days*86400 #+365*86400
+            # Get number of years for calculations
+            num_yrs[case] = (int(end_year)-int(start_year)) #+1
+
+
+        # Calculated duration of time period in seconds?
+        #durations[case] = (end_period-start_period).days*86400 #+365*86400
+        # Get number of years for calculations
+        #num_yrs[case] = (int(end_year)-int(start_year)) #+1
```
justin-richling, line 196: Should we also remove these commented out lines?

scripts/analysis/emissions_table.py
```diff
@@ -0,0 +1,897 @@
...
+            Dic_scn_var_comp[case]= make_Dic_scn_var_comp_2D(adfobj, VARIABLES, output_location, dic_SE, File_mean, ext1_SE, AEROSOLS)
+        else:
+            print(f'Using raw files instead of annual mean files.')
+
+            # NOTE: The calculations can take a long time...
+            Dic_scn_var_comp[case] = make_Dic_scn_var_comp_2D(adfobj, VARIABLES, data_dir, dic_SE, Files, ext1_SE, AEROSOLS)
+
+        # Gather dictionary data for current case
+        # NOTE: The calculations can take a long time...
+        #Dic_scn_var_comp[case] = make_Dic_scn_var_comp_2D(adfobj, VARIABLES, data_dir, dic_SE, Files, ext1_SE, AEROSOLS)
```
justin-richling, line 262: Remove this line and line 267?

scripts/analysis/emissions_table.py
```diff
@@ -0,0 +1,897 @@
...
+
+    # gas constanct
+    Rgas=287.04 #[J/K/Kg]=8.314/0.028965
+
+    # Set lists to gather necessary variables for logging
+    missing_vars = []
+    needed_vars = []
+    Dic_all_data={}
+
+#    all_data=[]
```
justin-richling, line 649: Can we remove this line?

scripts/analysis/emissions_table.py
```diff
@@ -0,0 +1,897 @@
...
+            for i in range(len(Plevel)):
+                Plevel[i]=hyai[i]*P0+hybi[i]*PS
+
+            delP=Plevel[1:]-Plevel[:-1]
+
+        for var in vars:
+            if file == 0:
+                    Dic_all_data[var]=[]
+
+       # Star gathering of variable data
```
justin-richling, line 689: Let's indent this to match the formatting and fix typo to `# Start gathering ...`

scripts/analysis/emissions_table.py
```diff
@@ -0,0 +1,897 @@
...
+                    Pres=np.zeros_like(np.array(ds['U'+ext1_SE].isel(time=0)))
+                    for i in range(len(Pres)):
+                        Pres[i]=hyam[i]*P0+hybm[i]*PS
+                rho= Pres/(Rgas*Temp)
+                data=data*delP/rho
+            elif ('BURDEN' in var):
+                data=data*delP
+            else:
+                data=data
+        # End if
```
justin-richling, line 756: Looks like this is also off on the formatting, could you indent/move it to the correct spot?

scripts/analysis/emissions_table.py
```diff
@@ -0,0 +1,897 @@
...
+    output_location = Path(output_locs[0])
+
+    # Loop over model cases
+
+    for i,case in enumerate(nicknames):
+
+        nickname = case
+
+        # Collect row data in a list of dictionaries
+        #durations[case]
```
justin-richling, line 853: Remove this commented out line.

scripts/analysis/aerosol_gas_tables.py
```diff
@@ -136,7 +153,7 @@ def aerosol_gas_tables(adfobj, trop_val=None, **kwargs):
...
 
     ### NOT WORKING FOR NOW
```
justin-richling, line 158: Is this still not currently working, or can we remove this line too?

justin-richling (changes requested):
This PR is mostly good @Behrooz-Roozitalab, thank you for breaking this up logically! It looks like a lot of suggested changes, but it's really only related to commented out code and some indentation suggestions. I will merge this when you get to these, thanks again!

brianpm:
> **A note on this review.** This is a new, AI-generated review process we're trying out, aimed at catching more errors and converging on consistent style and conventions across the ADF. It follows the rules in the new `AGENTS.md` file. If you disagree with any of the conventions it applies, think a finding is wrong, or have other comments on the process, please say so — that feedback is genuinely useful and will shape both the review process and `AGENTS.md` itself. Thanks!

**Summary.** Adds two new analysis scripts (`emissions_table.py`, `depositions_table.py`, ~1900 lines, both derived from `aerosol_gas_tables.py`) and reworks `aerosol_gas_tables.py` to add a `Climate` fast-path (pre-average history files with `ncra`) and to move the `Tropospheric` switch from YAML into a function kwarg. The feature is useful and the PR merges cleanly against `main` with zero conflicts against any other open PR. **Not ready as-is** — there is a guaranteed `NameError`, a copy/paste bug in a budget term, an `ncra` call that runs twice, and a broad `except: → 0` pattern that turns missing model fields into zeros in published tables.

Classification: new/modified diagnostic scripts (`scripts/analysis/`) plus a defaults change. §4.1, §4.2, §4.3, §4.5, §6 apply. CI does not lint or test any of these files (§3), so everything below is un-caught by the workflows.

## Blocking

**1. `depositions_table.py:248` — `subprocess` is never imported.**
The file uses `subprocess.run` at lines 248 and 250 and `subprocess.CalledProcessError` at 253, but the import block (lines 1–11) has no `import subprocess` — `emissions_table.py:4` and `aerosol_gas_tables.py:4` both have it, this one was dropped. The import in `aerosol_gas_tables.py` does not cover this: it binds the name in that module's globals only, and `depositions_table` imports nothing from it.
*Failure:* any config running `depositions_table` with `{depositions_table: {kwargs: {Climate: True}}}` dies with `NameError: name 'subprocess' is not defined` on the first case. `Climate: False` (the default) never reaches the line, which is why it hasn't shown up.

**2. `aerosol_gas_tables.py:1257` — the CHMP fallback assigns the wrong variable.**

```python
try:
    spc_chmp = Dic_scn_var_comp[current_var][current_var+'_CHMP']
except:
    spc_chml = 0                                    # <- should be spc_chmp
spc_chmp = np.where(np.isnan(trop), np.nan, spc_chmp)   # line 1258
```

Two bugs in three lines: the handler doesn't define `spc_chmp`, so line 1258 raises `UnboundLocalError` anyway (the `try` protects nothing); and it clobbers `spc_chml`, which was legitimately read at line 1240 and is still needed. *Failure:* any run whose history files lack `<var>_CHMP` — the exact case this `try` was added for.

**3. `ncra` runs twice, and the guarded call is the second one.**
`aerosol_gas_tables.py:329-335`, `emissions_table.py:242-248`, `depositions_table.py:248-254`:

```python
subprocess.run(cmd, check=True)                                  # unguarded, runs first
try:
    subprocess.run(cmd, check=True, capture_output=True, text=True)   # same cmd again
except FileNotFoundError: ...
except subprocess.CalledProcessError as e: ...
```

The first line looks like a leftover. It doubles the cost of the most expensive step in the `Climate` path (an `ncra` over every history file in the period), and because it is unguarded it raises before the `try` ever executes — so the error handlers below are unreachable. Delete the bare call and keep the guarded one. Also §4.5: on a failure the code prints and then falls through to `make_Dic_scn_var_comp*` with a `{case}_ANN.nc` that was never written, producing a second, more confusing error — it should warn and fall back to the raw-file path, or `continue`.

**4. The `except: <term> = 0` pattern silently fabricates budget numbers (§5.2).**
~15 new blocks in `aerosol_gas_tables.py` (1201, 1210, 1220, 1230, 1241, 1274, 1286, 1297, 1312, 1322, 1335, 1345, 1377, 1394) and the equivalents in the two new files replace a missing field with the scalar `0`, which propagates to a rounded `0.0` in the published table. A missing `<var>_DDF` and a genuinely zero dry deposition are then indistinguishable to the reader, and any total computed from those terms is wrong with no indication.

This is the §4.2 "missing input files are normal" case, but the required behavior there is warn-and-skip, not warn-and-zero. The script already tracks `missing_vars_tot` — these terms should be omitted from `chem_dict` (so the column is absent or `NaN`), with the variable name accumulated into the existing missing-variable message. All of these are also bare `except:` (§6.3) and will swallow `KeyboardInterrupt` and any bug in the indexing expression; they should be `except KeyError:`.

**5. New scripts are not registered in `config_cam_baseline_example.yaml` (§4.1).**
`emissions_table` and `depositions_table` appear in no config file in the repo. The established pattern is the commented-out opt-in line — `config_cam_baseline_example.yaml:515` already has `#- aerosol_gas_tables`. Without it there is nothing telling a user these exist or how to pass `Climate`/`Tropospheric` via the `kwargs` dict form. (Entry-point names themselves are correct: `emissions_table.py` → `emissions_table()`, `depositions_table.py` → `depositions_table()`, and neither basename collides on the flat `scripts/` `sys.path`.)

**6. `Tropospheric` removed from `adf_variable_defaults.yaml` with no migration path (§4.3).**
The key is deleted from the `budget_tables` block and the read at `aerosol_gas_tables.py:156` is commented out; it is now a kwarg defaulting to `True`. Anyone running with a custom `variable_defaults` file that sets `Tropospheric: False` will have that setting **silently ignored** and get tropospheric budgets instead of total-column — no error, just different numbers. The PR description says this option was *added* to the YAML file; it was in fact removed from it. Needs either a backward-compatible fallback (`Tropospheric = kwarg if not None else bres.get('Tropospheric', True)`) or a loud note in the PR description per §4.3.

## Non-blocking

**7. Removing the `AREA` fallback may break unstructured grids.** `aerosol_gas_tables.py:452-467` — the old code tried `area`, then `AREA`, then a rectilinear `dlon/dlat` computation; now `AREA` is gone and a `KeyError` on `area` falls straight to the rectilinear branch. For an SE/unstructured grid, `lat`/`lon` are 1-D `ncol` arrays, so `np.meshgrid(lon, lat)` builds an `(ncol, ncol)` field and `lat[1]-lat[0]` is meaningless. The stated motivation (CESM `AREA` is wrong in some versions) is fair, but the replacement should detect an unstructured grid and warn/skip rather than compute a rectilinear area for it.

**8. `ncra` output lands in the website plot directory.** All three scripts write `{case}_ANN.nc` to `Path(adfobj.plot_location[0])`. Two issues: it puts a potentially large intermediate data file in the directory that becomes the published website, and `[0]` is hard-coded regardless of the loop index `i`, so every case's intermediate goes into the *first* case's plot dir. A scratch/temp location (or `plot_location[i]`) would be better, plus cleanup after use.

**9. Duration is inconsistent between the two branches.** `emissions_table.py:183` vs `:188` — `Climate: True` hard-codes `365*86400`, while the raw-file path uses `(end_period - start_period).days * 86400` from `datetime`, i.e. the *proleptic Gregorian* calendar. For a period spanning leap years those disagree by ~0.07% per leap year, and CAM's default calendar is `noleap`, so the `datetime` value is the wrong one. Worth making both `365 * 86400 * n_years`.

**10. Docstrings do not follow §6.2.** Both new files: no module docstring; the entry-point docstrings are free prose with no `Parameters` / `Returns` sections and no `Notes` block listing the `adfobj` attributes used (these scripts touch `variable_defaults`, `get_cam_info`, `case_nicknames`, `climo_yrs`, `hist_string`, `plot_location`, `compare_obs`, `get_baseline_info`, `debug_log`, `add_website_data` — exactly what that convention exists to surface). The heading underline at `emissions_table.py:21` is also shorter than its title, which Sphinx will drop. §6.2 applies to new code even though the file it was copied from predates the convention.

**11. `black` on new code (§6.1).** The two new files are new code, so §6.1 applies in full — currently they carry the source file's spacing (`Climate=None,**kwargs`, an f-string with no placeholders at `emissions_table.py:262`). A `darker --check --diff --revision main...HEAD` pass would cover it. Low priority relative to everything above.

**12. YAML formatting and a behavior question.** `adf_variable_defaults.yaml:2438-2443` — the `AEROSOLS` continuation lines mix 3- and 4-space indentation, and the diff drops two blank lines around `ext1_SE` for no reason. Also worth confirming the intent of adding the modal species (`so4_a1`…`dst_a3`): membership in `AEROSOLS` selects which branch `calc_budget_data` takes (DDF+DDFC / WDF+WDFC vs plain DDF/WDF), so this changes results for any existing user who has those in `VARIABLES`.

## Optional / follow-up (pre-existing, not this PR's job)

- `Get_files` returns `areas` unbound when `area=False` (`aerosol_gas_tables.py:446-470`) — `NameError` if ever called that way.
- Time averaging is unweighted throughout (`np.nanmean(..., axis=0)` at `emissions_table.py:761`, and `ncra` without `-w`), so monthly means are averaged without month-length weighting. Pre-existing and consistent between the old and new paths, so `Climate` does not make it worse — but it is a real ~0.1% bias in all of these tables and would make a good standalone issue.
- `env/conda_environment.yaml` has no `nco` entry even though the framework already requires `ncrcat`/`ncatted` for time-series generation (`lib/adf_diag.py:668`). Pre-existing framework gap, worth its own issue.
- Duplicate `import numpy as np` at lines 1 and 6 of all three files.

## Not verified

- **Whether the `Climate: True` numbers actually match the raw-file numbers.** This needs a real ADF run on chemistry history output; it was not run for this review. Given the duration/`num_yrs` differences between the branches (finding #9), a side-by-side of the two paths on the same case is probably the most valuable check.
- **Science correctness of the new emission/deposition term lists and the `so4 → MW 32.066` assignment** (`aerosol_gas_tables.py:172-176`) — a call for the chemistry owner, not a code review. Note the substring test is lowercase `'so4' in var`, so it applies to the new modal species but not to `SO4`, which is presumably already in the `MW` dict; worth confirming that's deliberate.
- Whether any downstream consumer depends on `budget_tables: Tropospheric` in a custom variable-defaults file (finding #6) — only this repo was grepped.

---

Findings #1–#3 are mechanical and quick. #4 (the silent zeros) has the real scientific consequence and is the biggest ask. Happy to help with any of these.

scripts/analysis/aerosol_gas_tables.py
Behrooz-Roozitalab: It should be removed now.

scripts/analysis/aerosol_gas_tables.py
Behrooz-Roozitalab: It is still not working. Let's keep it so I know I should do it!

Behrooz-Roozitalab:
Hi Justin,

Thanks for reviewing it. And the AI review was helpful. About the AI comment on 'silent zeros', we actually don't report any 0 value and provide a comment. I clarified the comment.

Issue #335, opened 2024-10-17
file indexing for files containing more than 1yr of data fails if start or end year are not named in files


bitterbark:
### ADF run type

Model vs. Model

### What happened?

If there is more than one year of data in a file, every year is not represented in the list of file names. Calling for a start or end year that is not alos one of the file names throws an error that the years are not in range (even though they are).

Eg: I have 3 years of data in each file, model run from year 1-22, so file names are
> b.e30_beta02.BLT1850.ne30_t232.104_mdtf.cam.h5a.0001-01-01-10800.nc
> b.e30_beta02.BLT1850.ne30_t232.104_mdtf.cam.h5a.0004-01-01-10800.nc
> ...
> b.e30_beta02.BLT1850.ne30_t232.104_mdtf.cam.h5a.0022-01-01-10800.nc
> 

If I try to run with start year = 2 and end year = 6, I get an error that the years are out of range. But in fact it is only seeing if the years 3 & 6 are *in the set* of the years from the file names:

> case_climo_yrs = [int(str(i).partition(f"{hist_str}.")[2][0:4]) for i in file_list]
> print(f"{case_climo_yrs=} {case_found_syr=} {case_found_eyr=}")
> 
 _case_climo_yrs=[1, 4, 7, 10, 13, 16, 19, 22] case_found_syr=1 case_found_eyr=22_

So we should just check if the years are in the range `[case_found_syr,case_found_eyr]` instead of in the set `case_climo_yrs`

### ADF Hash you are using

5f825af

### What machine were you running the ADF on?

CISL machine

### What python environment were you using?

NPL (CISL machines only)

brianpm:
Addressed in #458

Issue #306, opened 2024-05-16
Incorrect year span on plots


fzhu2e:
### ADF run type

Model vs. Model

### What happened?

Hi, in my ADF runs, I skipped the "create timeseries" and "create climo" steps and did `regrid_climo()` and other following steps directly after `adf = AdfDiag(config_file)`

The output plots of the baseline cases somehow didn't use the year span I set in the configuration files: [configs.zip](https://github.com/NCAR/ADF/files/15340855/configs.zip)
<img width="1037" alt="image" src="https://github.com/NCAR/ADF/assets/1380844/aca268d3-f555-4956-8b9b-3f1ad6155161">
In the configuration file, I set 1-500 for this case, but got 1-324 in the plot.

<img width="1023" alt="image" src="https://github.com/NCAR/ADF/assets/1380844/5e1ebfa3-c0d7-4dc6-a7db-49126cb123af">
In the configuration file, I set 1-504 for this case, but got 1-320 in the plot.

### ADF Hash you are using

536325f

### What machine were you running the ADF on?

CISL machine

### What python environment were you using?

NPL (CISL machines only)

brianpm:
Addressed in #458

PR #458, opened 2026-06-25
Fix year-range validation for start/end years (#306, #335)

brianpm:
Validate user-supplied start_year/end_year against the discovered [found_syear, found_eyear] range instead of membership in a constructed collection, at all four sites (test/baseline x timeseries/history).

- #306: the timeseries paths used np.arange(found_syear, found_eyear, 1), which is end-exclusive, so a config end_year equal to the last available year was wrongly rejected and silently overridden (shrinking the plotted span by one). Replaced with an inclusive range comparison and removed the now-unused found_yr_range.
- #335: the history paths tested membership in case_climo_yrs/base_climo_yrs (the set of years parsed from filenames, one per file). When files span more than one year, an in-range year not literally named in a filename was wrongly rejected. Replaced with a range comparison against the found min/max years.

Also corrected the baseline timeseries end-year warning ("first" -> "last"). The warn-and-override fallback is retained; it now fires only for genuinely out-of-range requests.

### DIFF
--- a/lib/adf_info.py
+++ b/lib/adf_info.py
@@ -213,7 +213,6 @@ class AdfInfo(AdfConfig):
                 #Get years from pre-made timeseries file(s)
                 found_syear_baseline, found_eyear_baseline = self.get_climo_yrs_from_ts(
                     input_ts_loc, data_name)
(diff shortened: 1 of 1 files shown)


nusbaume (approved):
Thanks @brianpm!

Issue #410, opened 2025-08-20
Create TS for multiple streams (h0, h1, h2)


wwieder:
### ADF run type

Model vs. Model

### What happened?

Currently passing a list of `hist_str` fails on the clm-diag branch.

Also this can be avoided by running config file for h1 files and then larger script with h0 files, but then users ALSO have to make sure the climo files are created from the h0 files (not from h1), as climo file names don't include information from `hist_str`

### ADF Hash you are using

53a4d784

### What machine were you running the ADF on?

CISL machine

### What python environment were you using?

Other (please explain below)

### Extra info

running with cupid-analysis conda environment

brianpm:
Addressed in #459

PR #459, opened 2026-06-25
Include history stream in climo file names (#410)

brianpm:
Climo files are now named {case}_{hist_str}_{var}_climo.nc and are produced per configured history stream, so a variable present in more than one stream (e.g. h0 and h1) no longer collides. Climo/timeseries lookups resolve the stream-aware name in the case's hist_str priority order, then fall back to the older {case}_{var}_climo.nc convention so pre-existing climo files still work.

- create_climo_files.py: loop over the case's streams; stream-agnostic fallback when the stream is unknown (e.g. baseline run on pre-made time series).
- adf_dataset.py: optional hist_str filter on the timeseries getters; priority + old-convention fallback in get_climo_file/get_reference_climo_file.

Also includes a pylint cleanup of both touched files (whitespace, line length, import order, raise-from, redundant worker imports).

### DIFF
--- a/lib/adf_dataset.py
+++ b/lib/adf_dataset.py
@@ -125,7 +125,7 @@ class AdfData:
         """Return list of test time series files.
 
         If hist_str is given, restrict the search to that history stream
-        (time series files are named {case}.{hist_str}.{field}.*nc).
+        (time series files are named {case}.{hist_str}.{field}.*.nc).
         """
         # list of paths (could be multiple cases)
         ts_locs = self.adf.get_cam_info("cam_ts_loc", required=True)
--- a/lib/adf_dataset.py
+++ b/lib/adf_dataset.py
@@ -1,12 +1,14 @@
+"""Data-access layer: locate and load ADF time series, climo, and regridded files."""
+import warnings # use to warn user about missing files
 from pathlib import Path
+
 import xarray as xr
 
 import adf_utils as utils
-import warnings # use to warn user about missing files
 warnings.formatwarning = utils.my_formatwarning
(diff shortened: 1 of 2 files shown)


lib/adf_dataset.py
```diff
@@ -86,38 +94,75 @@ def set_ref_var_loc(self):
...
     #------------------
     # Test case(s)
-    def get_timeseries_file(self, case, field):
-        """Return list of test time series files"""
-        ts_locs = self.adf.get_cam_info("cam_ts_loc", required=True) # list of paths (could be multiple cases)
+    def get_timeseries_file(self, case, field, hist_str=None):
+        """Return list of test time series files.
+
+        If hist_str is given, restrict the search to that history stream
+        (time series files are named {case}.{hist_str}.{field}.*nc).
```
nusbaume: Nit-pick, but I might change the glob search strings to end with `*.nc` instead of `*nc`, simply because if there is a time series file that doesn't end with a `.nc` then I would be worried that it is not actually a NetCDF file, which could problems farther upstream that would be harder to debug.  So for example:


```suggestion
        (time series files are named {case}.{hist_str}.{field}.*.nc).
```

This same request holds for the glob search strings in `get_ref_timeseries_file` as well.  Thanks!

nusbaume (approved):
Thanks @brianpm!  I had one optional request, but nothing that should hold up this PR.

lib/adf_dataset.py
brianpm: I agree, I'll make this change.

PR #463, opened 2026-08-06
Fix website generation for non-default plot types

brianpm:
Any plotting script using a plot_type outside "default_ptypes" crashed website generation with "TypeError: unhashable type: 'collections.OrderedDict'" from template_index.html.

The loop meant to register an unrecognized plot type appended the plot_types dict rather than the plot type itself, so index.html then tried to use that dict as a dictionary key.

Also copy the default list before appending. "res" is the shared variable-defaults dictionary and this loop runs once per web-data entry, so appending in place permanently modified the loaded defaults.

Verified with a plotting script registering plot_type "CloudRegimes": the tab now renders on index.html and the generated site has no broken internal links.

### DIFF
--- a/env/conda_environment.yaml
+++ b/env/conda_environment.yaml
@@ -1,4 +1,4 @@
-name: adf_v1.0.1
+name: adf_v1.0.3
 channels:
   - conda-forge
   - defaults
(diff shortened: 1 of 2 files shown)


nusbaume (approved):
Thanks @brianpm!

Issue #462, opened 2026-08-06
Problem with adf_v1.0.0 environment, do we need an updated conda_environment.yaml file?


TomasTorsvik:
I was recently getting an error when activating the `adf_v1.0.0` environment for NorESM diagnostics. The nature of the problem suggests that this is a general problem for the ADF `conda_environment.yaml` file, hence I raise the issue here.

The reason that `adf_v1.0.0` does not work is that the `xskillscore` version is using a module `pkg_resources` that has been deprecated and is no longer supported. If the environment was built some time ago, the python version may still be compatible, but for a new conda environment build, the `pkg_resources` is not available.

There are two obvious option to solve this:
- downgrade `setuptools` to an earlier version, i.e. "setuptools<=82.0.0" and keep the conda environment as it is now.
- Update the packages in the `conda_environment.yaml` and make a new version, e.g. `adf_v1.0.1` or `adf_v1.1.0`. The package versions were last updated in June 2025.

As far as I can see, there are no existing issues/PRs that will update the conda environment. If you agree, I will prepare a new `conda_environment.yaml` and provide a draft PR for this for testing.

TomasTorsvik:
@mvertens, @adagj, @YanchunHe - I made an issue here in the NCAR/ADF repository instead of in the NorESMhub/ADF repository, since it is probably best to coordinate any environment updates with a larger ADF community, not only for NorESM.

mvertens:
@TomasTorsvik - please put your issue here - and not in the NCAR ADF. At some point I will do a PR back to NCAR - but all the current development we are doing is in the NorESMHub respository - so please keep issues in this repository.

mvertens:
@TomasTorsvik - please activate the conda environment as outlined in the README.md. That has worked for everyone else who is using the updated ADF. Please let me know if you encounter problems at that point.

mvertens:
@TomasTorsvik - Just as a quick summary - clone the ADF, make sure you are on the noresm branch and then issue
`> conda activate /projects/NS9560K/diagnostics/ADF/envs/adf_v0.13`
Everything should work for you. Please let me know if it does not.

TomasTorsvik:
@mvertens - the `adf_v0.13` environment works, as long as it remains in place and we do not need to re-create it. I am just pointing out that if take the file from the ADF repository, do
```
> mamba env create -f conda_environment.yaml
```
[...]

PR #464, opened 2026-08-11
Remove un-used 'xskillscore' from provided conda environment.

nusbaume:
This PR removes the `xskillscore` package, which doesn't appear to actually be used, and was causing an issue with its currently pinned version (see issue #462 ).  If we eventually decide that we do want the `xskillscore` package then we can always add it back with an updated version.

Also please note that a new tag will be needed once this PR has been merged (as the conda environment will be updated).

### DIFF
--- a/env/conda_environment.yaml
+++ b/env/conda_environment.yaml
@@ -1,4 +1,4 @@
-name: adf_v1.0.0
+name: adf_v1.0.1
 channels:
   - conda-forge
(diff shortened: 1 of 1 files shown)


justin-richling (approved):
This looks good @nusbaume, hopefully we can figure this out soon so we can be ready when `xskillscore` is needed, thanks!

PR #469, opened 2026-09-03
Use the configured years to choose time series files (#433)

### LINKED ISSUE #433: ADF and LDF cannot create plots with specified date ranges if tseries are available for multiple time ranges
### ADF run type

Model vs. Model

### What happened?

ADF and LDF both fail when multiple timeseries files are found. This is a bit of a hidden 'gotcha' that is coming up regularly when running these diagnostics via CUPiD on recent development runs.

In these cases, tseries have been created already (say, years 1-20), and then the baseline run continues, and we want to use a larger range of timesteps. Thus, we end up creating a second set of tseries (eg, years 1-40). Then, with the ADF timeseries tool, we end up with tseries for 1-20 AND tseries for 1-40. I think this is fine, but it causes issues down the line when ADF looks for timeseries and finds both.

Is it possible to specify the years when looking for tseries files to avoid running into this issue?

### ADF Hash you are using

6aff7b6

CISL machine

Other (please explain below)

### Extra info

conda environment: cupid-analysis

brianpm:
Fixes #433.

## The problem

A time series directory can end up holding two sets of files for the same
variable. This happens when a run is post-processed over years 1-20, the run
is then extended, and the time series are made again over years 1-40. Both
sets are left in place.

ADF found all of those files and opened them together, because nothing looked
at `start_year` and `end_year` when searching for files. What happened next
depended on the two sets:

- If the sets only partly overlap (years 1-20 next to years 10-40), xarray
  raised `Resulting object does not have monotonic global indexes along
  dimension time` and the run stopped.
- If the sets start in the same year (years 1-20 next to years 1-40), the
  files opened without any error, and the values in the overlapping years came
  from whichever file happened to win. This gave an answer that looked fine
  but was not built from the data the user asked for.

The first place this happens is `AdfInfo.get_climo_yrs_from_ts`, which runs
while ADF is still reading its configuration. So the run died before any
diagnostic started.

## The fix

`start_year` and `end_year` are already in the config file. They now decide
which files are used, so no new setting is needed.

- New `adf_file_utils.select_ts_files(files, start_year, end_year)`. It reads
  the dates out of each file name and keeps only the files needed to cover the
  requested years. At each step it takes the file that begins at or before the
  first year still needed, preferring the smallest file that holds all of the
  rest, and otherwise the file that reaches furthest. A variable split into
  consecutive pieces (what GenTS writes when `slice_years` is set) therefore
  keeps every piece it needs, while a duplicate set is passed over.
- It is called from every place that finds time series files: the shared
  search in `AdfData` (which covers climatologies, regridding and the time
  series plots), `amwg_table`, `adf_derive` (the years are passed down through
  `derive_variable` from both time series back ends), the three plotting
  scripts that do their own search (`qbo`, `tape_recorder`,
  `regional_map_multicase`), and the MDTF file move.
- The MDTF file move used to take the first file it found and warn that it was
  "continuing with best guess". A variable split into consecutive files is now
  written out as the single file MDTF expects. A lone file is still copied.
- `AdfData` writes a line to the debug log naming the files it used whenever a
  directory turned out to hold more than one set.
- `AdfInfo.get_climo_yrs_from_ts` now takes the available years from the file
  names instead of opening the files. This is what keeps the run alive long
  enough for the configured years to be used.

Only a set of files that cannot be opened as it stands is narrowed. Anything
that already combines cleanly is passed through untouched and ADF behaves
exactly as before, which matters because the plots that show a whole record
(`global_mean_timeseries`, for one) ask for every file a case has, not only
the climatology years. The other pass-through cases are: fewer than two files,
no years given, a year range given backwards, file names whose dates cannot be
read, dates that would have to be compared at finer than year resolution, a
gap in the requested range, and a set that cannot be reduced to a usable
choice at all.

One case is improved rather than fixed. If two sets only partly overlap
(years 1-20 beside years 10-40) *and* the config gives no years at all, the
available range is reported as years 1-40, both files are needed to cover it,
and opening them still fails. It now fails in the climatology step, where it
is caught and reported for that one variable, instead of stopping the whole
run during setup. Giving the years, which is what the issue asks for, avoids
it entirely.

## Testing

Unit tests: 89 pass. New cases cover the two overlapping layouts, consecutive
pieces, a range that sits inside a single file, a chunked set next to a
whole-period set, files split inside a year, a set that cannot be reduced to a
usable choice, and every pass-through case. Three cases were added for
`get_climo_yrs_from_ts` and one for `derive_variable`.

I also corrected five file names in `test_adf_info_climo_yrs.py` that claimed
one year more than the file held (for example `010001-010512.nc` holding years
100-104). Nothing read those names before, so they were harmless until now.

End to end on Casper, model against model, two deliberately different cases
(CESM3 alpha 1850 control, `cam.h0a`, 93 levels; and an AMIP run, `cam.h0`,
32 levels):

1. Made a 5-year set of time series.
2. Made a 10-year set in the same directory, which is the situation in the
   issue.
3. Ran the 5-year years again with both sets present.

All three finished with no errors, and each climatology file records in its own
attributes that it used the single correct set of files.

Regression check, same machine, with the three existing test configurations
(built-in back end, GenTS back end, and comparison against observations): all
finished, with the same number of climatology, regridded, plot, table and web
page products as before the change, and the two time series back ends still
produce identical data.

`global_mean_timeseries`, `tape_recorder` and `regional_map_multicase` were
also run against a directory holding two sets, and the debug log confirms which
file each variable used.

## Two things left for follow-up

- The one-line change to `qbo` is compile-checked but was not run. It needs a
  `U` time series, which the test configuration does not have. Same one-line
  change as the other two scripts.
- Three problems that already exist on `main` came to light while running those
  scripts. None is touched here, and each looks worth its own issue:
  1. `tape_recorder` line 147 calls `pd.date_range(..., freq='M')`, which
     recent pandas refuses ("'M' is no longer supported for offsets. Please use
     'ME' instead"). It stops the whole plotting stage. Picking the right fix
     is a question of which pandas versions the ADF supports, since `'ME'` does
     not exist before pandas 2.2.
  2. `AdfInfo` only records the baseline history string when it reads history
     files, so with `cam_ts_done: true` it stays an empty string.
     `tape_recorder` then builds a shorter list of history strings than it has
     cases and raises `IndexError`, which means it cannot run against pre-made
     time series at all.
  3. `regional_map_multicase` calls `prep_contour_plot` with three positional
     arguments, but that function now requires four (`pctdata`), so it raises
     `TypeError` as soon as it tries to plot.

### DIFF
--- a/lib/adf_dataset.py
+++ b/lib/adf_dataset.py
@@ -140,10 +140,15 @@ class AdfData:
         """
         chosen = utils.select_ts_files(fils, syr, eyr)
         if len(chosen) != len(fils):
-            msg = f"time series for '{field}': {len(fils)} files found, using "
-            msg += f"the {len(chosen)} needed for years {syr}-{eyr}: "
-            msg += ", ".join(str(Path(f).name) for f in chosen)
-            self.adf.debug_log(msg)
+            msg = f"\t    INFO: '{field}' has {len(fils)} time series files that "
+            msg += "cannot be used together, so the "
+            msg += f"{len(chosen)} needed for years {syr}-{eyr} were used."
+            #Say it on screen as well as in the log: the numbers a user sees
+            #change when this happens, and that should not be silent.
+            print(msg)
+            self.adf.debug_log(
+                msg + "  Files used: "
+                + ", ".join(str(Path(f).name) for f in chosen))
         #End if
         return chosen
 
--- a/lib/adf_dataset.py
+++ b/lib/adf_dataset.py
@@ -137,6 +137,11 @@ class AdfData:
 
         If hist_str is given, restrict the search to that history stream
         (time series files are named {case}.{hist_str}.{field}.*.nc).
+
+        The files are narrowed to those needed for the case's climatology
+        years, so that a directory holding more than one set for the same
+        variable (years 1-20 alongside years 1-40, say) does not produce a
+        combined time axis with duplicate times.
         """
         # list of paths (could be multiple cases)
         ts_locs = self.adf.get_cam_info("cam_ts_loc", required=True)
@@ -146,13 +151,17 @@ class AdfData:
             ts_filenames = f'{case}.{hist_str}.{field}.*nc'
         else:
             ts_filenames = f'{case}.*.{field}.*nc'
-        return utils.find_ts_files(ts_loc, ts_filenames)
+        fils = utils.find_ts_files(ts_loc, ts_filenames)
+        climo_yrs = self.adf.climo_yrs
+        return utils.select_ts_files(fils, climo_yrs["syears"][caseindex],
+                                     climo_yrs["eyears"][caseindex])
 
     # Reference case (baseline/obs)
     def get_ref_timeseries_file(self, field, hist_str=None):
         """Return list of reference time series files.
 
         If hist_str is given, restrict the search to that history stream.
+        Narrowed to the baseline's climatology years, as for the test cases.
         """
         if self.adf.compare_obs:
             warnings.warn("\t    WARNING: ADF does not currently expect "
@@ -163,7 +172,10 @@ class AdfData:
             ts_filenames = f'{self.ref_case_label}.{hist_str}.{field}.*nc'
         else:
             ts_filenames = f'{self.ref_case_label}.*.{field}.*nc'
-        return utils.find_ts_files(ts_loc, ts_filenames)
+        fils = utils.find_ts_files(ts_loc, ts_filenames)
+        climo_yrs = self.adf.climo_yrs
+        return utils.select_ts_files(fils, climo_yrs["syear_baseline"],
+                                     climo_yrs["eyear_baseline"])
 
 
     def load_timeseries_dataset(self, fils):
--- a/lib/adf_dataset.py
+++ b/lib/adf_dataset.py
@@ -132,6 +132,21 @@ class AdfData:
     # Time series files
     #------------------
     # Test case(s)
+    def _select_ts_files(self, fils, syr, eyr, field):
+        """Narrow time series files to the years wanted, saying so in the log.
+
+        Files are only narrowed when they could not be opened together, so a
+        message here means the directory held more than one set for `field`.
+        """
+        chosen = utils.select_ts_files(fils, syr, eyr)
+        if len(chosen) != len(fils):
+            msg = f"time series for '{field}': {len(fils)} files found, using "
+            msg += f"the {len(chosen)} needed for years {syr}-{eyr}: "
+            msg += ", ".join(str(Path(f).name) for f in chosen)
+            self.adf.debug_log(msg)
+        #End if
+        return chosen
+
     def get_timeseries_file(self, case, field, hist_str=None):
         """Return list of test time series files.
 
@@ -153,8 +168,8 @@ class AdfData:
             ts_filenames = f'{case}.*.{field}.*nc'
         fils = utils.find_ts_files(ts_loc, ts_filenames)
         climo_yrs = self.adf.climo_yrs
-        return utils.select_ts_files(fils, climo_yrs["syears"][caseindex],
-                                     climo_yrs["eyears"][caseindex])
+        return self._select_ts_files(fils, climo_yrs["syears"][caseindex],
+                                     climo_yrs["eyears"][caseindex], field)
 
     # Reference case (baseline/obs)
     def get_ref_timeseries_file(self, field, hist_str=None):
@@ -174,8 +189,8 @@ class AdfData:
             ts_filenames = f'{self.ref_case_label}.*.{field}.*nc'
         fils = utils.find_ts_files(ts_loc, ts_filenames)
         climo_yrs = self.adf.climo_yrs
-        return utils.select_ts_files(fils, climo_yrs["syear_baseline"],
-                                     climo_yrs["eyear_baseline"])
+        return self._select_ts_files(fils, climo_yrs["syear_baseline"],
+                                     climo_yrs["eyear_baseline"], field)
 
 
     def load_timeseries_dataset(self, fils):
--- a/lib/adf_derive.py
+++ b/lib/adf_derive.py
@@ -152,7 +152,8 @@ def check_derive(self, res, var, case_name, diag_var_list, constit_dict, hist_fi
 
 ########
 
-def _find_constit(ts_dir, case_name, constit, hist_str=None):
+def _find_constit(ts_dir, case_name, constit, hist_str=None, *,
+                  syr=None, eyr=None):
     """
     Locate a constituent's time series file(s) for one case and stream.
 
@@ -174,6 +175,11 @@ def _find_constit(ts_dir, case_name, constit, hist_str=None):
         variable name to search for
     hist_str : str, optional
         history stream being processed, e.g. "cam.h0a"
+    syr, eyr : int, optional
+        first and last year being processed.  When given, only the files
+        needed to cover them are returned, so a directory holding more than
+        one set of files for the same constituent (years 1-20 alongside years
+        1-40, say) can still be used.
 
     Returns
     -------
@@ -189,17 +195,24 @@ def _find_constit(ts_dir, case_name, constit, hist_str=None):
     for pattern in patterns:
         found = utils.find_ts_files(ts_dir, pattern)
         if found:
-            return found
+            return utils.select_ts_files(found, syr, eyr)
         #End if
     #End for
     return []
 
 
 def derive_variable(self, case_name, var, res=None, ts_dir=None,
-                         constit_list=None, overwrite=None, hist_str=None):
+                         constit_list=None, overwrite=None, hist_str=None, *,
+                         syr=None, eyr=None):
     """
     Derive variables acccording to steps given here.  Since derivations will depend on the
     variable, each variable to derive will need its own set of steps below.
+
+    The years being processed (`syr`, `eyr`) narrow each constituent's files
+    to the ones needed, so a directory that holds two sets for the same
+    constituent -- years 1-20 alongside years 1-40, which happens when a run
+    is extended and the time series are remade over a longer period -- can
+    still be used.
 
     A constituent may be split across several consecutive time series files,
     which is what GenTS produces when 'gents_slice_years' is set.  All of a
@@ -232,7 +245,8 @@ def derive_variable(self, case_name, var, res=None, ts_dir=None,
     constit_matches = {}
     for constit in constit_list:
         # Check if the constituent file(s) are present, if so add them to the dict
-        matches = _find_constit(ts_dir, case_name, constit, hist_str)
+        matches = _find_constit(ts_dir, case_name, constit, hist_str,
+                                syr=syr, eyr=eyr)
         if not matches:
             continue
         if utils.ts_files_overlap(matches):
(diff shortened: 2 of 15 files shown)


brianpm-ucar (commented):
## Review of the change as originally submitted (commits through `dc2fd62`)

The core idea is right and it does fix the layout in #433 — I reproduced the fix
working. Tests are real (84/84 passing), pylint stays above the 9.5 gate, no config
keys, dependencies or script interfaces change. Three problems stood in the way of
merging, listed first, followed by smaller points.

### Blocking at the time of review

**1. The full-record plots were truncated.** `AdfData.get_timeseries_file` is shared
by two consumers with opposite needs. `create_climo_files` wants only the
climatology years, but `load_timeseries_da` feeds
`scripts/plotting/global_mean_timeseries.py`, which plots the whole record against
the LENS2 envelope and does no year subsetting of its own. Concrete case: an
archive chunked in ten-year files covering years 1-100 with `start_year: 91`,
`end_year: 100`. Before the change the plot showed 100 years; after it showed 10.
Single-file time series are unaffected, which is why the end-to-end runs did not
catch it — but chunked archives are exactly the audience of #433. It also left
`qbo.py` and `tape_recorder.py`, which glob for themselves, showing the full record
while `global_mean_timeseries` no longer did.

**2. Selection compared years only, so a file holding part of a year could be
dropped.** Two verified cases:

- `['case.cam.h1.PRECT.00010101-00010630.nc', 'case.cam.h1.PRECT.00010701-00011231.nc']`
  with years 1-1 returned only the first file. Half the year is gone with no
  warning; on `main` both are opened and the climatology is correct.
- `['...00010101-00050630.nc', '...00050701-00101231.nc']` with years 1-5 returned
  only the first, dropping July-December of year 5 from a "five year" climatology.

ADF's own writer and GenTS are year-aligned, so this is unlikely — but nothing
validates it and the failure is invisible.

**3. The greedy walk could return a set that still cannot be opened.** With a
25-year set beside 10-year chunks of the same run (`000101-002512` plus
`000101-001012`, `001101-002012`, `002101-003012`) and years 1-30 requested, the
result was the 25-year file plus the last chunk, which duplicate years 21-25. I
confirmed `xr.open_mfdataset` raises `ValueError: Resulting object does not have
monotonic global indexes along dimension time` on that pair. The three chunks were
a perfectly good answer. A brute-force search over random layouts found 28 cases in
4000 where a valid cover exists but the greedy walk misses it. This is not a
regression — `main` fails the same way, and the failure degrades gracefully to a
per-variable `Failed:` message — but a function that chooses should not hand back a
choice it can see is unusable.

### Non-blocking

- The description's claim that the selector is called from every place that finds
  time series files is not accurate. `scripts/plotting/qbo.py:98`,
  `scripts/plotting/tape_recorder.py:183` and
  `scripts/plotting/regional_map_multicase.py:205` still hand-roll their own globs
  and remain exposed to #433. One line each; pre-existing debt, not required here.
- MDTF is only half fixed. `lib/adf_diag.py` still takes `adf_file_list[0]` and
  still prints "Continuing with best guess" when more than one file survives, which
  is now the normal outcome for a chunked set.
- Nothing told the user that files had been passed over. When two sets are present,
  a line naming the set actually used saves real confusion about why a number
  changed between runs.
- The efficiency rationale in the docstring is not always realized: with chunks plus
  a whole-period file present, years 1-15 select the 40-year file rather than the
  two chunks. The answer is still correct, since the climatology step slices by
  year; it is just more reading than the comment promises.
- New comments follow the file's `#Comment` idiom, which `black` would rewrite to
  `# Comment`. Internally consistent, so take it or leave it.

### Checked and found sound

- The #433 layout genuinely resolves, for both a full and a partial requested range.
- Chunked (GenTS `slice_years`) sets are preserved, including sub-ranges and chunks
  whose year spans touch at the boundary because of end-of-interval naming.
- Every documented pass-through case behaves as claimed: fewer than two files,
  missing or empty years, unreadable names, a gap in the range, a backwards range.
  `Path` objects and string years both work, and year 0 is not swallowed by the
  falsy check.
- Index alignment, where an off-by-one was most likely:
  `climo_yrs["syears"][caseindex]` in `AdfData` is indexed by the same list
  `AdfInfo` builds the years from; `amwg_table` appends the baseline to case names
  and both year lists together; the MDTF loop enumerates the same case list. The
  years are always integers by the time they are read, so `int()` cannot raise from
  any in-tree caller.
- Reading the available period off the file names in `get_climo_yrs_from_ts` agrees
  with the old data-reading path for every realistic naming I could construct,
  including the end-of-interval trap the old `time_bnds` averaging existed to avoid.
  The fallback to opening the files is retained and tested.
- Selection happens before the `spawn` multiprocessing pool, so nothing unpicklable
  crosses into the workers.
- The five corrected test file names are right: `_write_ts(..., 100, 5)` writes years
  100-104, so `010001-010412.nc` is correct and the old `010512` was wrong.

### Could not confirm

- The end-to-end Casper runs and the regression sweep. I did not run the ADF on real
  data.
- The 78 to 84 test count; I measured 84 on the branch but did not rebuild the
  baseline on `main`.
- The partial-overlap case is not fully fixed when the config gives no years:
  `get_climo_yrs_from_ts` reports the union span, both files are then needed to
  cover it, and the same `ValueError` reappears — moved from setup to the
  climatology step, where it is at least caught per variable.
- Not examined: LDF, the MDTF path beyond the changed lines, and the visual outcome
  of any plot.

---

### Addendum: the three blocking items are addressed in `196d95b`

- Files that already combine cleanly are now returned untouched, so the full-record
  plots keep every file and there was never anything to resolve for them.
- What gets dropped is now checked against the requested period at the resolution
  the file names use, so halves of a year and mid-year set boundaries are left
  alone, while a file that merely runs on past the requested years is still dropped
  safely.
- The chosen set is checked for overlap before it is returned, so the 25-year plus
  chunks layout comes back whole rather than unusable.
- `AdfData` now writes a debug log line naming the files it used whenever a
  directory held more than one set.

Unit tests are 89 and passing, including cases for each of the three findings. The
non-blocking points are not addressed and are left for follow-up.

brianpm-ucar (commented):
## Second review, of commits through `892682f`

One blocker, introduced by the most recent commit at the time of review. The
rest of the change held up under a fair amount of attempted breaking.

### Blocking

**The MDTF file move could end the whole ADF run.** `lib/adf_diag.py`, in
`move_tsfiles_for_mdtf`.

`select_ts_files` is built to hand back the files it was given when it cannot
resolve them, so that the caller keeps the behaviour it had before. That
contract did not hold for this caller. Before the change, a list of several
files took the first one, printed "Continuing with best guess", and copied it.
After the change, any list of more than one file was treated as consecutive
chunks and combined with `xr.open_mfdataset`.

Reachable input: a time series directory holding the partial-overlap layout
from #433, years 1-20 beside years 10-40, with `cam_ts_done: true`,
`mdtf_run: true`, and either no years given or years 1-40. The available range
is reported as years 1-40, the greedy walk produces an overlapping pair,
`select_ts_files` correctly declines to choose and passes both files through,
and the combine then raises `ValueError: Resulting object does not have
monotonic global indexes along dimension time`. Nothing catches it, so the run
ends before the climatologies are calculated.

`mdtf_run` defaults to false, so only MDTF users are affected, but framework
code should not end a run with a raw traceback.

### Non-blocking

- `_ts_file_years` in `lib/adf_file_utils.py` has no callers anywhere in the
  tree; an earlier revision left it behind.
- `_ts_file_spans` reads its "unreadable names" case through a sentinel tuple.
  The behaviour is identical to `main`, but a plain check on the parsed list
  says the same thing more clearly.
- The greedy walk still declines some layouts that could be covered. Over
  8,800 randomly generated overlapping-but-coverable layouts, 567 (6 per cent)
  were passed through although a valid non-overlapping cover existed. These are
  always pass-throughs, never wrong answers, and match what `main` does. Worth
  a separate issue.
- In the duplicate-set case only, the full-record plots do get a shortened
  record: a chunked archive of years 1-100 plus a duplicate whole-period file,
  with `start_year: 91`, hands `global_mean_timeseries` the single 91-100 chunk.
  Unavoidable, since the full set cannot be opened, but a user may expect 100
  years, so it deserves a message on screen rather than only in the debug log.
- The MDTF combine writes with no encoding, so the source file's compression is
  not carried over for the combined case. Global attributes, including
  `time_period_freq`, are preserved.
- The MDTF notice about a directory holding several files moved from default
  verbosity to a higher one, so an ordinary run no longer mentions it.

### Checked and found sound

- `select_ts_files` never loses requested data and never returns a set that
  cannot be opened. This was fuzzed against a model of each file's real time
  coverage: 20,000 `YYYYMM` layouts with deliberate mid-year starts and ends,
  8,000 daily `YYYYMMDD` layouts, and 8,000 annual `YYYY` layouts. No case lost
  a requested month or day that the full set held, and no narrowed result
  contained two overlapping files. Every attempt to build a lossy input --
  halves of a year, a set boundary inside a year, a kept file that stops in June
  of the last requested year, a 25-year file beside 10-year chunks -- was caught
  and turned into a pass-through.
- The #433 layout resolves through the real data access code, not only the unit
  tests: driving `AdfData.get_timeseries_file` and
  `get_ref_timeseries_file` against a directory holding years 1-20, 10-40 and
  1-40 returned the single correct file for years 1-20, with and without
  `hist_str`, and left a cleanly chunked variable untouched.
- The MDTF combine does work for the case it was written for: two consecutive
  chunks combined to 240 monthly steps with a continuous time axis and the
  source global attributes intact.
- Index alignment wherever years are looked up, in `AdfData`, `amwg_table`,
  `tape_recorder`, the MDTF loop, and `regional_map_multicase`.
- `qbo` appending the baseline year to the list it got from `climo_yrs` cannot
  corrupt framework state: that property returns a copy on every access.
- Empty-string years, which is what the baseline years are under `compare_obs`,
  short-circuit correctly. Year zero is not mistaken for a missing year.
- Selection happens before the multiprocessing pool in `create_climo_files`,
  and the years used for selection are the same ones used for the time slice.
- Reading the available period from the file names agrees with the old
  data-reading path for every realistic naming tried, including CAM's
  end-of-interval stamping; where they differ, the new answer is the better one.
- Unit tests: 89 pass on the branch against 71 on `main`. Pylint on the six
  linted framework files stays above the 9.5 gate. The new selection tests live
  in the test file that imports only `pathlib`, so they really do run in CI.
- No new configuration keys, no new dependencies, no hard-coded paths, no
  changes to script entry points or the website.
- The three pre-existing bugs in #470, #471 and #472 were confirmed to exist on
  `main` and are not touched by this change.

### Could not confirm

- The end-to-end runs on Casper and the regression sweep. These were not
  repeated; that needs real data on the HPC system.
- That `qbo`'s change is compile-checked only. Consistent with what is there,
  but it could not be exercised either, since no `U` time series was available.
- Not examined: LDF, the CVDP path, GenTS runtime behaviour, the visual outcome
  of any plot, and `move_tsfiles_for_mdtf` beyond the changed lines and the
  failure traced above.

---

### Addendum: fixed in `904643e`

- Files are combined for MDTF only when they do not overlap, which is the case
  the combine was written for. A combine that fails for any other reason now
  falls back to copying one file instead of raising: MDTF getting part of a
  record is better than the run ending. Overlapping files keep the older
  behaviour of copying the first and warning about it, and that warning is back
  at default verbosity.
- `_ts_file_years` is removed, and `_ts_file_spans` now says plainly what it
  does with unreadable names.
- The message about files that cannot be used together is printed as well as
  logged.

The remaining non-blocking points are left as they are: the greedy walk's
declined layouts deserve their own issue, and the MDTF combine's encoding is
left alone for now.

PR #473, opened 2026-09-03
Add Black CI workflow.

nusbaume:
Adds a `black` formatter Github Action workflow that checks if any python code under the `lib` directory is following the black formatting style, and if not then generates a CI failure. 

Also updates some of the actions and python versions used in the `ADF_Linting.yaml` workflow file.


.github/workflows/ADF_linting.yaml
```diff
@@ -36,3 +36,27 @@ jobs:
...
+        python-version: '3.12' # Use python 3.12
+    # install black, pinned so that new black releases (which can
+    # change the "stable style") don't break CI unexpectedly:
+    - name: Install black
+      run: |
+        python -m pip install --upgrade pip  # Install latest version of PIP
+        pip install 'black==25.11.0'         # Install Black python package
+    # check the formatting of all python code under "lib":
+    - name: Run black on ADF lib
+      run: black --check --diff lib
```
nusbaume: `black --check --diff lib` also walks `lib/externals/`, which holds vendored upstream code (CVDP, brought in wholesale by 16cdfc7). Commit e47c438 consequently reformatted `lib/externals/CVDP/ncl_scripts/runTasks.py`, which is now permanently divergent from upstream CVDP. Concrete failure: the next time someone re-syncs CVDP from its upstream repo, the freshly copied `runTasks.py` will be unformatted and this job fails CI, forcing every external refresh to be re-blackened by hand and making `diff` against upstream noisy forever.

Suggest excluding the externals tree so black only governs ADF-authored code:

```suggestion
      run: black --check --diff --extend-exclude '^/lib/externals/' lib
```

.github/workflows/ADF_linting.yaml
```diff
@@ -36,3 +36,27 @@ jobs:
...
+    - name: Set up Python 3.12
+      uses: actions/setup-python@v6
+      with:
+        python-version: '3.12' # Use python 3.12
+    # install black, pinned so that new black releases (which can
+    # change the "stable style") don't break CI unexpectedly:
+    - name: Install black
+      run: |
+        python -m pip install --upgrade pip  # Install latest version of PIP
+        pip install 'black==25.11.0'         # Install Black python package
```
nusbaume: The pin lives only inside this workflow: the repo has no `pyproject.toml` `[tool.black]` section and `.pre-commit-config.yaml` still contains only `check-yaml`. So nothing tells a contributor which black to run, and nothing enforces it before push. Concrete scenario: a developer with the `black` already in their conda env (any version other than 25.11.0 — the 2024 and 2026 stable-style releases both changed output) runs `black lib`, gets a clean local result, pushes, and CI fails with a diff they cannot reproduce; worse, "fixing" it by re-running their own black re-churns unrelated files. Recommend adding a `[tool.black]` block to a repo-root `pyproject.toml` (documenting line-length/target-version) and a `psf/black` pre-commit hook pinned to `25.11.0` so the pre-commit job and this job agree.

.github/workflows/ADF_linting.yaml
```diff
@@ -36,3 +36,27 @@ jobs:
         ACCESS_TOKEN: ${{ secrets.GITHUB_TOKEN }}
       run: .github/scripts/pr_mod_file_tests.py --access_token $ACCESS_TOKEN --pr_num $PR_NUMBER --rcfile lib/test/pylintrc --pylint_level 9.5
 
+  #This job is designed to run the "black" formatter against
+  #every python file under the ADF "lib" directory.  Black is
+  #run in "--check" mode, which makes no modifications but
+  #exits with a non-zero status (and thus fails CI) if any
+  #file is not formatted the way black would format it.
```
nusbaume: Coverage gap worth calling out explicitly: `lib` is only about half the Python in the repo. `scripts/{analysis,averaging,plotting,regridding}` plus `.github/scripts` hold ~25 more `.py` files, and the top-level `run_adf_diag` entry point is Python without a `.py` suffix (black would not pick it up even if the path were passed). Those are ADF-authored, user-facing code, so the tree stays permanently half-formatted and any future attempt to extend the job to `scripts` will land as another giant reformat commit. If the split is intentional, a note here saying `scripts` is deliberately deferred would save the next reader the check.

.github/workflows/ADF_linting.yaml
nusbaume: This comment is from Claude.

.github/workflows/ADF_linting.yaml
nusbaume: This comment is from Claude.

.github/workflows/ADF_linting.yaml
nusbaume: This comment is from Claude.

nusbaume:
## Review of PR #473 — "Add Black CI workflow."

*Review performed by Claude (Opus 5), following the guidance in `AGENTS.md`.*

**Summary.** The PR does what it says: it adds a pinned `black` hook to `.pre-commit-config.yaml` (making `ADF_pre-commit.yaml` the single CI gate for formatting), adds a repo-root `pyproject.toml` holding black's settings, modernizes the actions/Python versions in `ADF_linting.yaml` and `ADF_pre-commit.yaml`, and reformats all 29 black-eligible files under `lib/`. The reformat was verified as mechanical rather than taken on trust: the whole `lib/` diff is exactly what `black 25.11.0` produces from `main` under this PR's config, apart from three duplicated comment lines. It looks ready to merge once those three lines and the stale `AGENTS.md` §6 text are cleaned up. No blocking findings.

### What was run

| Check | Result |
| --- | --- |
| `black --check lib/` | 29 files unchanged; `lib/externals/` correctly skipped |
| `pre-commit run -a --show-diff-on-failure` | `check-yaml` Passed, `black` Passed (real hook at `rev: 25.11.0`) |
| `force-exclude` behavior | Skips `lib/externals/.../runTasks.py` both when handed explicitly (the pre-commit path) and when invoked from inside `lib/` — the `force-exclude`-vs-`extend-exclude` comment in `pyproject.toml` is correct, and `runTasks.py` is byte-identical to `main` |
| Reproduce the reformat: `git archive main` → `black` with this PR's `pyproject.toml` → `diff -r` vs the PR's `lib/` | Identical except the three duplicated comments below. No hidden functional edits anywhere in the 3671/2598 diff |
| AST equality vs `main`, all `lib/*.py` | Equivalent; the only string-constant differences are black's docstring re-indentation (3 docstrings in `adf_dataset.py`, 2 in `test_adf_config.py`) |
| pylint @ 9.5 on all six `testable_files` | All pass, and every score rose or held: `adf_base` 9.52→10.00, `adf_config` 9.52→9.76, `adf_info` 9.54→9.64, `adf_obs` 9.86→10.00, `adf_web` 9.57→9.60, `adf_diag` 9.79→9.85 (pylint 4.0.8) |
| `ast.parse(..., feature_version=(3,9))` on all `lib/*.py` | No failures — confirms the unset `target-version` reasoning in `pyproject.toml`; nothing black emitted needs >3.9, which matters since unit tests run on 3.9 |

### Non-blocking

1. **Three duplicated comment lines — merge-resolution artifacts, and the only non-mechanical change in the diff.** Removing them makes the `lib/` diff provably 100% black output:
   - `lib/adf_utils.py:253-256` — `# End def` twice
   - `lib/adf_dataset.py:296-298` — `# ------------------` twice (`main` has one, closing the previous section before the `# Climatology files` header)
   - `lib/test/unit_tests/test_adf_file_utils.py:493-495` — `# ++++++++++++++++++` twice

2. **`AGENTS.md` §6 is now stale and contradicts §3, which this PR did update.** `AGENTS.md:193-196` still says black formatting is a convention "not yet enforced by CI" and that black "is [not] in ... `.pre-commit-config.yaml`"; `AGENTS.md:202` still says "The existing tree is not black-clean." All three are false after this PR for `lib/`. §6.1's `darker` / "only report hunks that overlap changed lines" advice is still right for `scripts/`, but for `lib/` the check is now simply `black --check` (or `pre-commit run -a`) and a formatting failure there *is* a CI failure. Worth splitting §6.1 into the `lib/` (enforced) and `scripts/` (convention) cases.

3. **`.github/workflows/ADF_unit_tests.yaml` was left on `actions/checkout@v3` / `actions/setup-python@v4`.** The PR modernized the other two workflows to `@v6`; finishing the third keeps the three consistent and off the deprecated-node path. Its Python matrix (3.9–3.13) is fine as-is.

4. **Add a `.git-blame-ignore-revs`.** A 29-file reformat makes `git blame` on all of `lib/` point at this PR indefinitely. The file needs the post-merge SHA, so it has to be a follow-up commit (and it is simpler if this PR is squash-merged to a single SHA). Pair it with a wiki/README line on `git config blame.ignoreRevsFile .git-blame-ignore-revs`.

5. **`black` is still absent from `env/conda_environment.yaml`.** Contributors now need `black` (or `pre-commit`) locally to pass CI. Either add `pre-commit` to the dev environment or document `pip install pre-commit && pre-commit install` on the wiki — otherwise the first failure people see is a red CI check with no local reproduction available in the ADF env.

### Optional / follow-up (pre-existing)

- `lib/plotting_functions.py` and `lib/adf_utils.py` contain non-raw strings with invalid escapes (`\m`, from LaTeX in docstrings), which emit `SyntaxWarning` on 3.12+ and are slated to become `SyntaxError`. Unit tests already run on 3.13, so this will eventually bite. This is exactly `AGENTS.md` §6.2's raw-string rule; it is pre-existing debt, not this PR's job — an issue is the right home.
- `.pre-commit-config.yaml` still pins `pre-commit/pre-commit-hooks` at `v4.3.0`. Unrelated to this PR, but a natural thing to bump the next time that file is touched.

### Not verified

- `pytest lib/test/unit_tests` — `pytest` was not installed in the environment used for this review. The test-file changes are black-only (confirmed by the AST check above), so CI should be green, but the suite was not run locally.
- Whether the pinned `rev: 25.11.0` resolves in a clean GitHub-hosted environment. It resolved locally through `pre-commit`, and `black 25.11.0` is a real release, so this is very likely fine — the PR's own CI run is the authoritative check.
- No ADF run was performed. None is needed: the change is provably formatting-only under black's AST-safety guarantee plus the byte-level reproduction above.

richling-ucar:
@nusbaume All works as intended on my test run.

PR #480, opened 2026-09-04
Report progress while the time series files are written

### LINKED ISSUE #442: Time Series creation timeout

brianpm-ucar:
Closes #442.

### The problem

`create_time_series` prints one line per variable while it assembles the NCO
commands, and then goes completely silent inside three blocking `Pool.map`
calls while every `ncrcat` actually runs. With a long `diag_var_list` that is a
stretch of dead time with no terminal output at all — long enough for an ssh
session to hit its idle timeout and take the run down with it, which is what
#442 reports.

### The change

Consume the pool results with `imap_unordered` instead of `map` and print a
line as each command finishes:

```
	     ncrcat: 7 of 10 done
	     global attributes: 7 of 10 done
	     history attribute: 7 of 10 done
```

Same parallelism, same `num_procs`, and the results were being discarded in
both cases, so nothing about the work changes — the completions just get
reported as they land instead of all at once at the end. `flush=True` so the
progress still shows up when stdout is a pipe or a log file rather than a tty.

No new dependencies, and no attempt to keep a session alive: a long run still
belongs in a batch job or under `tmux`/`screen`. This only makes the silence
visible.

### Testing

Ran a time-series-only model-vs-model config (10 variables including derived
`PRECT`, 2 years, `num_procs: 4`) over two dissimilar cases — a CESM3 alpha
1850 control on `cam.h0a` and an AMIP run on `cam.h0` — in a Casper batch job
with stdout piped rather than a tty, so the flushing was exercised too.

- progress lines appear for all three phases, 1..10 in completion order
- 10 time series files per case, 24 time steps each, derived `PRECT` present
- `adf_user` / `hist_file_path` / `hist_file_list` / `time_period_freq` global
  attributes set, and the `history` attribute removed, i.e. the two `ncatted`
  phases still do their job in the right order

### DIFF
--- a/lib/adf_diag.py
+++ b/lib/adf_diag.py
@@ -414,6 +414,35 @@ class AdfDiag(AdfWeb):
             return subprocess.run(cmd, shell=False)
         # End def
 
+        def run_pool(commands, label):
+            """Run NCO commands in parallel, reporting each one as it finishes.
+
+            Results are consumed with ``imap_unordered`` rather than ``map`` so
+            that the progress lines appear while the commands are still running.
+            Without them the terminal sits silent for the whole batch, long
+            enough for an ssh session to time out on a large ``diag_var_list``.
+
+            Parameters
+            ----------
+            commands : list
+                NCO commands, each an argument list for :func:`call_ncrcat`.
+            label : str
+                Name of this batch of commands, used in the progress output.
+
+            Returns
+            -------
+            None
+                Writes the files produced by ``commands``.
+            """
+            ntot = len(commands)
+            with mp.Pool(processes=self.num_procs) as mpool:
+                for num, _ in enumerate(mpool.imap_unordered(call_ncrcat, commands), 1):
+                    print(f"\t     {label}: {num} of {ntot} done", flush=True)
+                # End for
+            # End with
+
+        # End def
+
         # Gather the per-case configuration (shared with the GenTS back end):
         cfg = self.get_ts_case_config(baseline=baseline)
         case_names = cfg["case_names"]
@@ -719,17 +748,14 @@ class AdfDiag(AdfWeb):
                 # End variable loop
 
                 # Now run the "ncrcat" subprocesses in parallel:
-                with mp.Pool(processes=self.num_procs) as mpool:
-                    _ = mpool.map(call_ncrcat, list_of_commands)
+                run_pool(list_of_commands, "ncrcat")
 
                 # Run ncatted commands after ncrcat is done
-                with mp.Pool(processes=self.num_procs) as mpool:
-                    _ = mpool.map(call_ncrcat, list_of_ncatted_commands)
+                run_pool(list_of_ncatted_commands, "global attributes")
 
                 # Run ncatted command to remove history attribute
                 # after the global attributes are set
-                with mp.Pool(processes=self.num_procs) as mpool:
-                    _ = mpool.map(call_ncrcat, list_of_hist_commands)
+                run_pool(list_of_hist_commands, "history attribute")
 
                 # Finally, run through the derived variables if applicable
                 if constit_dict:


brianpm-ucar:
## Self-review against `AGENTS.md`

*This is a Claude-assisted review of my own PR, run against the draft `AGENTS.md` in the repo root — partly to check the change, partly to see how well that file works as a review guide. Posting it for transparency, not as a substitute for a human reviewer.*

A 32-line, single-file change to `lib/adf_diag.py`: three silent `Pool.map` calls in `create_time_series` become an `imap_unordered` helper that prints as each command completes. Framework (`lib/`) change with no science, no config keys, and no interface surface, so most of §4 doesn't apply. It satisfies `AGENTS.md` now, but two things had to be fixed to get there — both amended in (`58b50d0` → `a5abab2`).

### Fixed during review

- **§6.1 (black on new code)** — the original hand-wrapped `enumerate(mpool.imap_unordered(call_ncrcat,\n<spaces>commands), 1)` is not what black produces. `darker --check --diff --revision upstream/main` now reports zero changes on the added lines. The single-line form lands at exactly 88 columns, inside both black's target and pylint's `max-line-length=100`.
- **§6.2 (Sphinx-ready NumPy docstrings)** — `run_pool` had free prose only. It now carries the one-line summary, blank line, and correctly underlined `Parameters` / `Returns` sections in `name : type` form, matching the style used elsewhere in `lib/`.

### Contracts checked

- **§4.5, parallelism** — the only object pickled is the existing module-level `call_ncrcat`; `run_pool` is a closure that stays in the parent process. No shared mutable state, and each worker still writes its own file. `imap_unordered` changes only the order in which results are *yielded*, and they were discarded before, so nothing depends on that order.
- **§4.5, error handling** — unchanged. `subprocess.run` without `check=` still swallows a failing `ncrcat` exactly as it did before; pylint already flags it (`W1510` at `lib/adf_diag.py:414`). Noting it as pre-existing debt rather than expanding this PR's scope, per §5.
- **§4.1 / §4.2 / §4.3 / §4.4** — not applicable: no stage script, no data access, no new config key, no `add_website_data` call.

### §3 checks, reproduced locally

| Check | Result |
| --- | --- |
| `pytest lib/test/unit_tests` | 146 passed |
| `pylint --rcfile=lib/test/pylintrc lib/adf_diag.py` | **9.79/10** (gate is 9.5) |
| `python -m py_compile` on changed files | clean |
| `config_cam_baseline_example.yaml` parses | yes |

I also linted `upstream/main`'s copy of the file and diffed the message counts: identical apart from five `E0401`s that are an artifact of linting it outside the tree. So no new pylint message. The two warnings adjacent to the change (`W0601` at :406, `W1510` at :414) sit on untouched pre-existing lines.

### One deliberate deviation, flagged for a reviewer

§4.5 says *"Debug output goes through `adfobj.debug_log(...)`, not stray `print()` in library code"* — and this adds `print()` to library code. That's intentional: `debug_log` only emits under `--debug`, so routing the progress there would leave a default run exactly as silent and wouldn't fix #442 at all. This is user-facing progress output rather than debug output, and the surrounding method already reports its progress with `print()`. Happy to revisit if a reviewer reads that rule more strictly.

### Not verified

Anything about behavior at real scale — the end-to-end run reported in the PR description used 10 variables and 2 years, which is enough to exercise all three command batches and the derived-variable path, but it is not a stress test of a long `diag_var_list` over decades. The claim being made is narrow: the progress lines appear as the work happens, and the output files are unchanged.

PR #485, opened 2026-09-19
Allow "all" in diag_var_list to process every input variable

### LINKED ISSUE #208: Allow ADF to process all variables present in input files
### New feature type

New infrastructure or infrastructure enhancement

### What is this new feature?

Currently a user must list all of the variables that they will want the ADF to process from the case's history or time series files in the config file, which can be difficult if there are hundreds of variables to analyze.  Thus it would be helpful if a user could add a special character or entry to the `diag_var_list` config variable to indicate that the ADF should find and process all variables within the provided input files.

### Assistance required?

### Extra info

This feature has been explicitly requested by @cecilehannay

brianpm-ucar:
Closes #208.

Listing hundreds of variables by hand is what stands between a user and running the ADF over everything a case wrote, so `diag_var_list` now accepts a single `all` entry and works the list out from the input files instead:

```yaml
diag_var_list:
    - all
    - PRECT   # kept: derived, in no input file
```

### How it works

`diag_var_list` is read in exactly one place (`AdfInfo.__init__`), so `all` is resolved there and every other module goes on seeing an ordinary list of variable names.

- **Pre-made time series** (`cam_ts_done: true`): the variable is in the file name, so the names are parsed rather than the files opened.
- **History files**: `data_vars` of the first file of each configured stream — the same "every file in a stream holds the same variables" assumption the time series step already makes.
- Only variables with **both a time dimension and a horizontal one** (`lat`/`lon`/`ncol`/`lndgrid`) are kept, which leaves out the coefficients, counters and weights sitting alongside the fields (`hyam`, `date`, `gw`, `nsteph`, ...).
- The union over all test cases is used. Variables listed alongside `all` are kept, so a derived variable that is in no input file can still be asked for. Nothing found is an error, not an empty run.

Only the test case files are searched: a baseline case missing one of the variables is already handled downstream, exactly as when the user lists a variable by hand, and against observations there is no baseline to search.

### Testing

- 10 new unit tests. The file-name parser lives in `adf_file_utils` (no xarray), so half of them run in CI; the rest skip there, as `test_adf_info_climo_yrs` already does.
- Full unit test suite: 187 passed.
- Real run against a BHIST `330_branch` archive: `all` resolved to **404 variables** in ~11 s, clutter dropped, explicitly listed `PRECT`/`SST` kept.


brianpm-ucar:
Reviewed this and pushed four corrections in f0ea024d.

**Fixed**

- **A missing `cam_ts_loc` crashed with a traceback.** With `cam_ts_done: true` and no `cam_ts_loc`, the time series branch reached `Path(None)` and raised `TypeError: argument should be a str or an os.PathLike object ... not 'NoneType'` from inside `AdfInfo.__init__` — before the framework's own `required=True` read of that key at `adf_info.py:523` could report it properly. Guarded, as the history branch already was.
- **The time series search ignored `hist_str`.** It globbed `{case}*.nc`, so a `cam_ts_loc` holding more than one stream contributed variables from streams the run does not use, which the rest of the run then warn-skipped one by one. It now anchors on `{case}.{hist_str}.*.nc` per configured stream, matching both the history branch and `AdfData.get_timeseries_file`, and falls back to the old pattern when no stream is configured.
- **An unusable input directory looked like an empty case.** A missing or unreadable `cam_hist_loc`/`cam_ts_loc` fell through to the generic "No variables were found ..." message. It now goes through `describe_dir_problem()` and names the actual problem, the same as `create_time_series` does.
- **`expand_var_list_all` is private now** (`_expand_var_list_all`). Its only caller is `__init__`, and as a public method it added a 21st public method to `AdfInfo`, tripping pylint's `too-many-public-methods`.

`"all"` was already matched with `.lower()`, so `ALL` and `All` work too; that now has a test and a comment saying so.

**Checks**

- `pytest lib/test/unit_tests`: **191 passed** (was 187 — four new tests: the missing-`cam_ts_loc` case, stream anchoring for time series, the unusable-directory error, and the upper-case entry).
- `pylint --rcfile=lib/test/pylintrc lib/adf_info.py`: **9.90** (was 9.88; `main` is 9.46).
- `black --check lib/`: clean.
- Real data: the BHIST `330_branch` archive still resolves to **404 variables** in 3.4 s, and a real ADF `ts` directory gives the same 20 variables anchored or unanchored, and nothing for a stream that is not there.

PR #487, opened 2026-09-19
Apply 'mask: ocean' in compare_obs mode, and drop the undeclared bottleneck dependency

### LINKED ISSUE #483: compare_obs: 'mask: ocean' never applies (SST stats invalid), and undeclared bottleneck dependency crashes Taylor diagrams
Two problems found while running a model-vs-obs comparison (`compare_obs: true`) on a
CESM3 BHIST case. Both reproduce on `main` (35d758d9). I intend to address both in a
PR later — opening this to record the diagnosis.

---

## 1. `mask: ocean` never applies in a model-vs-obs run, silently corrupting SST

`SST` is the only variable in `lib/adf_variable_defaults.yaml` carrying `mask: ocean`,
and that mask can never be applied when comparing against observations.

The chain:

1. `scripts/regridding/regrid_and_vert_interp.py:56-59` skips any variable that has no
   reference dataset:

   ```python
   if var in adf.data.ref_var_nam:
       target_name = adf.data.ref_labels[var]
   else:
       print(f"\t ERROR: No reference data available for {var}.")
       continue
   ```

2. `OCNFRAC` has no `obs_file` in the shipped defaults, so in `compare_obs` mode it
   always takes that `continue` and `<target>_<case>_OCNFRAC_regridded.nc` is never
   written.

3. The masking block at `regrid_and_vert_interp.py:125-131` then can't find that file
   and degrades to a warning:

   ```
   ERROR: No reference data available for OCNFRAC.
      WARNING: OCNFRAC not found, unable to apply mask to 'SST'
   ```

So `SST` is written and plotted completely unmasked. ADF reports success.

### Why this is more than cosmetic

The shipped `SST` default also points at ERAI `TS`:

```yaml
SST:
  obs_file: "ERAI_all_climo.nc"
  obs_name: "ERAI"
  obs_var_name: "TS"
  mask: "ocean"
```

`TS` is skin temperature over the whole globe, not SST — it has valid values over land
(no NaNs). CAM's `SST` field is ~0 K over land. With the mask absent, the difference
is roughly -290 K across every continent, and the summary statistics printed on the
figure are meaningless:

| | ANN, 1979-2010 |
|---|---|
| RMSE | 139.663 K |
| Mean bias | -72.62 K |
| Test panel mean | 215.67 K |

Over ocean the fields are fine (a few K of Southern Ocean warm bias), so the figure
looks plausible at a glance while every number on it is wrong. That is the part I find
most concerning — nothing in the output flags it beyond one easily-missed warning.

### Reproducer

Any `compare_obs: true` run with `SST` in `diag_var_list`.

### Possible fixes

Two separate things are wrong, and I think both want addressing:

- **The mask.** The mask uses the *model's* `OCNFRAC` regridded to the reference grid;
  the reference values themselves are never used, only the target grid. So requiring a
  reference dataset for `OCNFRAC` is an artificial constraint. Options: regrid mask
  variables to the target grid without demanding a reference, derive `OCNFRAC` from the
  `LANDFRAC` obs that already ships, or at minimum promote the warning to something that
  can't be missed when a requested mask silently doesn't apply.
- **The reference variable.** `ERAI_all_climo.nc` contains `SSTK_SFC` (a real SST field,
  already NaN over land) alongside `TS`. Pointing `SST` at `SSTK_SFC` looks more correct
  regardless of how the mask issue is resolved. The default `contour_levels_range` of
  `[220, 320, 5]` is also tuned for `TS`; real SST spans roughly 270-306 K.

### Workaround in the meantime

A local defaults file with `SST` -> `obs_var_name: SSTK_SFC` and an `OCNFRAC` entry
pointing at a hand-built ocean-fraction file on the reference grid (with `obs_name`
matching SST's, so the regridded filename lines up with what the masking step looks for).

---

## 2. Undeclared `bottleneck` dependency crashes the Taylor diagram

`scripts/plotting/cam_taylor_diagram.py:510`:

```python
return diff.fillna(edge_diff.bfill(dim).ffill(dim))
```

xarray's `bfill`/`ffill` require `bottleneck` (or `numbagg`); neither is in
`env/conda_environment.yaml`. In an environment without it, every Taylor diagram dies:

```
File "scripts/plotting/cam_taylor_diagram.py", line 486, in get_vertical_average
    weights = calculate_thickness_approx(ds[level_dim], dim=level_dim)
File "scripts/plotting/cam_taylor_diagram.py", line 510, in calculate_thickness_approx
    return diff.fillna(edge_diff.bfill(dim).ffill(dim))
...
File ".../xarray/core/duck_array_ops.py", line 877, in _push
    import bottleneck as bn
ModuleNotFoundError: No module named 'bottleneck'
```

This is an unhandled exception, so it takes the whole ADF run down at the Taylor stage
rather than skipping that one diagnostic.

It may be that the reference ADF conda env picks `bottleneck` up transitively and this
only bites people running in their own env — but either way it isn't declared.

### Possible fixes

Add `bottleneck` to `env/conda_environment.yaml`, or avoid `bfill`/`ffill` in
`calculate_thickness_approx` (the edge handling there is simple enough to do with
`isel`-based indexing and no optional accelerator).

brianpm-ucar:
Closes #483.

## 1. `mask: ocean` never applied in `compare_obs` mode

`SST` is the only variable in the defaults carrying `mask: ocean`, and the mask could
never be applied against observations: `OCNFRAC` has no obs counterpart, so
`regrid_and_vert_interp` skipped it, `<target>_<case>_OCNFRAC_regridded.nc` was never
written, and the masking block degraded to a warning. SST was plotted and scored
completely unmasked while ADF reported success.

The mask only ever needs the **model's** ocean fraction on the reference grid — the
reference OCNFRAC values are never used, only its grid — so requiring a reference was an
artificial constraint. New `_get_ocean_fraction()` uses the regridded file when it exists
and otherwise loads the case's OCNFRAC climatology and regrids it on the fly. Nothing
changes for a model-vs-model run, where that file already exists.

If OCNFRAC is genuinely unavailable, the variable is now **skipped with a warning**
instead of written unmasked: an unmasked SST is not a degraded result, it is a wrong one,
and it looks entirely plausible on the page. (`AdfObs` adds `OCNFRAC` to `diag_var_list`
on its own whenever any variable wants a mask, so this only comes up when the case has no
OCNFRAC output at all.)

## 2. `bfill`/`ffill` in the Taylor diagram

`calculate_thickness_approx()` used `.bfill().ffill()`, which xarray implements via
`bottleneck`/`numbagg`. Neither is in `env/conda_environment.yaml`, so the Taylor stage
died with an `ImportError` in a clean ADF environment. The edges are now filled by
position instead — no new dependency, and no dependence on the level coordinate being
present and monotonic.

**This also changes numbers.** The old chain was a no-op at the leading edge: `edge_diff`
is one element shorter than `diff`, so the first level's thickness stayed NaN. Because
`weighted_vertical_average()` sums with the default `skipna=True`, that NaN weight
silently dropped the top model layer out of every column average. So anyone who *was*
getting Taylor output (an environment with `bottleneck` present) will see
`ColumnTemperature` and `ColumnRelativeHumidity` shift slightly — one thin
top-of-atmosphere layer now included. The first level gets a real thickness
(`[nan, 150, 175, ...]` → `[150, 150, 175, ...]` on a 1000→50 hPa coordinate).

## 3. Taylor diagram: read each field once, not once per season

The season loop is outermost, so every entry was retrieved five times — and retrieving
`ColumnTemperature`/`ColumnRelativeHumidity` re-reads the whole 93-level regridded
`T`/`RELHUM`, for the model *and* the reference. Only the time mean differs between
seasons, so the retrieval is memoized. Happy to split this into its own PR if you would
rather keep the bug fix clean.

## 4. SST defaults

`adf_variable_defaults_era5-1deg.yaml` still pointed `SST` at ERAI `TS` — skin
temperature, valid over land — so the reference panel of an "SST" plot carried land
values. It now points at HadISST/OI.v2, the same 1° product the main defaults file uses.

Both files also get SST contour levels that suit an ocean-only field:
`[220, 320, 5]` dates from the comparison against skin temperature and spent most of its
colorbar on temperatures the ocean never reaches. HadISST/OI.v2 runs 272.0-307.3 K
monthly (273.3-303.2 K annual mean) and the CESM3 BHIST case runs 271.5-305.1 K, so
`[272, 308, 2]` gives 18 contours over 272-306 K — seawater freezes near 271.4 K, so
nothing physical falls below the first level, and the top leaves room for a warmer model.

## Testing

Model-vs-obs run (BHIST `330_branch`, 1979-2010, ERAI/ERA5/CERES obs) on Casper with the
**stock** `lib/adf_variable_defaults.yaml`, compared against a previous run of the same
case that used a hand-built OCNFRAC obs file as a workaround:

```
this PR      (12,121,240)  nan frac 0.502  min 272.24  max 302.37
workaround   (12,121,240)  nan frac 0.502  min 272.24  max 302.37
max |diff| : 0.0      nan pattern identical: True
```

For reference, before the fix the same SST figure reported RMSE 139.7 K / bias -72.6 K
(unmasked); masked it is 1.26 K / -0.55 K.

Taylor stage: 18:26 → under 3 minutes, with all five diagrams pixel-for-pixel identical
before and after the memoization (`(a == b).all()` on the PNG arrays, max pixel
difference 0). The positional edge fill was checked against the previous `reindex_like`
version on a 1-D level coordinate, on a coordinate-less and a non-monotonic `lev` (both
of which `reindex_like` raises on), and on a 3-D pressure field.

Not verified: no environment without `bottleneck` was available here, so the original
`ImportError` itself was only observed before the change, not its absence after. The
warn-and-skip path for a genuinely missing OCNFRAC was not exercised end to end — every
real run gets OCNFRAC added automatically.


brianpm-ucar:
Pushed a second commit (a01a8d0f) to this branch: a performance fix in the same file.

`cam_taylor_diagram` loops season outermost, so each of the ten diagram entries was
retrieved five times — and retrieving `ColumnTemperature` / `ColumnRelativeHumidity`
re-reads the whole 93-level regridded `T`/`RELHUM`, for the model *and* the reference.
Only the time mean differs between seasons, so the retrieval is now memoized and the
(2-D) result held.

On the 1979-2010 model-vs-obs run above, the Taylor stage went from **18:26 to under
3 minutes**, and all five diagrams come out pixel-for-pixel identical to the ones from
the previous run (`(a == b).all()` on the PNG arrays, max pixel difference 0).

Happy to split this into its own PR if you would rather keep the bug fix clean.

Issue #199, opened 2022-07-14
Vertical interpolation should always try to use pressure (PMID) when possible


nusbaume:
### New feature type

New regridding method or regridding enhancement

### What is this new feature?

It was raised by @JulioTBacmeister that when interpolating to the default ADF pressure levels, the ADF should just try to use the model mid-level pressure (PMID) when possible, as most of the CAM dycores are now capable of outputting it as a history variable.  Then only if pressure is unavailable should it try to use alternative methods (e.g. converting hybrid or height coordinates to pressure levels). 

Before this is done, however, the ADF should modify it's current handling of PMID such that it is only output once, instead of for every 3-D variable.

### Assistance required?

No, I will make a PR when the feature is ready

### Extra info

Implementing this feature might also benefit from having the ADF determine what the vertical coordinate of the model data is near the start of the ADF run itself (such as in the upcoming `AdfInfo` class initialization).

brianpm:
I'm going to add on to this issue slightly based on input from the AMWG dev group in this issue: https://github.com/NCAR/amwg_dev/issues/211

For models that have higher tops than the default pressure levels we're using in ADF, we should provide an option to either: (a) use user-specified pressure levels for the interpolation or (b) try to use the model's hybrid coefficients (or similar) to define levels.

In either case, those values would be passed as the kwarg `new_levels` to either `pmid_to_plev` or `lev_to_plev` from the regridding script (e.g., `regrid_and_vert_interp.py`). 

When `PMID` is available, maybe we'd have to do a global average to get levels? For hybrid, we could use Adam's suggestion of `P0*hyam + P0*hybm` as a rough guide. In the case of hybrid-sigma levels, there's a point (around 90hPa) where the levels become pure pressure levels, so maybe we could instead construct the average pressure field `P0*hyam + PS*hybm` and do a global average; that would provide the actually correct pressure levels above the cutoff, and would be appropriate for log(p) plots, I think.

brianpm-ucar:
Status check on this, since part of it has landed since it was opened.

## Done

`b7da2de8` ("Prefer the model's own pressure field for vertical interpolation", on `main`)
implements the core request in `scripts/regridding/regrid_and_vert_interp.py`:

- `_find_pressure_field()` (line 209) returns `PMID` for midpoint data and `PINT` for
  interface data, taken from the variable's own climo dataset first and from a standalone
  climo second, and `None` when neither is there.
[...]

PR #488, opened 2026-09-20
Make the model's 3-D pressure field once, and find it by more than its name (#199)

brianpm-ucar:
Part 1 of #199.

The consumer side of that issue landed in `b7da2de8`: vertical interpolation prefers the
pressure the model wrote over reconstructing it from `PS` and the hybrid coefficients.
Nothing arranged for that field to exist, though. `lib/adf_diag.py` added `PMID` to the
`ncrcat` variable list only when the vertical coordinate was `height`, and then copied it
into *every* 3-D variable's time series — the duplication the issue text asks to remove.
For hybrid CAM, the case the issue was raised about, `PMID` was never requested at all, so
the preference never fired unless a user added `PMID` to `diag_var_list` by hand.

## What this does

**Asks for the pressure field once.** `request_pressure_field` (in `adf_utils`) adds it to
`diag_var_list` when the history files have one, so it gets a single time series and a
single climatology per case instead of a copy inside every 3-D file. Both back ends do it —
`ncrcat` and GenTS. Only the vertical dimensions the run actually uses are considered, so a
run of midpoint fields does not also produce interface pressure.

**Stops assuming the name.** `find_pressure_field` resolves, in order:

1. the `pressure_field_names` config entry, a mapping of vertical dimension to name;
2. CAM's `PMID` / `PINT` as the default;
3. any variable on the right vertical dimension carrying `standard_name: air_pressure`.

So a model that writes `pfull` is found from its CF metadata with nothing configured, and a
model with neither CAM's names nor CF metadata can be named in the config.
`pressure_field_names: False` ignores the model's pressure field entirely and keeps the old
`PS` + hybrid path — the escape hatch for anyone who does not want the extra 3-D field.

**`PINT` gets `plot_diagnostics: False`**, as `PMID` already had: it is a support variable,
not a diagnostic anyone asked to see.

**Dimension order.** `pmid_to_plev` stacks and unstacks, which left `lev` leading, so a
file written through the pressure-field path did not look like one written through the
hybrid path. Transposed back to the source order; values are unchanged (verified, see
below).

## What changes for existing runs

A hybrid CAM run that writes `PMID` now interpolates with it instead of with
`P0*hyam + PS*hybm`. **This changes results**, and is the point of the issue: for the
dry-mass vertical coordinate in recent CAM/WACCM the hybrid coefficients do not give
pressure. On the CESM3 alpha case below, regridded `Q` moved by up to 2.2e-4 kg/kg against
a field maximum of 2.0e-2 — about 1%, largest at 925 hPa and negligible above 100 hPa. The
cost is one more 3-D time series and climatology per case, and `pressure_field_names: False`
declines it.

## Testing

`pytest lib/test/unit_tests` — 201 passed, including a new `test_pressure_field.py` (config
override, CF-metadata fallback, the `False` escape hatch, the missing-field fallback, and
that a request adds the field once and skips an unused vertical dimension).
`pylint --rcfile=lib/test/pylintrc lib/adf_diag.py` — 9.86/10. `black --check lib/` and
`darker` against `main` on `scripts/` are clean.

Two model-vs-model runs on Casper, one per time-series back end, with a test case that
writes `PMID`/`PINT` (CESM3 alpha, `cam.h0a`, 93 levels) and a baseline that writes only
`PS` (AMIP, `cam.h0`, 32 levels), so both branches are exercised in one run:

- `PMID` gets exactly one time series and one climatology per case under both back ends;
  `Q`'s time series no longer carries a copy.
- The baseline, which has no `PMID`, falls back to `PS` + hybrid and only warns.
- The two back ends agree exactly (`max |diff| = 0` on regridded `Q`), and the transpose
  changes dimension order only (`max |diff| = 0` against the same run before it).

Not verified: a model that is not CAM — the CF-metadata path is covered by unit tests, not
by a real non-CAM run. Also, one GenTS run hung for 50 minutes inside the `RESTOM`
derivation right after GenTS wrote the 1.2 GB `PMID` file (peak job memory 37-81 GB), and
completed normally on a rerun; I could not reproduce it and cannot tie it to this change,
but a 93-level pressure field does raise GenTS's peak memory noticeably.


brianpm-ucar:
Pushed `5cccce5f` after an independent review of this PR and #489. Five things, four of
them paths where the model's own pressure field was *not* being preferred, or was not
being reported.

**1. The run now says which pressure it used**, once per case and vertical dimension, in
`regrid_and_vert_interp` — for the test cases and for the reference:

```
	 INFO: 'b.e30_alpha09e_m...389' (lev): vertical interpolation is using the model's own pressure field, 'PMID'.
```

and, when there is none:

```
	 INFO: '<case>' (lev): no pressure field was found, so pressure is being reconstructed
	 from PS and the hybrid coefficients.  That reproduces the model's pressure only for a
	 pure hybrid-sigma coordinate; for the dry-mass coordinate of recent CAM/WACCM it is an
	 approximation.  Add the model's pressure field to the run to avoid it.
```

Before this, nothing announced the source: the only messages on that code path fire when
the PS fallback *also* fails, and the provisioning INFO fires once per run, not per case.
Two runs of the same case could differ for no reason visible in the log.

**2. `cam_ts_done: true` skipped provisioning entirely.** Both back ends return before the
history files are read (`adf_diag.py:648`, `adf_gents.py:225`), so a run on pre-made time
series always fell back to PS + hybrid — however those time series were made, and whatever
the model wrote. The time series directory is now searched for the field by name (the only
handle available without a history file; the CF fallback needs a dataset).

**3. The CF scan could return the vertical coordinate itself.** CF-compliant pressure-level
output labels its own `lev` coordinate `standard_name: air_pressure`, and interpolating a
column against its own coordinate is circular. The scan now requires horizontal structure
(`ndim >= 2`).

**4. `compare_obs` no longer reports a non-problem.** `PMID` has no observational
counterpart and needs none — the interpolation reads it from the climo file — but the
regridding loop would have printed `ERROR: No reference data available for PMID.` on every
model-vs-obs run. A support variable (`plot_diagnostics: False`) now takes a quiet path.

**5. `amwg_table` loops over `plot_var_list`** instead of `diag_var_list`, so it no longer
announces that it is skipping a support variable nobody asked to tabulate.

**Testing.** 204 unit tests pass (four new: the CF rank guard, and provisioning from a time
series directory including the config override and the `False` opt-out); pylint on
`lib/adf_diag.py` 9.86; `black`/`darker` clean. One model-vs-obs run on Casper, deliberately
configured with `cam_ts_done: true` against time series made by an earlier run, exercises
four of the five at once:

```
	    INFO: Found a 'PMID' time series; adding it to the variable list as the pressure field ...
	 ERROR: No reference data available for RESTOM.
	 INFO: 'b.e30_alpha09e_m...389' (lev): vertical interpolation is using the model's own pressure field, 'PMID'.
	    WARNING: Variable 'Q' has a vertical dimension, which is currently not supported for the AMWG Table. Skipping...
```

`RESTOM` is the only ERROR (it genuinely has no obs), there is none for `PMID`, the AMWG
table complains only about a variable the user actually asked for, and no `PMID` plot
reaches the website.

**Known limitation, not addressed here:** the pressure field is looked for in the same
history stream as the 3-D variables, so a run that writes them to different streams still
falls back to PS + hybrid. Reaching across streams means looking outside the stream loop in
both back ends, which is a design change rather than a fix; happy to open an issue for it.
Now at least the run says what it did.

brianpm-ucar:
A second independent review went over both PRs looking for anything the first one missed or
the follow-up broke. It found two real defects in that follow-up, both fixed in `4c913c2b`.

**1. The new message claimed a fallback that does not exist.** The announcement fired
before the code branched, so a *height* coordinate with no pressure field was told
"pressure is being reconstructed from PS and the hybrid coefficients" — and one line later
that the variable could not be interpolated at all. There is no PS + hybrid path for a
height coordinate. It now says what actually happens:

```
	 INFO: 'mpas' (lev): no pressure field was found, and a 'height' vertical coordinate
	 cannot be interpolated without one, so its 3-D variables will be skipped.
```

**2. The time series search missed the layout it was meant for.** I used a plain
`Path(ts_dir).glob(...)`, which cannot see a GenTS archive laid out as
`<component>/proc/tseries/<frequency>/` — exactly what `find_ts_files`
(`lib/adf_file_utils.py:174`) exists to handle. A run with pre-made time series and a
perfectly good `PMID` two directories down still fell back to PS + hybrid. It now goes
through `find_ts_files`, and is anchored on the case name and history stream, since several
cases can share one time series tree.

Also in this commit:

- **The regridded files record the source** as `vert_interp_pressure`, e.g.
  `PMID (the model's own pressure field)` or `reconstructed from PS and the hybrid
  coefficients`. A later run reuses those files without repeating the choice or the message
  about it, so the answer belongs in the file too. (The review noted the announcement is
  silent on a fully cached re-run — nothing is interpolated then, so rather than invent a
  claim, the provenance is now in the file.)
- **The GenTS back end asks for the field after deriving**, matching the ncrcat back end.
  The review traced the ordering and confirmed it makes no difference today — `PMID` has no
  `derivable_from` — but there was no reason for the two to differ.
- **`plot_diagnostics: False` gained a documented second meaning** in the defaults header:
  the regridding step reads it as "needs no reference data".

Four new unit tests cover the nested GenTS layout, case anchoring and stream anchoring
(207 pass). Re-ran the model-vs-obs, `cam_ts_done: true` case on Casper:

```
	    INFO: Found a 'PMID' time series for 'b.e30_alpha09e_m...389'; adding it to the variable list ...
	 ERROR: No reference data available for RESTOM.
	 INFO: 'b.e30_alpha09e_m...389' (lev): vertical interpolation is using the model's own pressure field, 'PMID'.
```

and the regridded `Q` now carries `vert_interp_pressure: PMID (the model's own pressure
field)`, while the 2-D fields carry nothing. pylint 9.86, `black`/`darker` clean.

brianpm-ucar:
A third independent review (this one an Opus pass, asked specifically to find anything that
breaks an existing user's run rather than only to find bugs) turned up a **crash in the most
common ADF configuration**, plus a config break I introduced myself. Both fixed in
`3d507ebf`, with `973b80b2` on top.

## The crash

Any **model-vs-baseline** run whose *baseline* also writes `PMID` died at the start of
regridding with `KeyError: 'PMID'`, every run, with no baseline file written to get past it.

`AdfData.set_reference()` runs once at start-up, so `ref_var_nam` is fixed before
`request_pressure_field` adds `PMID` to `diag_var_list`. The test-case loop already skipped
a variable that is not in `ref_var_nam`, but `_write_reference_files` did not, and
`load_reference_climo_ds` indexes `ref_var_nam` as soon as `get_reference_climo_file` finds
a climo file — and that lookup is a filename glob, so it finds one.

Every run I reported on this PR was model-vs-obs (where `_write_reference_files` is never
called) or had a baseline without `PMID` (where the lookup returns nothing first). The
reviewer found it by tracing rather than running. `_write_reference_files` now skips a
variable the reference bookkeeping does not know about — which is what regridding a pressure
field onto pressure levels deserves in any case.

Verified with a model-vs-baseline run where both sides are a case that writes `PMID`:
before, `KeyError`; now it completes, both sides announce `PMID`, and the regridded files
carry `vert_interp_pressure: PMID (the model's own pressure field)`.

## The config break

Marking **`PINT` `plot_diagnostics: False`** would have silently removed the plots and the
AMWG table row of anyone who lists `PINT` in `diag_var_list` today. That is exactly the
"existing user configs must keep working" clause, for a variable that is only occasionally
the ADF's own. Reverted — `PINT` is plottable again.

`PMID`, which the ADF has always added for itself, keeps the flag, and now carries it in
`adf_variable_defaults_era5-1deg.yaml` too, where it was missing: with this PR that shipped
config would have printed an `ERROR` for `PMID` on every run and plotted it. The unit test
now checks both defaults files.

## Also

- The time-series lookup checks that the run has something on model levels before fetching
  a 3-D pressure field, as the history-file lookup already did. A 2-D run pointed at a tree
  that happens to hold a `PMID` series was computing a whole climatology for nothing.
- The announcement is made inside each branch, once the source is settled, so it can no
  longer claim a reconstruction that then fails; and it is given the field's *name* rather
  than the field, since regridding drops the name — a case using `PMID` was being announced
  as using "the pressure field".
- The `vert_coord_type` gate is gone from the ncrcat back end, so the two back ends decide
  the same way; `request_pressure_field` already asks the question that matters.
- Cleanups the reviewer caught: the dead `pathlib` import left by the previous round, the
  units warning always saying "PMID" even for a model that calls it something else, and
  `amwg_table`'s inputs block still saying `diag_var_list`.

208 unit tests pass (three new, including one asserting a 2-D run does not acquire a
pressure field), pylint 9.86, `black`/`darker` clean, and `git merge-tree` shows no conflict
with #489 — `973b80b2` removes two stray blank lines that had started colliding with it.

The reviewer's behaviour-change inventory is worth reading before merging: the one change
that alters results is the intended one — a hybrid CAM run that writes `PMID` now
interpolates with it instead of `P0*hyam + PS*hybm`, which is the point of #199 and
**belongs in the release notes**. The cost is one extra 3-D time series and climatology per
case, declined with `pressure_field_names: False`.

brianpm-ucar:
Fourth review pass (a Fable model this time, on `3d507ebf` specifically) — **no blocking
findings**, the first clean round. It did confirm the crash fix is correctly placed, that
`compare_obs` and every user-listed variable are unaffected by it, and that the new tests
actually fail when the code is broken (it broke them in a scratch copy to check). Two real
items, fixed in `bac79511`:

**The support-variable test was the wrong test.** The regridding stage decided what counts
as "a variable the ADF added for itself" by reading `plot_diagnostics: False` — which means
"not a diagnostic to plot" and was never meant to answer that question. It now asks for the
pressure field names directly, the only variables the ADF adds here. A variable that is
merely unplotted can no longer take the quiet path, and the second meaning I had documented
in the defaults header is gone again.

**`_ts_uses_level` was walking the tree far too often.** It called `find_ts_files` with the
default `recursive=True` once per variable per stream — so every variable with no time
series triggered an `rglob` of the whole archive, twice over, which is exactly what
`find_ts_files`' own docstring tells callers with many patterns not to do. It now lists each
stream once and filters names in Python, and the question is only asked after a pressure
time series has actually been found. On a 2000-file nested tree: the common 3-D case went
from seconds to **0.02 s**; a 2-D run with a stray PMID series in the tree pays 1.8 s once.

**And the crash now has a test.** `lib/test/unit_tests/test_reference_support_vars.py`
drives `_write_reference_files` with a stub ADF: the pressure field is skipped, a configured
pressure name is skipped, an ordinary variable missing from the reference bookkeeping still
raises rather than being swallowed, and a user-listed `PMID` is written as usual.

Smaller: pass the field's name rather than `str()` of it (`str(None)` is the truthy string
`"None"`), and drop an assignment that went dead when the announcement moved into its branch.

212 tests pass, pylint 9.86, `black`/`darker` clean, `git merge-tree` clean against #489
(their two insertion points had drifted onto the same line again).

Two things I did **not** change, for the maintainers to weigh:

1. **`PINT` is plottable again** (reverted last round after it was flagged as a config
   break). The reviewer points out the other side: when the ADF adds `PINT` for a run with
   interface variables, it is then handed to the plotting scripts and the AMWG table, which
   report a missing regridded file. Doing this properly means recording *what the ADF added*
   rather than reading it off the variable defaults — a small new mechanism, and I would
   rather not add one in a fourth round of fixes. Say the word and I will.
2. **The root cause is still there.** `AdfData` snapshots the variable list when it is
   built, and `add_diag_var` never refreshes it. The guard covers the pressure fields;
   `adf_derive`'s aerosol block also appends `T` late, and on `main` today a baseline run
   with aerosols crashes at the same `ref_var_nam` lookup if the baseline has a `T` climo.
   That predates this PR and I have deliberately left its behaviour alone. Worth its own
   issue.

Unrelated, but found on the way: `test_adf_base.py::test_AdfBase_debug_write` is flaky.
`AdfBase` names its log file to the second and `logging.basicConfig` only takes effect once
per process, so whichever test builds the first debug-enabled `AdfBase` owns the handler; a
later one in a different second asserts on a file nothing was written to. It failed once in
212 tests for me and passes on re-run. CI runs this suite on five Python versions, so it
will surface eventually.

brianpm-ucar:
I've been through several rounds of Claude reviews, and I've actually looked at the code. I think this PR and it's companion are working and probably have over-engineered the response to this issue. 

I'm going to merge. If we run into any issues with vertical interpolation, let's open a new issue. My guess is that things are working well, but we could probably make some simplifying assumptions by asking users to be clear about what they want to happen in the config file.

PR #489, opened 2026-09-20
Let the config choose the vertical interpolation levels (#199)

brianpm-ucar:
Part 2 of #199, independent of #488 (part 1) — either can go in first.

Every 3-D field was interpolated onto one hard-coded set of pressure levels, 1000 to
1 hPa (`regrid_and_vert_interp.py:14-21`). A model whose top is above 1 hPa had nothing to
show up there, and `plot_press_levels` could only select from that fixed set.

## What this does

`interp_press_levels` (hPa, in `diag_basic_info`) replaces the set, for the test cases and
the reference alike, through all three interpolation paths: the model's own pressure field,
`PS` + hybrid coefficients, and pressure to pressure. Left out, the ADF's standard levels
are used exactly as before.

The levels are validated. A level that is zero, negative or not a number interpolates to
nothing but NaN, so the run stops at the start of regridding with a message naming the
offending value, rather than writing NaN into every 3-D field. They are sorted largest
pressure first, matching the ADF's own set, so the order in the config does not matter.

`plot_press_levels` is checked against them at the same point — a plotting level that was
never interpolated to cannot be plotted, and saying so once up front beats a plotting
script finding nothing later.

## Two fixes that this needs

Both were reachable before, but asking for a level above a model's top makes them easy to
hit, and either one ends the whole run:

- **`plotting_utils.get_difference_colors`** took its range with `np.min`/`np.max`, so a
  single missing point in the field made every branch false and left `cmap` unassigned:
  `UnboundLocalError: cannot access local variable 'cmap'`. It now ignores non-finite
  values, and returns a usable norm for a field that is missing everywhere.
- **`polar_map`** selected the plotting level without checking that it exists
  (`polar_map.py:253`), so an unavailable level ended the run in a `KeyError` from `.sel`.
  It now warns and skips, the way `global_latlon_map` already did.

## Testing

`pytest lib/test/unit_tests` — 208 passed, including new `test_interp_levels.py` (defaults,
hPa to Pa, ordering, sub-1 hPa levels, and each bad input rejected) and
`test_difference_colors.py` (all-missing, one-missing-point, positive/negative definite,
straddling zero). `black --check lib/` and `darker` against `main` on `scripts/` are clean.

Model-vs-model run on Casper with `interp_press_levels: [0.1, 0.5, 850, 200, 1000]` and
`plot_press_levels: [200, 850, 0.5, 500]` — a deliberately unsorted set, two levels above
the 32-level baseline's top, and one plotting level that is not in the set:

```
INFO: Interpolating onto 5 pressure levels (hPa): [1000.0, 850.0, 200.0, 0.5, 0.1]
WARNING: 'plot_press_levels' asks for [500] hPa, which is not among the levels being
         interpolated to, so those levels cannot be plotted.  Add them to
         'interp_press_levels'.
```

The regridded files carry exactly those five levels in that order; the 93-level test case
has data at 0.1 hPa and the 32-level baseline is NaN there; `global_latlon_map` and
`polar_map` both skip 500 hPa with a warning and the run finishes. Before the two fixes
above, the same run died in `polar_map` with the `UnboundLocalError`.

Note for anyone using this: changing the levels invalidates existing regridded files, so
set `cam_overwrite_regrid: true` the first time. The example config says so.

Not verified: deriving levels from the model's own coordinate (the `P0*hyam + PS*hybm`
global-mean idea in the issue thread) is not attempted here — this is the "user-specified
levels" half, and the derived-levels half probably wants its own issue.


brianpm-ucar:
An independent review of this PR and #488 found no correctness problems in this one's own
code. Its one blocking finding against this PR was shared with #488 — nothing told the user
whether a case interpolated with the model's own pressure field or reconstructed pressure
from PS and the hybrid coefficients. That is now fixed in #488 (`5cccce5f`), in this same
function, since it belongs with the provisioning change; the two branches still merge
cleanly in either order (verified with `git merge-tree`).

Two smaller points from the review, left as they are:

- `interp_press_levels: False` is rejected with
  `Bad 'interp_press_levels' in the config file: expected a list of pressure levels in hPa,
  got False`, rather than meaning "use the default" the way `pressure_field_names: False`
  means "do not use the model's pressure field" in #488. There is no sensible "off" for the
  interpolation levels, so a clear hard stop seemed better than a silent default. Happy to
  change it if the inconsistency bothers anyone.
- The `plot_press_levels` cross-check compares floats after a hPa/Pa round trip. Exact for
  the round numbers in every realistic level set; a level like `862.53` could in principle
  miss. It only controls a warning, so the cost of a miss is a warning that does not appear.

brianpm-ucar:
A second independent review found nothing blocking in this PR — it separately traced the
`get_difference_colors` rewrite and confirmed the branch removed from the old code was
unreachable for finite input, so nothing of value was lost, and that the `polar_map` guard
matches `global_latlon_map`'s existing convention.

Two small things fixed in `397ae7fe`:

- **The merge conflict I wrongly said did not exist.** Both #199 branches documented their
  new config entry immediately after `plot_press_levels`, which collides in
  `config_cam_baseline_example.yaml` for whichever lands second. This one moves a few lines
  down, still among the other plot/level settings. `git merge-tree` now reports no conflict
  between the two branches at all — my earlier comment was right about the code and wrong
  about the YAML.
- The `raise` after `end_diag_fail` in `_interp_levels` is explicit, so a bad
  `interp_press_levels` cannot reach the lines below if `end_diag_fail` ever stops raising.

brianpm-ucar:
Third independent review: nothing blocking in this PR, and it separately confirmed that with
`interp_press_levels` unset the levels are bit-identical to the old hard-coded set, so an
existing config sees no numerical change and no new log output.

Three small things fixed in `37148b6f`, none of them a behaviour change for a valid config:

- A scalar `plot_press_levels: 850` (rather than `[850]`) made the new cross-check raise
  `TypeError` from the regridding stage. That config was already broken further downstream,
  but this PR moved the failure earlier and turned it into a traceback. The check is a
  convenience and now stays out of the way.
- `interp_press_levels` starts from the default before validation, so there is no path where
  the lines below could run on an unbound name and no bare re-raise of the `ValueError` that
  `end_diag_fail` exists to avoid. (This replaces the explicit `raise` from `397ae7fe`.)
- Repeated levels are dropped — `[500, 500]` would have been interpolated and written twice.

`git merge-tree` shows no conflict with #488 (a couple of blank lines on that branch had
started colliding; they are gone).

Issue #246, opened 2023-05-09
Model vs Obs Lat Lon plots of U200


JulioTBacmeister:
### New feature type

New plot and/or plot enhancement

### What is this new feature?

I'd like maps of U200 (Zonal wind at 200 hPa) vs reanalyses - ERA5 or MERRA would be fine.  

### Assistance required?

Yes, I am requesting a new feature but won't have time to develop it

brianpm:
I recently added monthly mean U from ERA5, and it has 200hPa. So this should be working already, I think. Maybe @justin-richling can check if it actually does work.

justin-richling:
Indeed, it does appear that [U200](https://project.cgd.ucar.edu/projects/ADF/tests/U200/html_img/plot_page_U_200hpa_ANN_LatLon_Mean.html) works with the new obs @brianpm put in recently.

nusbaume:
@JulioTBacmeister Are you just wanting to compare the zonal winds between CAM and ERA5 at 200 hPa using the `U` variable from CAM?  If so then @brianpm is correct that we already have that functionality.

However, if you are looking specifically for a comparison between the `U200` CAM output variable and ERA5 or MERRA then yes, it looks like we'll need to create a new ADF variable defaults entry and ERA5 obs variable in order to do so.

brianpm:
@nusbaume -- That's right, I didn't think about that distinction (U200 in the CAM output vs specifying the 200hPa level with `plot_press_levels`.

JulioTBacmeister:
Thanks. This looks good.  I just wanted a comparison of U200 from CAM against any reanalysis. I'm not picky.  It would also be nice to add U200 to the exp vs CTL plots.

PR #491, opened 2026-09-21
Make and compare fields on a pressure surface (#227, #246)

### LINKED ISSUE #227: Create TS files for pressure levels when requested
### New feature type

New regridding method or regridding enhancement

### What is this new feature?

@justin-richling 

We would find it extremely useful (especially when working with other tool-boxes (i.e. the NOAA mdtf) if we could specify the creation of TS netcdf files for specific model levels when they are requested (i.e. [200, 850,...etc]) from the 3D fields.

IE. when we add: 

'''
diag_var_list:
         -U850 
         -U200
'''

Those TS nc's are created from the 3D U field.  
This interpolation code is currently available in the plotting routines (i.e., /ADF/lib/plotting_functions.py), but it would be awesome if we could utilize it for standard pressure level output also! 

### Assistance required?

### Extra info

_No response_

brianpm-ucar:
Closes #227. Closes the remaining half of #246.

Two halves of one feature, done together because neither is much use alone:

- **#227 (model side):** `U200`, `U850` in `diag_var_list` should produce time series
  interpolated from the 3-D field, for the ADF and for toolboxes downstream (MDTF).
- **#246 (observation side):** those fields have to be compared against a reanalysis, and
  reanalyses come as whole columns. Staging a file per level does not scale — this reads the
  level out of the file the ADF already ships.

## Model side

A variable default may now say:

```yaml
U200:
  derivable_from: ["U"]
  derive_level: 200      # hPa
```

A run whose history files have `U` but not `U200` gets `U200` made during time series
creation — the stage that already derives `PRECT` from `PRECC` and `PRECL`. The pressure is
the model's own 3-D field where there is one and `PS` + hybrid coefficients otherwise, the
same order of preference the regridding stage follows (#199). **A run that writes `U200`
itself keeps what it wrote**; the derivation only fires when the variable is absent.

One subtlety worth a reviewer's eye: asking for the pressure field has to wait until the
constituent is known. Until `check_derive` adds `U`, nothing in the run is on model levels,
so `request_pressure_field` would decline — it now takes the variables to judge that by. Two
tests cover exactly this, because my first attempt got it wrong in both directions (it
silently fell back to PS+hybrid, and it clobbered the variable list the loop was walking).

## Observation side

```yaml
  obs_file: "U_ERA5_monthly_climo_197901-202112.nc"
  obs_var_name: "U"
  obs_lev: 200           # hPa
```

`AdfData` takes that surface as the observation is read, so one file serves every level.
The nearest level is used with a warning past 1%; a file that is *already* a single surface
is left alone, so an entry works against either; coordinates in Pa and the usual dimension
names (`level`, `plev`) are understood.

## Shipped

`U200`, `V200`, `U850`, `V850` in both defaults files — no configuration needed. `U200`
moves off the file staged for it in #490 and onto the level of the 3-D file; that file is
unchanged, still readable, and holds the same numbers (verified bit-identical).

`pressure_in_pa` moves from the regridding script into `adf_utils`, since both stages need
it now.

## Testing

248 unit tests pass (18 new, covering both mechanisms and their edges: Pa coordinates,
alternative dimension names, the distant-level warning, a 2-D observation, a 2-D model
field, no pressure at all, and the ordering bug above). `black`/`darker` clean; no file in
CI's linted set is touched.

On the CESM3 alpha case, whose `cam.h0a` stream writes `U` and `V` but none of the four
surfaces:

```
INFO: Adding 'PMID' to the variable list; it is the pressure field ...
INFO: 'U200' interpolated to 200 hPa using PMID.
INFO: 'U850' interpolated to 850 hPa using PMID.
INFO: 'V200' interpolated to 200 hPa using PMID.
INFO: 'V850' interpolated to 850 hPa using PMID.
```

All four regridded, plotted and tabulated against the ERA5 levels. `U850` scores RMSE
1.05 m/s, bias -0.12; `U200` RMSE 2.4, bias -0.5.

**The check that matters:** against CAM's *own* `U200`, written on the daily `cam.h2a`
stream of the same case and years, the derived field agrees to an RMS of **0.12 m/s** (max
0.53) — which is the difference between interpolating a monthly mean and averaging daily
interpolations, not an error in the interpolation.

## Notes for review

- Asking for `U200` pulls `U` into `diag_var_list`, so `U`'s own plots appear too. That is
  how `check_derive` has always worked (`PRECT` pulls in `PRECC` and `PRECL`); flagging it
  because the 3-D case is heavier than the 2-D one.
- At 850 hPa the model is missing below ground while the reanalysis is filled, so
  differences over the Himalayas, the Andes and Antarctica are not meaningful. Inherent to
  the comparison, not to this change.
- The percent-difference panel remains useless for a field that changes sign, as it is for
  `U` and `V` today.
- `Z500` is not shipped: ERA5 distributes geopotential, not geopotential height, so it needs
  a division by g and a decision about which field CAM's `Z500` should be compared to.


brianpm-ucar:
An independent review of the derivation hook found three blocking defects and — more usefully
— that the tests would not have caught either of the two bugs I had already shipped and fixed
in this branch. All fixed in `6f448345`.

## The one that mattered

**`U850` was inventing data below ground.** The model-pressure path goes through `np.interp`,
which clamps outside the column instead of returning NaN, while the PS + hybrid path (geocat)
returns NaN. So the same field was missing or filled over Tibet, Greenland and Antarctica
depending only on whether the case wrote `PMID` — and where it was filled, the number plotted
as the 850 hPa wind was the lowest model level's wind. `U850` and `V850` are shipped on by
default, so this was default-on wrongness, and in a test-vs-baseline comparison it would have
been filled on one side and missing on the other.

Reproduced directly (700 hPa surface, asked for 850 hPa):

```
PMID path  -> [5. 5.]   (the model's bottom level, labelled 850 hPa)
PS+hybrid  -> [nan nan]
```

Points outside the column's own range are now masked, as the regridding stage already does.
On the test case, `U850` goes from **0.0000 to 0.1052 missing** — which is the fraction of the
globe where 850 hPa is underground.

## The other two

- **A pressure time series covering *more* times than the constituent** passed the coverage
  check and then met the field on the union of the two axes. Cut to the field's own times now,
  so a field cannot be paired with another period's pressure.
- **Memory.** `pmid_to_plev` stacks and calls `.values`, and nothing upstream is chunked, so
  the interpolation materialized the whole 3-D series twice — reviewer's extrapolation was
  ~155 GB for the 30-year daily case that #227 actually asks for. It now runs a few time steps
  at a time.

## Also fixed

- `np.interp` needs the source pressures increasing and does not check: a column stored
  bottom-up came back quietly wrong. (This bit one of my own new tests, which is how I believe
  it.)
- The fallback only looked for `PS` inside the constituent's file, so **GenTS runs and
  pre-made archives had no fallback at all** — GenTS writes `PS` to its own file by design.
- The interface case asked for `hyam`/`hybm` where CAM has `hyai`/`hybi`, and raised.
- A pressure-surface field could be derived from **another stream's** constituent and written
  over the copy CAM wrote itself. Both searches now stay in the stream the output is named
  for, so the claim "a run that writes U200 itself keeps what it wrote" is true for a
  multi-stream run too.

## The tests now fail when the hook does

Six deliberate regressions, each confirmed to fail the suite — and each confirmed to pass it
before this commit:

| regression | before | after |
|---|---|---|
| the loop's variable list replaced (dropped V200/V850) | 248 passed | 2 failed |
| pressure request judged by the wrong list (silent PS fallback) | 248 passed | 2 failed |
| below-ground mask removed | passed | 1 failed |
| pressure time selection removed | passed | 1 failed |
| column-ordering guard removed | passed | 1 failed |
| PS time series search removed | passed | 1 failed |

258 tests pass; `black`/`darker` clean. The end-to-end run is unchanged for `U200`
(`max |diff| = 0` against the previous run, RMS 0.118 m/s against CAM's own `U200`), and the
job used 16 GB.

## Left alone, deliberately

- No vertical-coordinate-type check in the fallback: it assumes hybrid-sigma when `PS`,
  `hyam` and `hybm` are all present. A height or pressure-level stream carrying all three is
  contrived, and the regridder's own check is not reachable from here without more plumbing.
- #227 asked for levels named in the config file; this delivers a generic mechanism with four
  entries shipped. A user wanting `T500` adds three lines to a defaults file rather than a line
  to the config. Happy to add more entries if there is a standard set people want.

brianpm-ucar:
A second review, this time of the whole PR (the observation side had never been looked at),
found two blocking defects. Both fixed in `bc3be2dd`, and both are the same shape: correct in
the case I tested, wrong one step to the side.

**The derived field loses its units in the environment the ADF pins.** The interpolation took
its attributes from the constituent *sum*, and whether arithmetic keeps attributes depends on
the xarray version:

```
xarray 2026.2 (my env):   0 + da  ->  {'units': 'm/s', 'long_name': 'Zonal wind'}
xarray 2024.1.1 (pinned): 0 + da  ->  {}
```

So in `env/conda_environment.yaml`'s environment the AMWG table prints `--` as the unit for
all four shipped variables and the long_name degrades to the variable name. Attributes now
come from the constituent itself — which is how the older derivations always took them — and
a test runs the derivation under both `keep_attrs` settings.

**The below-ground mask guessed its units.** `pmid_to_plev` is called with
`convert_to_mb=False`, so its levels are Pa, but the mask read them as "hPa if small". A
request for 10 hPa (1000 Pa) was read as hPa and masked the **entire field**; 5 hPa passed the
range test and turned the mask off instead. The shipped 200/850 entries are unaffected — which
is exactly why nothing showed it — but the docs invite any other surface, and the QBO levels
are where someone would start. One line: the levels are always Pa.

**Plus a 40× read cost on the observation side.** `load_da` does its unit conversion before
the level selection, and that arithmetic materializes the array, so each read of a 0.25° ERA5
file pulled in 1.8 GB to keep 50 MB. The reviewer's process was killed doing it; so was mine.
Measured on `U_ERA5_monthly_climo_197901-202112.nc`:

```
selection inside load_da (now):     peak RSS 0.47 GB, 4.7 s
selection after the scaling (was):  Killed
```

Also from that review: the level selection is now refused outright for a baseline simulation
rather than relying on its fields already being 2-D; a 3-D observation whose vertical
dimension has an unfamiliar name warns instead of passing a column through to the plotting
scripts; `V850` and `V200` get contour ranges that do not saturate on the Somali jet (it
reaches 18 m/s in the annual mean against a `[-8, 9]` range); the four entries now sit with
the fields they are derived from; `OBSERVATIONS.md` lists which variables each ERA5 file
serves; and the defaults header documents `derivable_from` once.

263 tests pass, `black`/`darker` clean. The end-to-end run is unchanged except where intended:

```
U200: units='m/s'  long_name='Zonal wind at 200 hPa'       nan=0.000
U850: units='m/s'  long_name='Zonal wind at 850 hPa'       nan=0.105
V200: units='m/s'  long_name='Meridional wind at 200 hPa'  nan=0.000
V850: units='m/s'  long_name='Meridional wind at 850 hPa'  nan=0.105
```

Known and left alone, both noted by the reviewer: the time-blocked interpolation exists only
on the model-pressure path, so a model without a pressure field still hands geocat the whole
series (that path is dask-aware, and it is not the modern default); and `vert_remap` is a
Python loop over columns, which is the real cost for #227's daily use case and wants
vectorizing in its own PR.

brianpm-ucar:
Two independent reviews — one on scope and blast radius, one asking only whether anything an
existing user does changes — plus a set of regression runs that now covers the paths nobody
was exercising. Neither review found a defect in the feature. Both found the diff carrying
things the feature does not need, and one found a behaviour change I had introduced without
meaning to. `632c96e4` addresses both.

## The diff is now the feature

The defaults files were most of the apparent bulk. Regrouping the entries had collapsed a
stray blank line before about thirty unrelated variables — whitespace churn of exactly the
kind AGENTS.md §6.1 asks to keep out. Both files are now `main`'s, plus three new entries,
the rewritten `U200`, and the two header notes: **34 hunks down to 5**, with a full parse
confirming the other 303 entries are value-identical and in the same order.

What remains, excluding tests and documentation, is **513 lines across five files**, and
`adf_derive.py`'s 362 added lines are 92 of comment and docstring to 270 of code.

## U200 keeps reading the file staged for it

Pointing it at the 200 hPa level of the 3-D file was tidier, but that file is 1.8 GB against
50 MB and every plotting script pays the difference — and a mirror of the observations
directory holding only the small file would have stopped working. The values are identical
either way (verified bit-for-bit), so the tidiness bought nothing. `U850`, `V200` and `V850`
have no staged file and read their level out of the 3-D one, which is what `obs_lev` is for.

Also: `load_reference_timeseries_da` asked for the level under the observation's variable
name rather than the ADF's, so the lookup could never match. Inert today — no script reaches
it under `compare_obs` — but a trap for whoever stages observational time series.

## Five regression runs, including the three paths nothing covered

A harness now lives in `/glade/derecho/scratch/brianpm/ADF_regress` (configs, a PBS script, a
`verify.py` that fingerprints every regridded field, and a recorded baseline). All five pass
with no tracebacks:

| run | what it covers |
| --- | --- |
| CAM7 `h0a` 93L vs CAM6 `h0` 32L | different vertical grids and stream names; both pressure sources in one job |
| one case with two streams at once | the stream loop |
| AMIP vs obs | 3-D T/U/Q against reanalysis, plus a surface against one level of a 3-D observation |
| `cam_ts_done: true` | **the pre-made-time-series path** |
| `ts_tool: gents` | **the GenTS back end** |

The last two were flagged as exercised by no test and no run. Both work, and both earn their
keep:

- On pre-made time series the two sides take different routes in the same job — the test
  case's series include `PMID` (`interpolated to 850 hPa using PMID`), the baseline's do not
  (`using PS and the hybrid coefficients`).
- The GenTS run **only works because of the PS-in-its-own-file search**, which the scope
  review had listed as the most droppable of the fixes. GenTS writes `PS` to a file of its
  own; I confirmed the `U` time series it produced contains `hyam`, `hybm` and `P0` but no
  `PS`. Without that search this run would derive nothing.

One finding from building the harness, unrelated to this PR: the ADF names every output file
after the case, so a case compared with itself across two history streams collides with
itself and silently produces no plots at all. Worth knowing; not something I have touched.

## What is still uncovered

`vert_remap` is a Python loop over columns, which is the real cost for #227's daily use case
and wants vectorising in its own PR. The `ilev` branch of the fallback (5 lines) is reachable
by nothing that ships and is tested only synthetically — droppable if a reviewer would rather
not carry it.

